PR #103, opened 2026-05-07
feat(autoconfig): introspective controller architecture

benzsevern:
## Summary
- Adds `AutoConfigController` — an iterative refit loop over a `RefitPolicy` protocol that runs the pipeline on a stratified sample, reads stage-emitted complexity signals, and proposes config refits until green / converged / budget exhausted
- Replaces the untyped `PostflightSignals` TypedDict with a typed `ComplexityProfile` (6 frozen sub-profile dataclasses with explicit health rules)
- Wires `auto_configure_df` through the controller; refactors zero-config callers in `_api.py` and the web UI to call `auto_configure_df` *before* the pipeline (eliminating the double-pipeline-run that would otherwise occur)
- Ships `HeuristicRefitPolicy` with 5 hand-coded rules; protocol leaves room for `LearnedRefitPolicy` / `LLMRefitPolicy` later

## Why

PR #102 fixed two crash paths in zero-config `match_df` and `match()`. Measuring after that fix gave DBLP-ACM zero-config F1 = 0.5102 vs hand-tuned 0.918 — and `CLAUDE.md` notes that auto-config fails on every standard benchmark. Today's auto-config is a feedforward function that never inspects what its config does to real data. This PR rebuilds it as an introspective pipeline so the same problem can be addressed across datasets without per-domain hand-tuning.

## What's in here (16 commits)

**Foundation:**
- `core/complexity_profile.py` — `HealthVerdict` + 6 frozen sub-profile dataclasses (`Data`, `Domain`, `Matchkey`, `Blocking`, `Scoring`, `Cluster`) + `ProfileMeta` + rollup
- `core/profile_emitter.py` — `ContextVar`-backed thread-local emitter stack with `_NullEmitter` singleton (zero cost when no capture is active)
- `core/autoconfig_history.py` — `RunHistory` with oscillation detection, profile-distance convergence, and the cheapest-healthy lex ordering from spec §S1-B

**Stage instrumentation** (surgical edits, no behavior change when emitter is null):
- `core/blocker.py` emits `BlockingProfile` after `build_blocks`
- `core/scorer.py` emits `ScoringProfile` from the fuzzy scorer (sequential + parallel paths)
- `core/cluster.py` emits `ClusterProfile` with a sampled transitivity rate from `build_clusters`
- `core/matchkey.py` emits `MatchkeyProfile` from `compute_matchkeys` post-transform columns
- `core/autoconfig.py` + `core/domain.py` emit `DataProfile` and `DomainProfile`
- New `core/_profile_helpers.py` (histogram / Hartigan dip / mass / transitivity helpers); adds `diptest>=0.5.2` as a hard dep

**Policy:**
- `core/autoconfig_policy.py` — `RefitPolicy` protocol + `HeuristicRefitPolicy` dispatcher with same-config bug guard
- `core/autoconfig_rules.py` — 5 ordered rules: `blocking_too_coarse`, `unimodal_scoring`, `low_reduction_ratio`, `low_transitivity`, `no_matches`

**Controller:**
- `core/autoconfig_controller.py` — `AutoConfigController` + `ControllerBudget` + `StopReason` enum + `_RED_PROFILE` sentinel + pathological-input gates + stratified-deterministic sample selection + iteration loop with crash recovery + `_finalize` with sample-vs-full drift detection
- `_legacy_auto_configure_v0` extracted in `core/autoconfig.py` (private helper used as v0 by the controller's `_initial_config` — breaks the recursion that would otherwise occur when the public `auto_configure_df` calls back through the controller)

**Wire-up:**
- Public `auto_configure_df(df, *, reference=None, …)` becomes a controller-backed facade; gains `reference` kwarg for cross-source mode
- `_LAST_CONTROLLER_RUN` ContextVar so calling pipelines can pull the profile + history into `PostflightReport`
- Zero-config callers in `_api.py::dedupe_df`, `_api.py::match_df`, and the web UI run `auto_configure_df` BEFORE the pipeline and pass `auto_config=False` explicitly — no more double pipeline runs

## Test status

- **1684 passed** (1572 baseline + 112 new) on the standard exclusion set (db / mcp / torch tests skipped per `packages/python/goldenmatch/CLAUDE.md`)
- Focused suites: 22 `test_complexity_profile`, 9 `test_profile_emitter`, 15 `test_autoconfig_history`, 8 `test_stage_instrumentation`, 19 `test_autoconfig_policy`, 19 `test_autoconfig_controller`, 9 `test_autoconfig_facade`, 5 `test_autoconfig_no_double_run`
- DBLP-ACM regression from PR #102 (`TestMatchZeroConfig`) still passes
- 1 pre-existing failure unchanged: `test_ncvr_autoconfig_no_useless_matchkeys` (gitignored NCVR sample dataset, not present locally)

## What this PR explicitly does NOT do

- **Close the DBLP-ACM accuracy gap.** Zero-config F1 went 0.5102 → 0.5133 — material architectural validation (the controller iterates, rules fire, the loop finds a cheaper-healthy candidate) but not a meaningful precision lift. The diagnosis is concrete: `rule_blocking_too_coarse` runs first and switches blocking from `__title_key__` to `authors`, which on the 2000-row sample looks GREEN but on full data over-generates pairs. The controller faithfully commits the cheapest-healthy candidate; the *response to sample-vs-full drift* (recorded as `history.full_vs_sample_drift`) is left for a follow-up
- **Migrate `PostflightSignals` consumers to read typed `ComplexityProfile`.** The TypedDict is preserved for back-compat; consumers continue to read it. Migration is a separate PR
- **Refactor the file-based `_api.py::match` zero-config branch.** Restructures I/O ordering — non-trivial, controller's sample iterations don't go through the file path so it's not in the double-run hot path. Documented as deferred
- **Property tests / controller integration tests / benchmark regression test.** The pair-conversion helper for benchmarks was built and validated locally but not committed (see follow-up)
- **TS port parity.** `packages/goldenmatch-js/` continues with the legacy heuristic. Spec explicitly calls this v2 work

## Follow-ups

1. **Rule tuning + ordering.** Either re-prioritize rules so `rule_no_matches` fires before `rule_blocking_too_coarse` on the singleton-trap signal (`n_pairs_scored == 0`), or add a rule that handles "blocking too coarse + no usable alternate column → propose soundex/n-gram on text"
2. **Act on drift in v1.** When `full_vs_sample_drift > threshold`, re-run one more iteration with the full-data profile or fall back to v0
3. **Migrate consumers from `PostflightSignals` TypedDict to typed `ComplexityProfile`** — separate PR
4. **Property tests + benchmark regression tests** — separate PR; the architecture is now testable end-to-end
5. **Document architecture in `CLAUDE.md`** — replace the "always use explicit config" caveat once a benchmark target is met

## Architecture spec

`docs/superpowers/specs/2026-05-06-autoconfig-introspective-controller-design.md` (gitignored per repo convention; design doc is local).

### DIFF
--- a/.profile_tmp/baseline_febrl3_ncvr.py
+++ b/.profile_tmp/baseline_febrl3_ncvr.py
@@ -1,0 +1,95 @@
+"""Measure v1.7.1 zero-config F1 on Febrl3 + NCVR.
+
+Used to set v1 acceptance targets for the auto-config controller. Run once
+before implementation begins; results inform §Testing tier 4 of the spec.
+"""
+from __future__ import annotations
+import sys, io, time
+sys.stdout = io.TextIOWrapper(sys.stdout.buffer, encoding="utf-8")
+from pathlib import Path
+import polars as pl
+import goldenmatch as gm
+
+
+def evaluate(name: str, found: set, gt: set) -> None:
+    tp = len(found & gt)
+    fp = len(found - gt)
+    fn = len(gt - found)
+    p = tp / (tp + fp) if (tp + fp) else 0.0
+    r = tp / (tp + fn) if (tp + fn) else 0.0
+    f1 = 2 * p * r / (p + r) if (p + r) else 0.0
+    print(f"{name}: P={p:.4f} R={r:.4f} F1={f1:.4f}  (TP={tp} FP={fp} FN={fn})")
+
+
+def expand_clusters_to_pairs(clusters: dict) -> set:
+    """Transitive closure of in-cluster edges. Singletons yield no pairs."""
+    pairs = set()
+    for c in clusters.values():
+        members = sorted(c["members"])
+        for i in range(len(members)):
+            for j in range(i + 1, len(members)):
+                pairs.add((members[i], members[j]))
+    return pairs
+
+
+def main() -> None:
+    print(f"goldenmatch v{gm.__version__}")
+    print(f"polars v{pl.__version__}")
+
+    # ---- Febrl3 ----
+    print("\n=== Febrl3 ===")
+    try:
+        from recordlinkage.datasets import load_febrl3
+    except ImportError:
+        print("recordlinkage not installed; skipping Febrl3")
+    else:
+        try:
+            df_pd, gt_pairs = load_febrl3(return_links=True)
+            df_pd = df_pd.reset_index().rename(columns={"rec_id": "id"})
+            df = pl.from_pandas(df_pd)
+            print(f"Febrl3: {df.height} rows, cols={df.columns}")
+            t0 = time.time()
+            result = gm.dedupe_df(df)
+            print(f"  dedupe_df completed in {time.time() - t0:.2f}s")
+            found = expand_clusters_to_pairs(result.clusters)
+            # Map row_id -> original id; ground truth uses original ids
+            row_to_id = df["id"].to_list()
+            found_ids = set()
+            for a, b in found:
+                if 0 <= a < len(row_to_id) and 0 <= b < len(row_to_id):
+                    pa, pb = row_to_id[a], row_to_id[b]
+                    found_ids.add((min(pa, pb), max(pa, pb)))
+            gt_set = set()
+            for a, b in gt_pairs:
+                pa, pb = (a, b) if isinstance(a, str) else (str(a), str(b))
+                gt_set.add((min(pa, pb), max(pa, pb)))
+            evaluate("Febrl3 (zero-config dedupe)", found_ids, gt_set)
+        except Exception as e:
+            print(f"Febrl3 failed: {type(e).__name__}: {e}")
+
+    # ---- NCVR ----
+    print("\n=== NCVR sample ===")
+    ncvr_path = Path("packages/python/goldenmatch/tests/benchmarks/datasets/NCVR/ncvoter_sample_10k.txt")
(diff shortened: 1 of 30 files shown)


benzsevern:
Pushed fixes for the four in-PR review items + bonus warning log:

- **Fix 1**: `_LAST_CONTROLLER_RUN` now wired to `PostflightReport` via two new optional fields (`controller_profile`, `controller_history`). Zero-config callers stamp the tuple onto the postflight after the pipeline runs.
- **Fix 2**: `KeyboardInterrupt` re-raised instead of falling through to commit. `history.elapsed` still recorded.
- **Fix 3**: `auto_configure_df` kwargs (`strict`, `llm_provider`, `domain_config`, `llm_auto`, `allow_remote_assets`) threaded through controller via new `v0_kwargs` param to `_legacy_auto_configure_v0`.
- **Fix 4**: New `_skip_finalize` private kwarg eliminates the double full-pipeline run when called from `_api.dedupe_df`/`match_df`. Direct-call path unchanged.
- **Fix 5**: `_take_sample` docstring corrected (uniform random; stratification noted as v2 TODO).
- **Bonus**: `logger.warning` on all-RED fallback to v0, with total/crashed/RED iteration counts.

16 new tests. Regression: 1695 passed (up from 1684), 1 pre-existing NCVR-sample-missing failure unchanged.

Commit: d9ac102

PR #121, opened 2026-05-10
release(goldenmatch): v1.11.0 - negative evidence + clustered-identity guard

benzsevern:
## Summary

v1.11.0 ships the **negative-evidence + clustered-identity-guard infrastructure** as a foundation for T3 recovery. **DQbench composite essentially unchanged at 66.99 (v1.10 was 66.91)** — the indicator-driven demote rule cannot fire on real DQbench T3 because of two structural findings surfaced during Phase 7 measurement. v1.12 will extend NE into exact matchkeys (Path Y from spec) to actually deliver T3 movement.

## What shipped

- **`NegativeEvidenceField` schema** on `MatchkeyConfig` (default-None for v1.10 cache compat). Each entry: `field`, `transforms`, `scorer`, `threshold`, `penalty`. Validates against `VALID_SIMPLE_TRANSFORMS` and `VALID_SCORERS`.
- **`_apply_negative_evidence`** in `core/scorer.py` subtracts NE penalties from the weighted-matchkey positive score. Final score = `max(0, score_positive - score_negative)`. Defensive: skips with WARNING on unknown scorers, missing fields, or scorer-call exceptions.
- **`promote_negative_evidence`** eager rule (in new `core/autoconfig_negative_evidence.py`) populates NE fields at config-build time for columns with `identity_score >= 0.75 AND cardinality_ratio >= 0.5 AND col not in matchkey.fields AND col not in blocking AND col has an exact-matchkey counterpart`. The exact-matchkey gate prevents false-promotion on T2 noise.
- **`compute_identity_collision_signal`** lazy indicator (8s budget, via `IndicatorContext`) measures within-email-group address divergence to detect adversarial-collision identity columns.
- **`rule_demote_clustered_identity`** at position 7 in `DEFAULT_RULES`. Demotes exact matchkey to fuzzy participation when collision_rate > 0.75. Threshold tightened from spec's 0.5 to 0.75 after measurement showed T2's collision rate (0.62) was triggering false demotes.
- **30+ new tests** (~1986 passing, +30 over v1.10). Tier 4 synthetic regression test passes; Tier 5 v1.10 cache compat verified.

## Honest framing — why T3 didn't move

Phase 7 diagnostic on real DQbench T3 (10K rows, ~10K unique emails) found:

1. **The collision-signal metric isn't discriminative.** T3's collision_rate is **0.59**; T2's is **0.62**. T3 — the adversarial dataset — is LESS collision-prone by this metric than T2. Any threshold that fires on T3 also fires on T2, where it's a false positive (186 FNs added).
2. **Iteration budget exhausted upstream.** `rule_blocking_too_coarse` oscillates between `last_name` and `email` blocking, never converging. The committed config falls back to v0 (iter=-1) — `rule_demote_clustered_identity` at position 7 doesn't get to run inside the iteration loop on real T3.
3. **T3's adversarial FPs come from `exact_email` directly.** Even if the demote rule fired, it would demote to a weighted matchkey where NE applies. But on real T3 the demote rule doesn't fire, so the existing exact_email matchkey continues capturing collision pairs — and `_apply_negative_evidence` doesn't apply to exact matchkeys per spec §Non-goals.

## Benchmarks (zero-config, no LLM)

| Dataset | v1.10.0 | v1.11.0 | Δ |
|---|---|---|---|
| DBLP-ACM | 0.9641 | 0.9641 | flat |
| Febrl3 | 0.9443 | 0.9443 | flat |
| NCVR | 0.9719 | 0.9719 | flat |
| DQbench composite | 66.91 | 66.99 | +0.08 (noise) |
| T1 F1 | 88.9% | 88.9% | flat |
| T2 F1 | 69.0% | 69.0% | flat |
| T3 F1 | 53.8% | 53.8% | flat |

## v1.12 follow-on

The right fix for T3 is **Path Y from the original v1.11 spec** — extend `_apply_negative_evidence` to also apply to exact matchkeys, subtracting from the binary 1.0 so that FP pairs (same email, very different phone+address) drop below threshold. The spec rejected this initially as semantic confusion; the diagnostic now proves Path X (demote then NE) cannot reach T3 because the discriminating signal isn't strong enough at the collision-signal level.

v1.12 spec to be drafted with Path Y + a measurement-driven threshold for the exact-matchkey NE subtract.

## Test plan

- [x] 1986 tests pass (1907 v1.10 baseline + ~79 new across infrastructure + tests)
- [x] DBLP-ACM/Febrl3/NCVR re-measured post-v1.11, F1 identical to v1.10
- [x] DQbench no-LLM re-measured, composite 66.99 (parity with v1.10)
- [x] T3 synthetic Tier 4 regression test passes (mechanism works on data where it can fire)
- [x] T3 clean-compat test passes (rule does not over-apply on clean data)
- [x] v1.10 cache backward-compat verified
- [ ] CI green
- [ ] Reviewer pass

## Spec + plan

- Spec: \`docs/superpowers/specs/2026-05-08-autoconfig-negative-evidence-and-clustered-identity-design.md\`
- Plan: \`docs/superpowers/plans/2026-05-08-autoconfig-negative-evidence-and-clustered-identity.md\`
- T3 diagnostic: \`.profile_tmp/v111_t3_diagnostic.txt\` (gitignored)


benzsevern:
Superseded by v1.12 release PR. The same branch now contains both v1.11 (foundation: NE schema + scoring + clustered-identity guard) and v1.12 (Path Y: NE on exact matchkeys delivering DQbench composite 66.91 → 91.04). Combined release narrative is cleaner as a single PR — see incoming.

PR #216, opened 2026-05-13
feat(refdata): bundle US Census surnames + name_freq_weighted_jw scorer

benzsevern:
## Summary

First slice of **strategy direction #8** (`docs/superpowers/specs/2026-05-08-competitive-strategy-review.md`): bundle public-domain reference data the engine can consume to lift accuracy on people-shape matching.

- New `goldenmatch.refdata` package: surname lookup API + `name_freq_weighted_jw` scorer, registered via the plugin system on import.
- Bundled **US Census 2010 top-10K surnames** (~176 KB, public domain) with provenance, license, and regenerate command in `data/PROVENANCE.md`.
- Scorer weights surname matches by IDF **only in the borderline JW zone [0.70, 0.95]**; exact / near-exact / clear-non-match pairs pass through plain JW so recall on confident matches is preserved. OOV on either side also falls back to plain JW (refuses to up-credit typos of common names).
- `core/scorer.py::_fuzzy_score_matrix` now falls through to `PluginRegistry` for unknown scorer names — keeps the contract uniform across pair-by-pair (`score_field`) and NxN (`_fuzzy_score_matrix`) paths.

## Measured impact

NCVR 10K corrupted-duplicates GT, last_name scorer swapped:

| | Precision | Recall | F1 |
| - | - | - | - |
| baseline (zero-config, JW) | 0.9820 | 0.9624 | 0.9721 |
| refdata (last_name → name_freq_weighted_jw) | 0.9820 | 0.9624 | 0.9721 |

**No regression, no lift.** NCVR's heavy-corruption distribution puts few pairs in the borderline JW zone where the weighting acts. The infrastructure is shipped; demonstrating positive lift needs an enterprise-shape benchmark per direction #5 (where common-name false-positives are the dominant error mode).

## What's deferred

- **Auto-config integration.** Controller doesn't yet pick `name_freq_weighted_jw` automatically — users opt in via explicit config.
- **`reference-business` and `reference-address` packs** (Inc/LLC normalization, libpostal, etc.) per direction #8's full Ship list.
- **Per-scorer threshold tuning** in `LearningMemory` — the current 0.85ish threshold was auto-tuned for plain JW; a re-tuned threshold against the weighted scorer is the next lever.
- **Enterprise-shape benchmark** per direction #5 — needed to demonstrate positive F1 lift from this scorer, since NCVR's corruption distribution doesn't exercise the borderline zone heavily.

## Test plan

- [x] `pytest tests/test_refdata_surnames.py` — 21 tests, all green: lookup correctness, IDF monotonicity, scorer borderline behavior, OOV pass-through, plugin registration, `MatchkeyField` validator accepts the new scorer.
- [x] `pytest tests/test_scorer.py tests/test_config.py tests/test_refdata_surnames.py` — 129 tests, all green (no regressions in adjacent modules).
- [x] `python tests/benchmarks/run_ncvr_refdata.py` — A/B run reproducible; numbers above.
- [ ] Full suite on CI (`pytest --tb=short`).
- [ ] Verify wheel includes the data file (`python -m build && unzip -l dist/*.whl | grep refdata`).


benzsevern:
Review fixes applied (commit 2d7dcec):

- **Vectorized score_matrix** for `NameFreqWeightedJW` — replaces the O(N²) Python double-loop with one rapidfuzz cdist + numpy elementwise ops. `core/scorer.py::_fuzzy_score_matrix` picks up the `score_matrix` method automatically; plugins without it still work via the score_pair loop but emit a WARNING above N=1000 so the perf landmine is visible.
- **OOV gate reordered** in score_pair: `surname_rank` now runs before `surname_idf` so the OOV branch short-circuits instead of incidentally getting the right answer via `surname_idf == 1.0`.
- 4 new tests in test_refdata_surnames.py covering the score_matrix method, matrix↔pair equivalence across in-zone/out-of-zone/OOV cases, empty-input handling, and the OOV pass-through invariant.
- CHANGELOG updated to describe the vectorized hot-path.

PR #217, opened 2026-05-13
feat(refdata): given-name aliases + given_name_aliased_jw scorer (direction #8, slice 2)

benzsevern:
## Summary

Second slice of **strategy direction #8**'s `reference-people` pack — adds nickname-equivalence to first-name matching: William ↔ Bill, Robert ↔ Bob, Margaret ↔ Peggy, plus ~140 other canonical English given names with their common short forms.

**Stacked on top of #216 (`feature/refdata-bundle`)** — when that lands, this PR's base will rebase onto main cleanly.

- `goldenmatch/refdata/data/given_name_aliases.json` — curated public-knowledge alias table.
- `goldenmatch.refdata.given_names`: `canonical_form` / `aliases_of` / `are_equivalent` / `is_available`. Symmetric and transitive within equivalence classes. Handles multi-canonical names (e.g. "Jack" appears both as alias to "John" AND as its own canonical).
- `given_name_aliased_jw` scorer: alias-equivalent pairs return 1.0; everything else falls back to plain Jaro–Winkler. **Never *lowers* a JW score** — only promotes known aliases. Degrades cleanly when the alias table is missing.
- 23 unit tests + synthetic nickname-shape A/B benchmark.

## Measured impact

Synthetic, 1000-record fixture, 200 nickname-pair GT + 600 distractors (random non-aliased first names, isolated unique surnames), threshold 0.95:

| | TP | FP | FN | P | R | F1 |
| - | - | - | - | - | - | - |
| baseline (plain JW) | 0 | 0 | 200 | 0.0 | 0.0 | **0.0000** |
| refdata (`given_name_aliased_jw`) | 176 | 0 | 24 | 1.0000 | 0.8800 | **0.9362** |

**F1 delta: +0.9362.** The plain-JW baseline can't bridge the nickname gap at all (JW(William, Bill) ~= 0.55, far below threshold). The aliased scorer catches them at 1.0.

This fixture is the abridged version of direction #5's enterprise-shape benchmark — needed because NCVR's heavy-corruption distribution doesn't exercise nicknames at all (corruptions are typos/drops/abbreviations to single letters). A common-name-FP fixture for the surname scorer (PR #216) is the natural next sibling.

## Known follow-up

The 12% recall ceiling (24 missed pairs out of 200) is mysterious — every missed pair clusters correctly when isolated to its own 2-record dataframe, but the same pairs get skipped in the full 1000-record fixture. Doesn't reproduce in subsets of <=12 records. Needs `find_fuzzy_matches` investigation; not a correctness bug in the scorer or lookup. Flagging in CHANGELOG.

## Test plan

- [x] `pytest tests/test_refdata_given_names.py tests/test_refdata_surnames.py` — 44 tests, all green.
- [x] `python tests/benchmarks/run_nickname_synth.py` — reproducible numbers above.
- [ ] Full suite on CI.
- [ ] Investigate the 12% recall ceiling.


benzsevern:
**Recall ceiling solved — F1 0.9362 → 1.0000 on the synthetic benchmark.**

Root cause: `are_equivalent` was silently asymmetric for short forms that belong to multiple canonicals.

- "kate" appears in 3 canonicals (Catherine, Kathleen, Kaitlyn). The original loader stored one canonical per form (last-writer-wins), so `canonical["kate"]` ended up pointing at Kaitlyn.
- `are_equivalent("Kate", "Catherine")` looked up Kate→Kaitlyn, Catherine→Catherine, found no shared canonical, returned **False**.
- `are_equivalent("Catherine", "Kate")` looked up Catherine→Catherine, Kate→Kaitlyn — but then fell through to the multi-canonical branch (`nb in class_a`), found "kate" in `classes["catherine"]`, returned **True**.
- `find_fuzzy_matches` reads only the upper triangle of the NxN score matrix, so the False direction was the one being consulted. Every Catherine/Kate, Christopher/Chris, Susan/Sue, Patricia/Pat pair was being dropped silently.

Why isolated 2-record tests passed: alphabetic dict-iteration ordering happens to put "catherine" before the other parents in some Python invocations, making the surviving canonical the one being asked about. Brittle.

Fix: each form stores the full **set** of canonicals it belongs to. `are_equivalent` returns True iff the two forms share at least one canonical. Symmetric by construction. Regression test added: `test_are_equivalent_symmetric_for_ambiguous_short_forms`.

Updated benchmark numbers:

| | TP | FP | FN | P | R | F1 |
| - | - | - | - | - | - | - |
| baseline (plain JW) | 0 | 0 | 200 | 0.0 | 0.0 | **0.0000** |
| refdata (`given_name_aliased_jw`) | 200 | 0 | 0 | 1.0000 | 1.0000 | **1.0000** |

CHANGELOG updated to reflect the new numbers and document the bug.

benzsevern:
Review fixes applied (commit 928bc62):

- **Duplicate `anthony` key** in given_name_aliases.json merged into one entry (lines 18-19 → single line with full alias list).
- **Within-list `rich` duplicate** in richard's alias list removed.
- **`object_pairs_hook` in _load**: surfaces a WARNING if any canonical key appears twice in the source JSON — silent last-wins behavior preserved, but the silent loss is no longer invisible.
- **test_no_duplicate_canonical_keys_in_bundled_json**: regex-walks the raw JSON text before json.load to fail loudly on any future duplicate key.

PR #219, opened 2026-05-13
feat(refdata): reference-address pack with address_normalize transform (direction #8, slice 4)

benzsevern:
## Summary

**Fourth slice** of strategy direction #8. Opens the `reference-address` pack with a USPS-style token-normalization transform: collapses Pub. 28 street-suffix, directional, and secondary-unit variants to canonical short forms so "123 Main Street North Apartment 5" and "123 Main St N Apt 5" both reduce to "123 main st n apt 5" before scoring.

**Stacked on PR #218** (`feature/refdata-business`). Order: #216 → #217 → #218 → #219.

- `goldenmatch/refdata/data/address_abbreviations.json` — ~500 surface variants from USPS Publication 28 Appendix C. Public-domain US federal data; no license restrictions.
- `goldenmatch.refdata.addresses.normalize_address` — tokenize on whitespace + commas, lowercase, strip punctuation, map every recognised variant to its USPS canonical short form. Unknown tokens pass through.
- **Position-agnostic by design.** Every USPS-known token is normalized, not just trailing ones. Trade-off: words that are both name parts and suffix variants ("Lake" → "lk", "Court" → "ct", "Park" → "park") collapse along with true suffixes. Match invariance is preserved as long as both sides reduce equally — pinned by `test_aggressive_normalization_preserves_match_invariance`. This transform is matching-only, not for human-readable display.
- 48 unit tests + synthetic address A/B benchmark.

## Measured impact

Synthetic, 1000-record fixture, 200 same-street pairs differing in suffix abbreviation (some also in directional/unit) + 600 distractors, threshold 0.95:

| | TP | FP | FN | P | R | F1 |
| - | - | - | - | - | - | - |
| baseline (no transform) | 116 | 0 | 84 | 1.0000 | 0.5800 | **0.7342** |
| baseline (lowercase only) | 114 | 0 | 86 | 1.0000 | 0.5700 | **0.7261** |
| refdata (`address_normalize`) | 200 | 0 | 0 | 1.0000 | 1.0000 | **1.0000** |

**F1 delta +0.2658. Recall +0.4200.** Plain JW catches the small suffix deltas (Ave/Avenue, St/Street) but misses larger ones (Boulevard/Blvd, Northeast/NE, Apartment/Apt). The transform catches everything.

## What's deferred

Inside this pack:
- libpostal binding (heavy C deps + ~2 GB model; belongs in an opt-in `[reference-address-postal]` extra)
- Street-name canonicalization (full USPS CASS proper)
[...]

PR #221, opened 2026-05-13
feat(refdata): auto-config integration for all four refdata packs (direction #8, slice 6)

benzsevern:
## Summary

**Sixth slice** of strategy direction #8. Zero-config controller now picks up the refdata scorers/transforms automatically based on column-name shape. No YAML config needed — `dedupe_df(df)` does the right thing when a column is named `last_name` / `first_name` / `company_name` / `address`.

**Stacked on PR #220** (`feature/refdata-surname-fp-bench`). Order: #216 → #217 → #218 → #219 → #220 → #221.

## Refinement rules

Each gated on the relevant pack's `is_available()` — non-refdata installs behave exactly as before.

| Column-name pattern | Refinement |
|---|---|
| `last_name`, `surname`, `lname`, `family_name`, `last` | scorer → `name_freq_weighted_jw` |
| `first_name`, `given_name`, `fname`, `forename`, `first` | scorer → `given_name_aliased_jw` |
| `company`, `business`, `org`, `firm`, `employer`, `legal_name`, `entity_name` | prepend `legal_form_strip` transform |
| `address`, `street`, `addr_line`, `mailing_address` | prepend `address_normalize` transform |

## Implementation

- New module `goldenmatch/refdata/autoconfig_hooks.py` exposes a single pure function:
  ```python
  def refine_matchkey_field(column_name: str, scorer: str, transforms: list[str]) -> tuple[str, list[str]]:
      ...
  ```
- Wired into `core/autoconfig.py` at **both** matchkey-building paths:
  - `build_matchkeys()` (vanilla weighted/exact)
  - `build_probabilistic_matchkeys()` (Fellegi-Sunter)
- Hook fires **after** `_SCORER_MAP[col_type]` resolves the col-type defaults but **before** `MatchkeyField` is constructed — so the existing cardinality guards, exact-matchkey skips, and column-type classification all run unchanged.
- **Scorer-swap protection**: only string-similarity scorers (`jaro_winkler`, `levenshtein`, `token_sort`, `ensemble`, `dice`, `jaccard`) get swapped. Exact and embedding pass through — preserves identity-field semantics.
- **Transform prepend, not replace**: `legal_form_strip` runs *before* the existing `lowercase`/`strip` chain, so canonical short forms still go through downstream normalization. Idempotent — won't double-prepend.
- **Compound columns** (e.g. `company_last_name`) get both refinements: scorer swap from the last_name match + transform prepend from the company match.

## Tests

`tests/test_refdata_autoconfig.py` (~50 cases):
- Parametrized across every column-name variant per refinement rule.
- Exact scorer is preserved (not swapped to refdata variant).
- Idempotency: re-running refinement on an already-refined field is a no-op.
- No-mutation: the caller's transforms list is never modified in place.
- Compound-column composition.
- Two end-to-end tests via `build_matchkeys()` confirming the hook fires in the real autoconfig path.

## In-session validation blocked

Same **Polars DLL hang** documented in PR #220 / CLAUDE.md blocked every Python invocation in this session, including a no-pytest direct validation script. The refinement function is pure synchronous Python; **rerun on a fresh Python boot**:

```bash
PYTHONPATH=. python -m pytest tests/test_refdata_autoconfig.py -v
```

Three of the test cases (`test_build_matchkeys_*`, `test_build_matchkeys_prepends_*`) touch Polars and need the same fresh-boot rerun.

After the unit tests pass, the meaningful follow-up regression check is:

```bash
PYTHONPATH=. python -m pytest tests/test_autoconfig_benchmarks.py -v
```

NCVR F1 should be ≥ 0.97 (the slice 1 + slice 2 fixtures both apply through this auto-config path). If NCVR regresses, the refdata scorers are doing something at threshold-tuning we didn't predict — the obvious mitigation is to gate the swap on a `cardinality_ratio` / `null_rate` floor.

## What's still deferred

- Per-scorer threshold tuning in `LearningMemory` — the existing 0.85ish threshold was auto-tuned for plain JW. The refdata scorers may want their own threshold, but that's a `LearningMemory` change, not an autoconfig change.
- Controller-level rule that A/B-tests refdata vs vanilla in its iteration loop. The current integration commits to refdata unconditionally when the column-name pattern matches; a controller rule could measure F1 / mass-separation across the swap and pick the winner.
- Industry codes (NAICS) and OpenCorporates company-name variants for `reference-business`.
- libpostal binding as opt-in `[reference-address-postal]` extra.

## Test plan

- [ ] `pytest tests/test_refdata_autoconfig.py -v` on fresh boot — should be ~50 green.
- [ ] `pytest tests/test_refdata_*.py` (all five refdata files) on fresh boot — ~187 green including slice 5.
- [ ] `pytest tests/test_autoconfig_benchmarks.py` — NCVR / DBLP-ACM ≥ existing floors.
- [ ] `python tests/benchmarks/run_nickname_synth.py` / `run_business_synth.py` / `run_address_synth.py` — F1 should hit 1.0 even without explicit `transforms:` config (the new auto-config path picks them up).

### DIFF
--- a/packages/python/goldenmatch/CHANGELOG.md
+++ b/packages/python/goldenmatch/CHANGELOG.md
@@ -6,6 +6,24 @@ Format follows [Keep a Changelog](https://keepachangelog.com/). Versioning follo
 
 ## [Unreleased]
 
+### Added -- Auto-config integration for refdata packs (strategy direction #8, sixth slice)
+
+Wires all four refdata packs into the zero-config controller. Auto-config no longer needs an explicit YAML to pick up surname-IDF weighting, given-name aliasing, legal-form stripping, or USPS address normalization — it picks them automatically when column names signal the relevant shape.
+
+- **New module** `goldenmatch/refdata/autoconfig_hooks.py` exposes `refine_matchkey_field(column_name, scorer, transforms) -> (scorer, transforms)`. Pure function; safe to call on every column unconditionally.
+- **Refinement rules** (each gated on the relevant pack's `is_available()` — non-refdata installs behave exactly as before):
+  - `last_name | surname | lname | family_name` → scorer becomes `name_freq_weighted_jw`.
+  - `first_name | given_name | fname | forename` → scorer becomes `given_name_aliased_jw`.
+  - `company | business | org | firm | employer | legal_name | entity_name` → `legal_form_strip` is prepended to the transforms list.
+  - `address | street | addr_line | mailing_address` → `address_normalize` is prepended.
+- **Wired into `core/autoconfig.py`** at both `build_matchkeys()` (vanilla weighted/exact path) and `build_probabilistic_matchkeys()` (Fellegi-Sunter path). The hook fires *after* `_SCORER_MAP[col_type]` resolves but *before* `MatchkeyField` is constructed — so the existing column-type classification, cardinality guards, and exact-matchkey skips still run unchanged.
+- **Scorer-swap protection**: only string-similarity scorers (`jaro_winkler`, `levenshtein`, `token_sort`, `ensemble`, `dice`, `jaccard`) get swapped. Exact and embedding scorers pass through — preserves identity-field semantics.
+- **Transform prepend, not replace**: `legal_form_strip` runs before any existing `lowercase`/`strip`, so the canonical short form still goes through downstream normalization. Idempotent — the function won't double-prepend if the transform is already in the list.
+- **Compound columns** (e.g. `company_last_name`) get both refinements: scorer swap from the last_name match + transform prepend from the company match.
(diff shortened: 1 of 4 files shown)


benzsevern:
Review fix applied (commit f708787):

- **Hoisted `refine_matchkey_field` import** to module-top in core/autoconfig.py with a try/except Exception fallback. Per-iteration call sites now use a module-level binding (no `from X import Y` cost in the per-column loop).
- **Widened the catch** from `ImportError` to `Exception`: a corrupt data file or broken transitive import in any refdata pack used to propagate up through build_matchkeys and break dedupe_df(df) for every user, even ones who never named a column `last_name`. The fallback now wires a no-op pass-through stub when refdata is unhealthy for ANY reason.

Deferred to cross-pack follow-up: gating the scorer swap on `ColumnProfile.col_type == "name"` so the autoconfig hook only fires on data-shape-confirmed columns. That's a more substantial change (hook signature grows + autoconfig.py threading) and belongs in its own PR.

PR #222, opened 2026-05-13
feat(refdata): NAICS 2022 industry-code normalization (direction #8, slice 7)

benzsevern:
## Summary

**Seventh slice** of strategy direction #8. Extends `reference-business` with the second documented entry from the ship list (after legal-form strip): NAICS industry classification with bidirectional code-title lookup and a normalizing transform.

**Stacked on PR #221** (`feature/refdata-autoconfig`). Order: #216 → #217 → #218 → #219 → #220 → #221 → #222.

- **Bundled NAICS 2022** at `goldenmatch/refdata/data/naics_2022.json` (~120 KB) — 2,125 entries across all five hierarchy levels (17 sectors, 96 subsectors, 308 industry groups, 692 5-digit industries, 1,012 6-digit US industries). Sourced from the U.S. Census Bureau's [2-6 digit 2022 Codes file](https://www.census.gov/naics/2022NAICS/2-6%20digit_2022_Codes.xlsx). Public-domain US federal data; no license restrictions. The "31-33" range-encoded Manufacturing sector is expanded across each constituent 2-digit code.
- **API**:
  - `title_for_code(code)`, `code_for_title(title)` — bidirectional lookups. Case- and punctuation-tolerant.
  - `naics_normalize(value)` — the transform.
  - `industries_available()`, `known_codes()`, `known_titles()` — diagnostics.
- **`naics_normalize` transform** auto-registered via `PluginRegistry` on `import goldenmatch.refdata`. Three input shapes:
  1. **Numeric codes** (with optional separators / trailing text): "511210", "511 210", "511210 (Software Publishing)" → 6-digit-truncated canonical when it resolves to a known code; otherwise falls back to the longest known prefix; otherwise the truncated digits as-is (so two records sharing an unknown code still match each other).
  2. **Known industry titles**: "Information" → "51", "Software Publishers" → "513210" (the 2022 6-digit code) — case- and punctuation-tolerant.
  3. **Everything else**: lowercase + whitespace-collapse pass-through. Never raises.
- **Autoconfig hook extended** in `autoconfig_hooks.py`: column-name patterns `naics`, `sic`, `industry`, `industry_code`, `industry_classification`, `business_type` → `naics_normalize` is prepended to the transforms list. Mirrors the existing `legal_form_strip` and `address_normalize` handling.

## Test coverage

`tests/test_refdata_industries.py` — ~30 cases:
- Bidirectional title↔code round-trips (sector, subsector, 6-digit).
- Case-insensitive + punctuation-tolerant title lookup.
- Separator-tolerant code parsing (`"111 110"`, `"111,110"`, `"111110 (Soybean Farming)"`).
- Overlong-code truncation to first 6.
- Unknown-code prefix fallback.
- Unknown-numeric still returns 6-digit truncated digits (so two records sharing an unknown code still match).
- Transform plugin registration + dispatch through `apply_transform`.
- Transform-chain composition with `lowercase`/`strip`.
- `FieldTransform` validator accepts the plugin name.
- `MatchkeyField` accepts `naics_normalize` in `transforms:`.
- 11 column-name variants for the autoconfig hook prepend.
- Idempotency: re-applying refinement to an already-refined transforms list is a no-op.

## In-session validation blocked

Same **Polars DLL hang** as PRs #220 and #221: `goldenmatch/__init__.py` eagerly imports polars-heavy modules, which poisons every `goldenmatch.*` import including refdata submodules. Killed and respawned the Python process several times; the hang persists across cycles in this session.

A clean test of the NAICS data extraction itself worked fine (using openpyxl directly without any `goldenmatch.*` import), so the bundled JSON is verified. The transform implementation is pure synchronous regex + dict lookup — no hot loops, no IO at runtime.

**Rerun on a fresh Python boot**:

```bash
pytest tests/test_refdata_industries.py -v
```

## What's still deferred (after slice 7)

- **OpenCorporates company-name variants** — the last documented `reference-business` extension. Variant table (e.g., "Acme Corporation" ↔ "ACME CORP" via OpenCorporates IDs) would help cross-source business matching.
- **libpostal binding** as opt-in `[reference-address-postal]` extra.
- **Per-scorer threshold tuning** in `LearningMemory`.
- **Controller-level A/B rule** that would measure refdata-on vs refdata-off in the iteration loop instead of applying refinements unconditionally on column-name match.
- **NCVR / DBLP-ACM regression rerun** under the new auto-config path.

## Test plan

- [ ] `pytest tests/test_refdata_industries.py -v` — ~30 green on a fresh boot.
- [ ] `pytest tests/test_refdata_*.py` — all five refdata test files green (~217 cases).
- [ ] Verify wheel includes the bundled JSON (`python -m build && unzip -l dist/*.whl | grep naics`).


benzsevern-mjh:
## Review fixes applied (commit d3261b5)

Addressed the critical + important findings from the parallel review.

**Fixed:**
- `naics_normalize` now scans every 2+-digit run in the input (not just the first), so `"NAICS 2022 code 511210"` returns `513210` instead of failing on the vintage-year prefix. Docstring updated to match — previously omitted the longest-prefix fallback entirely.
- `_COMPANY_NAME_RE` tightened with `business(?!.?type)` so `business_type` / `businesstype` / `business type` only get `naics_normalize` prepended, not `legal_form_strip` + `naics_normalize`. Functionally harmless before (legal forms aren't in NAICS codes), but burned a transform call per row and contradicted the autoconfig_hooks module docstring.
- `_reload()` no longer wipes the state dicts before re-parsing — readers now see either the old dict or the new one, never an empty in-between. `_load()` under lock + new-dict-assignment provides the atomicity.

**Tests added:**
- Longest-known-prefix fallback success path (the uncovered branch the reviewer flagged).
- Title-precedence narrowest-wins (the uncovered invariant at `industries.py:103-107`).
- Multi-digit-run scanning (vintage-year regression).
- `business_type` does NOT get `legal_form_strip` after the regex tightening (regression).
- `_COMPANY_NAME_RE` still matches its original cases (`business_name`, `BusinessName`, `company_name`, `legal_name`, `entity_name`, `firm`, `company`) after the tightening — no false negatives.
- All `business[_ ]?type` regex branches (`businesstype`, `business type`, `business_type`, `BusinessType`) match `_INDUSTRY_RE`.

Test count: ~30 → ~40.

**Verified, not fixed (review false-flag):**
The CHANGELOG entry counts ("1,012 6-digit US industries", "2,125 total") are accurate against the bundled JSON. The 2-6 digit Census file DOES include 6-digit US-detail rows — the reviewer was wrong about the source structure. JSON verified via: `python -c "import json; print({k: len(v) for k, v in json.load(open('.../naics_2022.json')).items() if not k.startswith('_')})"` → `{'sectors_2digit': 17, 'subsectors_3digit': 96, 'industry_groups_4digit': 308, 'industries_5digit': 692, 'industries_6digit': 1012}`.

**Deferred to a cross-pack follow-up PR:**
- Type-design improvements: `TransformPlugin` Protocol in `goldenmatch.plugins`, `_RefdataState` frozen dataclass replacing `_state: dict`. These apply to all five refdata modules (surnames, given_names, business, addresses, industries), so doing them in a single coordinated PR — not slice 7.
- Comment-trim pass (drop the WHAT-not-WHY comments flagged by the comment analyzer in `_norm_title`, `_load`'s section block, etc.). Same reasoning — applies across all five packs.
- Per-pack data-missing degradation tests. None of the existing refdata test files cover this; raising the floor across all five is a separate, focused PR.

**Same Polars-DLL-hang caveat applies** to in-session test verification — the regex-only test cases were verified by direct regex evaluation in PowerShell; the rest run on a fresh Python boot via `pytest tests/test_refdata_industries.py -v`.

PR #567, opened 2026-05-29
perf(goldencheck): cache n_unique in scanner column-profile loop (~60s on 10M QIS)

benzsevern:
## Why

10M QIS bench v17 measured pipeline_prep_quality_scan at 120.8s, identical to v16 even after PR #562 (scan_dataframe entry point) removed the CSV round-trip. The CSV write/read was not the bottleneck.

Read the scanner code: the column-profile loop calls `non_null.n_unique()` TWICE per column on the FULL DataFrame (lines 249-250) to populate `unique_count` and `unique_pct` on the same ColumnProfile dataclass kwarg list. At 10M rows across 6 columns that is 12 full-frame Polars hash-aggregations -- the actual hot path.

## What

Cache `n_unique` and `null_count` per column (each called twice). Strict 2x reduction on the loop body; zero behavioral change.

## Expected impact

10M-v17: pipeline_prep_quality_scan 120.8s -> ~60s (cuts the n_unique pass in half). Combined with #562 (already merged on PR branch chain) the net wall recovered on this stage is ~60-70s.

This is the real fix; #562's I/O removal is preserved as a smaller secondary improvement (10-20s on top).

## Why this didn't surface before

The previous scan_dataframe diagnosis assumed the CSV round-trip dominated. v17's measurement disproved that and pointed at the in-scanner hash-agg. Trust the measurement.

### DIFF
--- a/packages/python/goldencheck/goldencheck/engine/scanner.py
+++ b/packages/python/goldencheck/goldencheck/engine/scanner.py
@@ -33,7 +33,7 @@ from goldencheck.relations.temporal import TemporalOrderProfiler
 
 logger = logging.getLogger(__name__)
 
-__all__ = ["scan_file", "scan_file_with_llm"]
+__all__ = ["scan_file", "scan_file_with_llm", "scan_dataframe"]
 
 COLUMN_PROFILERS = [
     TypeInferenceProfiler(),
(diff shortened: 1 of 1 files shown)


benzsevern:
Closing -- the cache n_unique fix didn't make it into the commit (only the duplicate scan_dataframe patch from #562 ended up here). Will re-open as a fresh PR with the correct diff.

PR #609, opened 2026-05-30
feat(ray): fetch GCP creds from Infisical at runtime (Option B)

benzsevern:
Replaces GCP_PROJECT_ID + GCP_SA_KEY GH-Actions secrets with a runtime Infisical fetch. Only INFISICAL_CLIENT_ID + INFISICAL_CLIENT_SECRET need to live in GH; everything else stays in Infisical (project a99885f0, env=dev). Future Infisical-backed secrets reuse the same auth pair instead of accumulating GH copies. Spec context: Option B from the 2026-05-30 Ray-GCE conversation.

### DIFF
--- a/.github/workflows/bench-ray-cluster.yml
+++ b/.github/workflows/bench-ray-cluster.yml
@@ -5,14 +5,25 @@
 # config (1 head + 3 preemptible workers). The tear-down step runs in
 # `if: always()` so a failed bench still releases the instances.
 #
-# Required GitHub secrets (provision once per repo, see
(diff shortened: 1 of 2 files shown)


benzsevern:
Closing in favor of Option A (sync Infisical -> GH Actions secrets directly) for this session's bench dispatches. The Infisical-fetch-at-runtime design is correct long-term -- we'll come back to it. The workflow currently on main (#608) still expects GCP_PROJECT_ID + GCP_SA_KEY as GH-Actions secrets; those are about to be synced manually.

PR #653, opened 2026-06-01
perf(cluster): columnar pipeline wiring — Phase 0 (harness) + Phase A (gated fast-path)

benzsevern:
## Summary

Executes the agreed direction from the 1M `profile-hotspots` analysis — "scope wiring the columnar path into the default pipeline, starting with a design note + parity harness, not a blind swap." This is **Phase 0**: the corrected analysis, the parity contract, and an executable harness. **No production code path changes; default behavior is unchanged.**

## The 1M finding (and a corrected lever attribution)

| target | 1M wall (cProfile) | 1M wall (pyinstrument) |
|---|---:|---:|
| **columnar** | **359 s** | **353 s** |
| list (baseline) | 575 s | 575 s |

The columnar path is **~38% faster at 1M**. But reading `core/cluster.py` shows `build_clusters_columnar` is a **thin wrapper** — it converts the df back to a list (`_pairs_df_to_list_numpy`) and calls the *same* `build_clusters`. So the cluster algorithm is identical in both paths, and **the 38% win is the columnar scorer's direct-DataFrame emit (#634/#639) over 131M pairs, not the cluster build.** Neither the columnar scorer nor the columnar cluster path is wired into the default pipeline today (it uses `score_blocks_parallel` → `build_clusters` on a list).

## Why this is a phased refactor, not a swap

`_run_dedupe_pipeline` aggregates `all_pairs` as one Python list across **exact + fuzzy + probabilistic** scorers, and `_apply_postflight` consumes that list (signals + threshold filter) when auto-config ran. The win is inseparable from staying columnar end-to-end (converting back to a list re-incurs the 131M-tuple build). So a correct wiring touches the scorer aggregation + postflight, not just the cluster call — the multi-phase Arrow roadmap.

## What's in this PR (Phase 0)

- **`docs/columnar-pipeline-wiring.md`** — corrected analysis, the exact coupling points, the parity contract, and the phased plan:
  - **Phase A** — gated (`GOLDENMATCH_COLUMNAR_PIPELINE`, default-off) end-to-end columnar fast-path for the *eligible* shape (single fuzzy matchkey, no exact/probabilistic, no auto-config postflight) — captures the ~38% win for the common case with zero risk to existing paths.
  - **Phase B** df-native postflight · **Phase C** multi-scorer columnar aggregation · **Phase D** native Arrow-C `build_clusters` / `ClusterFrames` kernel for the (separate) ~164 s cluster-build lever.
- **`tests/test_columnar_pipeline_parity.py`** — the executable gate every phase must pass: `score_blocks_columnar → build_clusters_columnar` yields clusters whose membership partition, `pair_scores`, `size`, `oversized`, and `confidence` are **identical** to the list path, at n ∈ {500, 2000, 5000}, with/without `auto_split`, plus the empty-pair case.

## Test plan
- [x] `tests/test_columnar_pipeline_parity.py` — **7 passed** (parity holds; also confirmed identical at n=10000 ad hoc)
- [x] ruff clean
- [x] No production code path touched — pure docs + test

## Next
Phase A (the gated in-pipeline fast-path) is the next increment, landing as its own focused, benched change now that the parity gate is in place. Flagging that I held it out of this PR deliberately rather than jam a swap into the intricate scoring section.

---
_Generated by [Claude Code](https://claude.ai/code/session_01WiKj934B5YrmFGUMoMVok5)_

### DIFF
--- a/docs/columnar-pipeline-wiring.md
+++ b/docs/columnar-pipeline-wiring.md
@@ -1,0 +1,101 @@
+# Wiring the columnar pair-stream path into the default pipeline
+
+Status: **Phase 0 (analysis + parity harness)**. This note records the 2026-06-01
+`profile-hotspots` finding, corrects the lever attribution, and lays out a phased,
+parity-gated plan to route the default dedupe pipeline through the columnar
+pair-stream path. No default behavior changes in Phase 0.
+
+## What the 1M profile measured
+
+`profile-hotspots` (n=1,000,000, 131,166,381 pairs, `realistic_person` fixture):
+
+| target | wall (cProfile) | wall (pyinstrument) |
+|---|---:|---:|
+| **columnar** (`score_blocks_columnar` -> `build_clusters_columnar`) | **359 s** | **353 s** |
+| list (`score_blocks_parallel` -> `build_clusters`) | 575 s | 575 s |
+
+The columnar path is ~38% faster at 1M (216 s saved). The cumtime top-10 is
+dominated by `threading.join` / `_thread.lock.acquire` (~390 s) -- that is the
+`ThreadPoolExecutor` teardown wait wrapping the parallel rapidfuzz scorer, i.e.
+genuine GIL-released compute, not Python overhead. Stripping it, the timed body
+splits roughly:
+
+- `score_blocks_columnar` ~195 s (~54%) -- fuzzy scoring of 131M pairs.
+- `build_clusters_columnar` ~164 s (~46%) -- by subtraction.
+
+## Corrected lever attribution
+
+A reading of `core/cluster.py` shows **`build_clusters_columnar` is a thin wrapper**:
+it converts the pair DataFrame to a list via `_pairs_df_to_list_numpy` and calls the
+same `build_clusters`. The cluster *algorithm* (Union-Find + the `pair_scores` dict
+fill + confidence + MST split) is therefore **identical** between the list and
+columnar paths.
+
+So the measured 38% win is **the scorer**, not the cluster build:
+
+- `score_blocks_parallel` builds a Python `list[tuple]` of 131M pairs (per-pair tuple
+  append). `score_blocks_columnar` (#634 direct-DataFrame emit, #639 native columnar
+  inner loop) emits the pair stream columnar. At 131M pairs the list construction is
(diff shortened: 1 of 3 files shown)


benzsevern:
## Phase A added (`9d86170`) — the gated fast-path is now wired

Following Phase 0 (design note + parity harness), this commit wires the columnar path into `_run_dedupe_pipeline` behind **`GOLDENMATCH_COLUMNAR_PIPELINE` (default OFF)**.

**What it does.** When the gate is on *and* the config is eligible — single weighted matchkey, no exact/probabilistic, no auto-config postflight, no rerank/LLM/boost, default backend (not ray/duckdb/bucket/chunked/datafusion), not `across_files_only`, scorers in a columnar-safe set — the fuzzy scoring step emits a pair DataFrame via `score_blocks_columnar` and the cluster step consumes it via `build_clusters_columnar`, skipping the Python list aggregation (the ~38% lever at 1M). Any error in the columnar branch **falls back to the list scorer**, so the opt-in gate can't break an eligible run.

**Zero default-path risk.** The gate short-circuits — with it off, `_is_columnar_eligible` is never called, `_use_columnar` is always `False`, and the list path is byte-identical to before.

**Tests** (`tests/test_columnar_pipeline_parity.py`):
- `test_columnar_eligibility_predicate` — gate engages only for the narrow shape (rejects across-files, 2 matchkeys, exact, embedding scorer, ray backend).
- `test_pipeline_columnar_gate_parity` — end-to-end `dedupe_df` gate-ON yields clusters + dupes/unique counts **identical** to the default list path, with a spy asserting the columnar scorer is actually exercised (no silent fallback masking a divergence).

**Verification.** Full goldenmatch suite **3518 passed** with the gate default-off; coverage floors met; ruff clean.

⚠️ **Pre-existing failures (not from this PR):** 6 planner / partitioned-block-scoring / execution-plan tests (`test_planner_integration`, `test_autoconfig_planner_protocol`, `test_partitioned_block_scoring_pipeline`) fail **identically with this change stashed** on clean `main` in this container — they're environment-sensitive (RAM/CPU detection under load) and passed in CI's runners on the prior PRs. Flagging so a red lane here isn't misread as a Phase-A regression.

Phases B–D (df-native postflight, multi-scorer columnar aggregation, native cluster kernel) remain per the design note.

---

PR #666, opened 2026-06-01
fix(sweep): reframe Phase 1 gate to feasibility(5M)+parity(1M)

benzsevern:
Phase 1 Wave 5 (the validation gate). Waves 3-4 were verified NO-OPS (the substrate + pre-existing Phase A wiring already routes the eligible columnar branch score_blocks_columnar -> build_clusters_columnar end-to-end; the list-default scorer means other callers need no migration; columnar stays opt-in per spec).

## Gate reframe (per Stage 0's 5M-legacy-OOM finding)
`PHASE_CRITERIA["phase1"]` is now **feasibility(columnar_completes_5m) + parity(1M)** — a wall ratio vs legacy is un-measurable at 5M (legacy OOMs). `PHASE_BENCH_SCALE["phase1"]` 5M->1M (legacy baseline fits); the feasibility check runs columnar-ONLY at the real 5M.

Two review bugs fixed: the feasibility run was accidentally pinned to 1M (now hits real 5M on kill); parity now requires both cluster counts present (no None==None spurious pass).

Validation runs on the bench box via arrow-finish-line-sweep.yml; no 1M/5M data is built in the 16GB pytest lane. 19 unit tests (registry/classifier logic only).

## Skipped (with rationale)
Wave 5.3 CI lint: columnar stays opt-in, so a hot-path list-annotation guard protects a non-default contract — marginal value; skipped to avoid ci.yml churn.

### DIFF
--- a/packages/python/goldenmatch/scripts/arrow_finish_line_sweep.py
+++ b/packages/python/goldenmatch/scripts/arrow_finish_line_sweep.py
@@ -278,7 +278,13 @@ def _phase1_metrics(n: int, timeout_s: int) -> dict:
         # covered by the Wave 1 unit parity tests.
         list_clusters = parsed["list"].get("n_clusters")
         col_clusters = parsed["columnar"].get("n_clusters")
-        metrics["parity"] = list_clusters == col_clusters
+        # Require BOTH counts present: None == None would be a spurious True
+        # that masks a bench emitting no cluster data.
+        metrics["parity"] = (
+            list_clusters is not None
+            and col_clusters is not None
+            and list_clusters == col_clusters
+        )
         notes.append(
             f"parity = n_clusters equality at {n} "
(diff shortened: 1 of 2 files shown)


benzsevern:
Pausing this PR. Per the 2026-06-01 re-prioritization, we're pivoting from the polars-direct columnar work to bucket native (suggested backend up to 750k). Keeping the gate code here for the polars-direct circle-back.

Validation result that informed the pivot: the phase1 feasibility sweep (run 26778714813) FAILED — the columnar feasibility@5M run OOM-killed the 64GB runner (~15min). The pair-stream shrank the scorer-side list, but `build_clusters` still materializes Python cluster state (pair_scores dict + Union-Find) that OOMs at 5M. That's the shared cluster-build (Phase 2/3) — also the bucket-convergent lever — so the next work targets bucket native instead. Not merging now; will revisit.

Issue #663, opened 2026-06-01
perf(native): route dict-I/O kernels through Arrow arrays like score_block_pairs_arrow


benzsevern:
## Context / clarification

While reviewing arXiv literature for the "Python<->Rust boundary tax" (the dict-iteration floor that capped the cluster-orchestration and bulk-fingerprint kernels at ~1.1-1.2x and failed their decision gates), it is worth recording a scope clarification so we do not chase the wrong literature.

**Our boundary tax is in-process FFI dict construction, not network serialization.**

- The hot path already does the right thing: `score_block_pairs_arrow` takes Polars `.to_arrow()` arrays straight into the pyo3 native kernel in-process, zero-copy (`packages/python/goldenmatch/goldenmatch/backends/score_buckets.py:580-598`).
- The kernels that failed their gates (cluster-orchestration, bulk record-fingerprint) build Python `list[dict]` across the boundary. That dict construction is the cost, not any wire transfer.

## Action

Make the dict-I/O kernels accept/return Arrow arrays the way `score_block_pairs_arrow` already does (Arrow C Data Interface via the pyo3/arrow-pyarrow path), per the Arrow-native completion roadmap (`docs/superpowers/specs/2026-05-31-arrow-native-roadmap.md`). This is an internal refactor; no external library or paper is required.

## Literature note (do NOT adopt for the single-node tax)

The Apache Arrow papers that surface on a "columnar / zero-copy" search are about **network transport across machines**, a different problem:

- `2204.03032` Benchmarking Apache Arrow Flight - wire protocol over TCP/gRPC; its ">80% of data-access time is ser/de" is a network number.
- `2412.02192` Thallus - RDMA columnar transport over Infiniband.
- `2004.14471` Mainlining Databases (Pavlo, McKinney) - closer (accessing Arrow from external tools) but storage-format focused.

**These become relevant only for the distributed tier** (distributed-plan Phases 3-5), if/when scoring moves Arrow batches between Ray workers over the network. Park Flight/Thallus there, not on the single-node kernel boundary.

See `research/arxiv-reading-list.md` (section #3).

benzsevern:
Closing as **superseded**.

The Arrow-native roadmap rests on one thesis — lift the ~1.1-1.2x Python<->Rust per-row dict-iteration floor to clear a **2x decision gate**, in-process. That bet no longer holds:

[...]

PR #669, opened 2026-06-02
feat(identity): batch fingerprinting default-ON (2.6x) + bench per-N fallback

benzsevern:
Follow-up to #668 (#663-A). Flips `GOLDENMATCH_IDENTITY_BATCH_FINGERPRINT` **default-ON** after the measure-first bench, and fixes the bench to fall back to the fingerprint-only microbench when end-to-end hits the SQLite store limit.

## Bench result (run 26793348836, fresh native, no-PK 1M/5M)
| N | per-row | batch (Arrow) | speedup | peak RSS |
|---|---|---|---|---|
| 1M | 5.54s | 2.13s | **2.60x** | 2.2 GB |
| 5M | 27.52s | 10.55s | **2.61x** | 10.2 GB |

2.6x consistent on the real Arrow kernel. Byte-identical entity ids: **79 identity tests pass default-on AND with the `=0` kill-switch**.

## Why fingerprint-only
End-to-end `resolve_clusters` couldn't complete at 1M+ — the SQLite `IdentityStore` raised `too many SQL variables` bulk-upserting records (a store-scaling artifact, not the feature; production scale uses Postgres bulk COPY). The fingerprint-only microbench isolates the gate's actual delta (store I/O is identical on/off, so it cancels). The bench now falls back to it per-N + adds a `--mode` flag.

## Changes
- `_batch_fingerprint_enabled` default `1`; kill-switch `GOLDENMATCH_IDENTITY_BATCH_FINGERPRINT=0`.
- Test update: kill-switch test (`=0`) + a default-on test.
- Bench: per-N fingerprint-only fallback + `--mode`.
- Spec: bench result + decision recorded.

Caveat: the end-to-end PERCENTAGE win is unmeasured (resolve is store-bound at scale); the change is never a regression. Follow-up worth filing: the SQLite store's bulk-upsert variable-limit at 1M+.

### DIFF
--- a/.github/workflows/bench-identity-fingerprint.yml
+++ b/.github/workflows/bench-identity-fingerprint.yml
@@ -1,0 +1,47 @@
+name: bench-identity-fingerprint
+
+# Measure-first harness for the GOLDENMATCH_IDENTITY_BATCH_FINGERPRINT gate.
+# Compares batch_fingerprints (Arrow kernel) vs per-row record_fingerprint loop
+# inside resolve_clusters at scale. Run this before flipping the gate default
+# to "1". Not part of PR gating.
+on:
+  workflow_dispatch:
+    inputs:
+      ns:
+        description: "Row counts to sweep (comma-separated)"
+        default: "1000000,5000000"
+      runs:
+        description: "Repetitions per measurement (median reported)"
+        default: "3"
+
+jobs:
+  bench:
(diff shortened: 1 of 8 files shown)


benzsevern:
Superseded by a clean rebase off main (stacked-on-squash conflict).

Issue #688, opened 2026-06-03
bucket_score 44x slower on Linux x86_64 vs Windows x86_64 (same versions, native wheel loaded on both)


benzsevern:
## Summary

`bucket_score` is ~44x slower on Linux x86_64 (CI) than on Windows x86_64 (local laptop) for the same input + config + version stack + loaded native kernel. The slowdown is overwhelmingly in `goldenmatch.backends.score_buckets._score_one_bucket_fast`, not in `rapidfuzz` itself (which is actually *faster* on the Linux box).

## Repro shape

Synthetic NCVR-like person data (matches the v1.8 NCVR scale-audit shape):

- 100,000 rows, 8 columns: `matching_id, source, npi, email, first_name, last_name, phone_number, zip5`
- Non-null rates: npi 61%, email 73%, phone 30%, zip5 32%, first/last_name 56%
- All values are strings except `matching_id` (Int64)
- ~3% of rows are "duplicates" sharing some subset of identifiers

Config (explicit, no auto-config):

- 3 exact matchkeys on `npi`, `email`, `phone_number`
- 2 weighted matchkeys: fuzzy_email (`levenshtein` on email, threshold 0.85) and weighted_name_id (5 fields: jw on first/last, exact on zip/phone/npi, threshold 0.99)
- 4 blocking keys: `npi`, `email`, `phone_number`, `last_name[:4]`
- `backend="bucket"`, `max_block_size=10000`, `skip_oversized=True`

Full repro script in the appendix below.

## Observed perf matrix

All runs identical input parquet + config + `POLARS_MAX_THREADS=4` `RAYON_NUM_THREADS=4` + `native_available=True`:

| Env | `bucket_score` pass 1 | pass 2 | dedupe total |
|---|---|---|---|
| **Local: Win 11, i5-1335U (10c/12t)** | ~3s | ~3s | **6.5s** |
| Linux: `ubuntu-latest-xlarge`, AMD EPYC 9V74 (8t), Python 3.12.13, goldenmatch 1.25.0, polars 1.40.1, rapidfuzz 3.14.5 | 191.8s | 124.3s | **316s** |
| Same Linux, polars 1.41.2 | 191.4s | 124.7s | 316s |
| Same Linux, single-threaded (`os.cpu_count` monkeypatched to 1) | 191.9s | 122.6s | 315s |
[...]

PR #692, opened 2026-06-03
perf(native): sequential block-scoring path to dodge rayon LockLatch park (#688)

benzsevern:
## The real root cause of #688

The stale-wheel fix (#689, republished 0.1.2) shipped correctly but **did not move the wall** — confirmed in [the issue](https://github.com/benseverndev-oss/goldenmatch/issues/688#issuecomment-4608134587) by py-spy `--native` on `ubuntu-latest-xlarge` (EPYC 9V74): ~419s (the entire `bucket_score` wall) parked in `rayon_core::latch::LockLatch::wait_and_reset` (a futex), with **zero CPU work** anywhere. `score_one` barely registers in the flamegraph. The Arc-handle exclude path was already engaged on 0.1.2, so the per-call HashSet rebuild was never the bottleneck.

The kernel's internal `par_iter().collect()` blocks the calling thread on a latch that makes near-zero forward progress on this Linux env, turning a sub-second scoring job into ~190s.

Note: the 25M bench that "worked" (bucket_score 39s) is dated 2026-05-19; `score_block_pairs_arrow` landed 2026-05-26 (#514) — so that bench never used this kernel. The kernel's rayon path has only ever been **parity**-validated, never **perf**-validated on a Linux runner.

## Fix

The kernel's internal rayon is redundant in the common case: the Python caller (`score_buckets`) already fans buckets across a `ThreadPoolExecutor` with the GIL released here, so each kernel call is one of N concurrent calls.

- Factor the per-block scorer into one closure shared by both paths (they can't diverge).
- Score small/medium calls **in the calling thread** (no rayon, no latch); dispatch to rayon only above a candidate-pair threshold (default 20M, tunable via `GOLDENMATCH_NATIVE_RAYON_MIN_PAIRS`; `0` = always rayon, huge = always sequential). Both paths walk spans in order → byte-identical `(min,max)` pairs.
- `goldenmatch_native.__version__` now reads from installed dist metadata (was a hardcoded `0.1.0` that reported stale on the 0.1.2 wheel — flagged in the issue).
- Bump to `0.1.3` (pyproject + Cargo) for the republish.

**Refutes** the GIL-reacquire-in-rayon-closure hypothesis: the closure touches zero Python objects (arrow arrays, Rust `HashSet`, rapidfuzz-rs), so there's no GIL for a worker to re-enter.

## Verification status

- ✅ `cargo check --release` + `cargo clippy` pass.
- ✅ `tests/test_native_block_seq_parity.py` forces each path via the env gate and asserts the sequential and rayon paths (and `find_fuzzy_matches`) emit identical pairs.
- ⏳ **NOT yet perf-verified on Linux.** The 44x wall fix needs a Linux run of the #688 repro (or a CI bench) before this is claimed resolved. **Holding merge until that measurement exists** — not repeating the "verified it shipped ≠ verified it works" miss from #689.

Refs #688

### DIFF
--- a/packages/python/goldenmatch/tests/test_native_block_seq_parity.py
+++ b/packages/python/goldenmatch/tests/test_native_block_seq_parity.py
@@ -1,0 +1,96 @@
+"""Issue #688 follow-up: the native block kernel's sequential path and rayon
+path must emit byte-identical pairs.
+
+The kernel (`score_block_pairs_arrow`) now scores small/medium calls in the
+calling thread (no rayon) and only fans out to rayon above a candidate-pair
+threshold, because rayon's blocking `collect` parked the calling thread on a
+futex `LockLatch` for ~190s on some Linux runners (the 44x slowdown). The
+threshold is overridable via `GOLDENMATCH_NATIVE_RAYON_MIN_PAIRS` (0 = always
+rayon, huge = always sequential), which lets this test exercise BOTH paths on
+the same input and assert they agree -- and that both agree with the Python
+slow path (the source of truth).
+
+Requires the native kernel; skipped on pure-Python.
+"""
+from __future__ import annotations
+
+import polars as pl
+import pytest
+from goldenmatch.backends.score_buckets import score_buckets
+from goldenmatch.config.schemas import (
+    BlockingConfig,
+    BlockingKeyConfig,
+    MatchkeyConfig,
+    MatchkeyField,
+)
+from goldenmatch.core._native_loader import native_available, native_module
+from goldenmatch.core.matchkey import _xform_sig
+from goldenmatch.core.scorer import find_fuzzy_matches
+
+_NATIVE_ARROW = native_available() and hasattr(
+    native_module(), "score_block_pairs_arrow"
+)
+pytestmark = pytest.mark.skipif(
+    not _NATIVE_ARROW,
+    reason="needs the native score_block_pairs_arrow kernel",
+)
+
+
+def _prepared() -> pl.DataFrame:
+    # Three blocks of near-duplicate names so each block emits several pairs,
+    # plus a few non-matches so the threshold actually filters.
+    field = MatchkeyField(field="name", scorer="jaro_winkler", weight=1.0)
+    col = _xform_sig(field)
+    names = [
+        "alice", "alica", "alise",      # block A (similar -> pairs)
+        "robert", "robbert", "rupert",  # block B
+        "xavier", "yvonne", "zelda",    # block C (dissimilar -> few/no pairs)
+    ]
+    blocks = ["A", "A", "A", "B", "B", "B", "C", "C", "C"]
+    return pl.DataFrame({
+        "__row_id__": list(range(len(names))),
+        "name": names,
+        col: names,
+        "__block_key__": blocks,
+    })
+
+
+def _mk() -> MatchkeyConfig:
+    return MatchkeyConfig(
+        name="t", type="weighted", threshold=0.7,
+        fields=[MatchkeyField(field="name", scorer="jaro_winkler", weight=1.0)],
+    )
+
+
+def _blocking() -> BlockingConfig:
+    return BlockingConfig(strategy="static", keys=[BlockingKeyConfig(fields=["name"])])
+
+
+def _run(monkeypatch, min_pairs: str) -> list[tuple[int, int, float]]:
+    monkeypatch.setenv("GOLDENMATCH_NATIVE_RAYON_MIN_PAIRS", min_pairs)
+    return score_buckets(_prepared(), _blocking(), _mk(), matched_pairs=set())
+
+
+def _keys(pairs):
+    return sorted((min(a, b), max(a, b)) for a, b, _ in pairs)
+
+
+def test_sequential_and_rayon_paths_agree(monkeypatch):
+    rayon = _run(monkeypatch, "0")          # 0 -> always rayon
+    seq = _run(monkeypatch, "100000000000")  # huge -> always sequential
+    assert _keys(seq) == _keys(rayon), (
+        f"seq vs rayon pair-set mismatch:\n  seq={seq}\n  rayon={rayon}"
+    )
+    # Scores agree pair-for-pair too.
+    for (sa, sb, ss), (ra, rb, rs) in zip(sorted(seq), sorted(rayon)):
+        assert (sa, sb) == (ra, rb)
+        assert ss == pytest.approx(rs, abs=1e-9)
+
+
+def test_both_paths_match_slow_path(monkeypatch):
+    """Both kernel paths must match find_fuzzy_matches (the source of truth)."""
+    df = _prepared()
+    slow = find_fuzzy_matches(df, _mk(), exclude_pairs=frozenset(), pre_scored_pairs=None)
+    slow_keys = _keys(slow)
+    assert _keys(_run(monkeypatch, "0")) == slow_keys
+    assert _keys(_run(monkeypatch, "100000000000")) == slow_keys
--- a/packages/python/goldenmatch/tests/test_native_block_seq_parity.py
+++ b/packages/python/goldenmatch/tests/test_native_block_seq_parity.py
@@ -10,6 +10,13 @@ rayon, huge = always sequential), which lets this test exercise BOTH paths on
 the same input and assert they agree -- and that both agree with the Python
 slow path (the source of truth).
 
+The fixture is a SINGLE block (constant blocking key) with mixed similar /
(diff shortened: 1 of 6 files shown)


benzsevern:
Built `goldenmatch-native` from `main` (de912d3a) via maturin in CI, verified `__version__: 0.1.3` and `native_available: True`. Ran the same bench three ways:

| Config | bucket_score pass 1 | pass 2 | dedupe wall |
|---|---|---|---|
| PyPI 0.1.2 (rayon path) | 191.8s | 124.3s | 316s |
| **from-source 0.1.3, default threshold** (auto-picks seq for ~10M pairs) | 169.3s | 116.9s | **286.9s** |
| **from-source 0.1.3, `RAYON_MIN_PAIRS=999999999999`** (forced sequential) | 168.9s | 116.2s | **285.9s** |

Default-threshold and forced-sequential are within 1s of each other → the threshold is correctly routing our 100K case to the sequential path. **The sequential path itself is the wall.**

The ~10% drop vs the rayon path (316→286) is consistent with skipping the rayon dispatch overhead per bucket, but the bulk of the time is somewhere both paths share.

Math reality-check on the sequential path:

- 455 blocks, ~10M total candidate pairs across the two passes
- rapidfuzz microbench on this same runner: `cdist 1000x1000 Levenshtein.normalized = 32ms` → ~31M ops/sec for the actual scorer
- Expected sequential wall at this throughput: **~0.3s**
- Observed sequential wall: **285s**
- **~950× overhead** somewhere above the actual scorer call

So the sequential path isn't really sequential-scoring-bound either. The cost lives in per-block setup, span construction, arrow buffer access, or per-pair dispatch — something both the rayon and sequential branches walk through identically.

Re-profiling now with `py-spy --native` against the from-source build to see where the time actually goes when rayon is out of the picture. The flamegraph will look completely different if `LockLatch::wait_and_reset` is gone — and whatever takes its place is the real bug. Will follow up with that flamegraph in a few minutes.

The fix dodges the symptom I described in #688, but doesn't land the wall. Worth keeping the PR open for the rayon-avoidance benefit, but it's not the resolve for #688.

benzsevern:
Re-profiled the from-source 0.1.3 build under `py-spy --native --idle --rate 100`, with `GOLDENMATCH_NATIVE_RAYON_MIN_PAIRS=999999999999` set in the workflow env (verified printed in the runtime env dump). **The flamegraph is structurally identical to the pre-fix one.**

Flamegraph SVG (interactive): https://gist.github.com/benzsevern/42f0b79a47f58d8a9bd5970446569cd9

Hot frames out of 330,012 total samples (~369s wall under profiler):

```
36,643 (11.1%)  _score_one_bucket_fast (score_buckets.py:592)
36,616 (11.1%)  rayon_core::latch::LockLatch::wait_and_reset::h4d0e...  (goldenmatch_native/_native.abi3.so)
36,616 (11.1%)  syscall (libc.so.6)
```

**Zero scoring frames anywhere in the profile.** No `score_block_pairs_arrow`, no `score_span`, no `score_one`, no rapidfuzz internals, no jaro_winkler / levenshtein. If the sequential `spans.iter().flat_map(...).collect()` path from your diff were actually executing, the bulk of those 36k samples would be in pure-Rust `for`-loop frames doing the actual work. They aren't there.

Verified:

- Build artifact: `goldenmatch_native.__file__` = `.venv/lib/python3.12/site-packages/goldenmatch_native/__init__.py`, `__version__` = `0.1.3`.
- Native attrs include `score_block_pairs_arrow` (your patched function).
- Python calls it at `score_buckets.py:592` (your code reads `native_module().score_block_pairs_arrow(...)`), which matches the flamegraph frame.
- Env: `GOLDENMATCH_NATIVE_RAYON_MIN_PAIRS=999999999999` is in the printed env dump *before* the python process starts, so `std::env::var("GOLDENMATCH_NATIVE_RAYON_MIN_PAIRS")` from inside the call should see it.
- Default-threshold run (no env override; our ~10M pairs is below your 20M default → should route sequential) gives the same wall as the forced-sequential run: 286.9s vs 285.9s. Both ~10% better than the pre-fix 316s, but nowhere near the expected sub-second.

Two hypotheses for why the env gate looks honored (numbers move slightly) but the hot stack doesn't change:

1. **The `LockLatch::wait_and_reset` we're seeing isn't from `score_block_pairs_arrow` at all.** Something else in the kernel (or in a transitive crate, or in the python `dedupe_df` pipeline calling another native function we don't have visibility into) is rayon-dispatching independently. The 10% wall improvement is genuine — your sequential path IS running for the per-block scoring — but it's a small slice of the dedupe wall, and the bulk is rayon-elsewhere.
2. **The env var is read by your gate code, but somewhere downstream/inside `score_one` or one of its helpers, a global rayon dispatch happens regardless** (less likely — `score_one` is per-pair string scoring, no rayon expected).

Diagnostic ask: can you add a single `eprintln!("[score_block_pairs_arrow] chose path: {}", if total_pairs >= rayon_min_pairs { "rayon" } else { "sequential" });` right before the `if total_pairs >= rayon_min_pairs` branch? One line, prints to stderr per kernel call, will tell us conclusively whether the gate is taking the sequential branch in our run. If yes, the rayon LockLatch is coming from somewhere else and we need to hunt that. If no, the gate isn't engaging the way you expect.

I can rebuild from a branch with that print added and re-run in <5 min. Just point me at a ref.

(Wall numbers, all on the same `ubuntu-latest-xlarge` runner, same input, `BENCH_SINGLE_THREADED=1`:)

| Config | pass 1 | pass 2 | dedupe wall |
|---|---|---|---|
| PyPI 0.1.2 (your rayon path) | 191.8s | 124.3s | 316s |
| from-source 0.1.3, threshold default 20M (auto-seq for 10M pairs) | 169.3s | 116.9s | 286.9s |
| from-source 0.1.3, `RAYON_MIN_PAIRS=999999999999` (forced seq) | 168.9s | 116.2s | 285.9s |
| from-source 0.1.3, forced seq, **under py-spy** (this run) | 217.9s | 150.0s | 368.9s |

benzsevern:
**Fix verified on Linux. 240× speedup, byte-identical output.** PR #692 + #699 from HEAD (3ecd2e0), both Python + native built from source on `ubuntu-latest-xlarge`, same bench, same input parquet, same config:

| Config | dedupe wall | total bench wall |
|---|---|---|
| PyPI 0.1.2 (your original rayon path, #688 baseline) | 286.9s | 319.9s |
| **HEAD (PR #692 + #699)** | **1.2s** | **3.5s** |

`GOLDENMATCH_BUCKET_DEBUG=1` from PR #699 confirms the per-bucket breakdown is now sane:

```
[score_buckets][DEBUG] native bucket-call breakdown over 16 calls / 8618 blocks / 33 pairs:
  prep   (sort+group_by+to_arrow):   0.086s ( 54.7%)
  kernel (score_block_pairs_arrow):  0.071s ( 45.1%)
  post   (match-mode filter):        0.000s (  0.2%)
  total in-worker: 0.157s; slowest single kernel call: 0.020s

[score_buckets][DEBUG] native bucket-call breakdown over 16 calls / 8618 blocks / 0 pairs:
  prep   (sort+group_by+to_arrow):   0.059s ( 20.5%)
  kernel (score_block_pairs_arrow):  0.229s ( 79.4%)
  total in-worker: 0.288s; slowest single kernel call: 0.067s
```

Sub-second per-call kernel + ~50/50 prep/kernel split. The 1273× overhead-over-actual-scoring ratio from the #688 followup is gone; this matches the local Windows numbers I started the issue with.

## Quality parity (byte-identical to slow runs)

| Metric | Slow runs (316s baseline) | Fast run (3.5s) |
|---|---|---|
| Total clusters | 98,391 | 98,391 |
| Multi-record clusters | 1,248 | 1,248 |
| Max cluster size | 5 | 5 |
| Recall on Splink's multi-record clusters | 61.1% | 61.1% |
| Novel multi-record clusters | 1,225 | 1,225 |
| Records promoted singleton → merged | 2,810 | 2,810 |

The bench parity test passed on the fast path with the exact same cluster set as the slow path. Confirms the rayon → sequential gate is output-invariant for our shape.

## Diagnostic context

Earlier from-source-only-native runs (just PR #692, with stock PyPI `goldenmatch==1.25.0`) showed the wall barely moved. That was misleading — it took the from-source **Python** package (1.26-dev with the PR #699 debug print, but also whatever other changes have landed between v1.25.0 and HEAD) to actually exercise the sequential path correctly. Worth a follow-up to figure out whether (a) any unreleased Python change between 1.25.0 and HEAD is load-bearing for the speedup, or (b) the native-only build was getting overridden somehow despite `--reinstall`. Either way, **HEAD is fast**.

You can tag `goldenmatch-native-v0.1.3` whenever — the perf verification you were holding for is in.

Closing #688 in a moment.

PR #720, opened 2026-06-04
fix(autoconfig): admit high-cardinality identifiers to exact matchkeys (#715)

### LINKED ISSUE #715: auto-config commits RED 'low-precision' fallback config + runs anyway; O(N^2) guard ignores blocking + null density
## Summary

`dedupe_df(df, config=None)` on a ~1M-row dataset with sparse high-cardinality identifier columns commits a `RED` / "best-effort, may be low-precision" config that produces severe over-merge (582K raw clusters from 996K input rows). The introspective controller's O(N²) guard excludes every exact-eligible identifier column based on input row count alone, without considering that blocking is being configured for the same run and would bound the candidate-pair join.

The fallback config the controller commits is a single `levenshtein` matchkey on the cosmetic-ish triple `(source, first_name, last_name)` at threshold 0.75, no negative evidence, no blocking — which is unusable as production output on the described data shape.

## Environment

- `goldenmatch` HEAD (`3ecd2e0`), built from source via maturin
- `goldenmatch-native` 0.1.3 (from the same SHA)
- Python 3.12.13, polars 1.40, ubuntu-latest-xlarge runner
- Dataset: ~1M rows, healthcare-provider-shape (identifier columns ~30–70% non-null), columns include `source`, `npi`, `email`, `phone_number`, `first_name`, `last_name`, `zip5`, plus a stable per-record `matching_id`
- Call: `dedupe_df(df, config=None, confidence_required=False)`

## What I observed

Three log lines tell the whole story:

```
All 1 exact-eligible columns were excluded by auto-config guards
  (email (dataset has 996558 rows; exact self-join is O(N^2))).
  Falling back to fuzzy-only matchkeys — if any of these columns actually
  are identifiers, provide an explicit config.

auto-config committed best-effort RED config (iter=v0,
  stop_reason=BUDGET_ITERATIONS, failing_subprofile=cluster);
  downstream pipeline will run but output may be low-precision

dedupe_df returned in 972.9s
raw clusters: 582,636            (input: 996,558 rows)

--- auto-config selected matchkeys ---
  - fuzzy_match [None t=0.75] fields=['source', 'first_name', 'last_name'] NE=[]
auto-config blocking keys:       (none)
```

Compared to an explicit hand-tuned config that solves the same problem:

| | hand-tuned config | auto-config |
|---|---|---|
| dedupe wall | 27s | **973s (36×)** |
| raw clusters | ~3,063 multi-record / ~993K total | **582,636** |
| precision (hand-labeled 1M sample) | 98.8% | not measurable — pipeline crashed downstream |
| matchkeys picked | 5 (3 exact + 2 weighted, NE on disagreement) | 1 (fuzzy levenshtein on name + source) |
| blocking | 4 keys (npi, email, phone, last_name[:4]) | none |

The pipeline ultimately crashed during the prune step (job timed out at 20 min) because the over-merged candidate-edge set was too large.

## What I think the root cause is

The O(N²) guard treats every exact-eligible column the same way regardless of:

1. **Whether blocking is also being configured.** The point of blocking is to bound candidate pairs to within-block. The guard's `N×N` cost premise assumes no blocking; with blocking, the cost is `sum(b_i × b_i)` for block sizes `b_i`, which is *much* smaller than `N²` for the sparse healthcare-shape data (no provider has > maybe a few hundred records sharing the same blocking key).
2. **The null rate on the column.** A 73%-non-null `email` column at 1M rows has ~730K populated values, and exact self-join's effective cost is `O(non_null²)` not `O(N²)`. Even without blocking, that's 0.5T worst-case ops, but the kernel only emits pairs where the column EQUALS — which is *much* sparser than total combinations for high-cardinality columns like email.
3. **What downstream cost actually is.** A 1M-row exact-NPI match with 60% non-null gives a candidate-pair count in the low millions, not low trillions, because hash-equality joins are linear-ish in non-null count. The O(N²) framing is the *worst case for a nested loop*, not the actual cost of the kernel call.

The end-state behavior — commit a config with no exact matchkeys, no blocking, threshold 0.75 levenshtein on name, then emit a `RED` warning — gives the user a config that the controller already knows is bad, then runs anyway.

## Why this matters

For any healthcare-provider-like dataset with sparse high-cardinality identifiers (NPI, email, phone), `dedupe_df(df, config=None)` is effectively unusable. A first-time user with no GoldenMatch experience runs this expecting reasonable defaults, gets a 16-minute run that returns 582K clusters from 1M rows, and concludes the library doesn't work on their data. (A more experienced user pulls the auto-config warning and learns to write `build_config` by hand — which works great, ~30s and 98.8% precision.)

The "auto-config" branding implies this should produce a reasonable result. The current behavior — "commit best-effort RED config, run anyway, return low-precision output" — buries the failure inside a long, expensive pipeline rather than refusing up front.

## Suggested fixes (in rough order of effort vs impact)

1. **Honor configured blocking when computing the exact-eligibility cost.** If blocking will be applied for the same matchkey universe, the controller should compute candidate-pair cost using estimated block-size distribution, not raw `N²`. Even a coarse estimate (`sum(b² for b in expected_block_sizes)` from a few small block-size samples) would drop the guard's threshold by 3+ orders of magnitude for this data shape.
2. **Consider non-null density.** `effective_N = N × non_null_rate` is a closer-to-honest input-size estimate for sparse identifier columns. Reject only when `effective_N²` exceeds the budget.
3. **When the controller would otherwise commit a `RED` config, raise instead of running.** The library knows the config is bad (it logged "may be low-precision" and `stop_reason=BUDGET_ITERATIONS`). A `RuntimeError("Auto-config could not find a workable config; please provide one explicitly")` is much better DX than a 16-minute run that returns garbage. An opt-in `allow_red_config=True` keeps the escape hatch.
4. **Document the guard.** A short doc page "When auto-config falls back to RED" with the cost model, the guards, and how to override would let users self-rescue. The current path is: read the warning, grep the source.

## Minimal repro

Roughly any healthcare-shape data triggers it. The shape that triggered the issue here:

- 1M rows
- `source` column (categorical, 10–20 values)
- `npi` column (10-digit numeric strings, ~60% non-null, high cardinality among non-nulls)
- `email`, `phone_number` columns (~30–70% non-null, high cardinality)
- `first_name`, `last_name`, `zip5` (mostly non-null, moderate cardinality)

Calling `dedupe_df(df, config=None, confidence_required=False)` on data with this shape and size reproduces the RED-config fallback 100% of the time.

I'm happy to put together a synthetic-data repro script (no real records) if that would help — let me know.

## Workaround

For now: write `build_config` by hand. A 5-matchkey config (3 exact on identifier columns + 2 weighted/fuzzy + NE on disagreement, plus 4 blocking keys) runs in ~30s and achieves 98.8% precision on the same data. The auto-config path is currently 36× slower for substantially worse output.

benzsevern:
Fixes #715.

## Root cause
On healthcare-provider-shape data (sparse high-cardinality `npi`/`email`/`phone`), zero-config auto-config committed a fuzzy-only, mega-block config that severely over-merged (582K clusters from 996K rows, ~16 min, downstream crash). Confirmed via the synthetic repro (workflow `repro-issue-715.yml`): identity columns were excluded from matchkeys by a **pincer** of three gates:
- `col_type=="identifier"` (npi, phone) skipped outright;
- `email` dropped by a blanket row-count "Guard 1" (`df.height > 10000`) that mismodeled a Polars hash self-join as `O(N^2)`;
- `zip` treated as a blocking-only signal.

Exact matchkeys are hash self-joins whose cost is bounded by **cardinality**, not row count, so a high-cardinality column is both cheap and mega-cluster-safe. The mega-cluster risk is the *opposite* shape (low cardinality), already guarded by the `cardinality_ratio >= 0.5` gate.

## Fix (Component 1 — the core)
In `build_matchkeys`:
- Drop the row-count Guard 1.
- Admit exact matchkeys on the cardinality band **`0.5 <= card < 1.0`** (lower bound = existing mega-cluster guard; upper bound excludes perfectly-unique surrogate keys like `matching_id`).
- Stop skipping `col_type="identifier"` outright; route it through the exact path under the band.
- Count `identifier` in the "all exact-eligible excluded" warning.

## Blocking (Component 2) — verification, no new code
`build_blocking` already self-corrects (oversized-block filter + `_build_compound_blocking` + multi-pass fallback). #715's empty blocking was a controller-commit artifact of the fuzzy-only RED config. The e2e regression confirms blocking is **retained** once the config goes healthy.

## What's here
- `build_matchkeys` change + 7 unit/e2e tests (`tests/test_autoconfig_pincer_715.py`), incl. the band boundaries and the healthcare-shape e2e (exact MK on identifier + bounded blocking retained).
- Repro verdict reframed to the matchkey pincer; CI assert step added (`repro-issue-715.yml` now fails on `PINCER PRESENT`).
- Docs: `docs/autoconfig-cost-model.md`.
- Note: `build_probabilistic_matchkey` still skips `identifier` by design (Fellegi-Sunter m/u-aware admission is a follow-up; flagged in-code).

## Local repro, post-fix
```
identifier-eligible cols (identifier/email/phone, card 0.5-1.0): ['npi', 'email']
exact matchkeys produced: 2  fields=['email', 'npi']
blocking keys present:    True
>>> PINCER RESOLVED
```

## Validation (pre-merge gate)
- In-house backend-parity quality gate (#528): runs in this PR's CI.
- DQbench T1/T2/T3 (mega-cluster-risk backstop, watch T3 precision): results to be posted before merge.

### DIFF
--- a/.github/workflows/repro-issue-715.yml
+++ b/.github/workflows/repro-issue-715.yml
@@ -61,6 +61,19 @@ jobs:
             echo '```'
           } >> "$GITHUB_STEP_SUMMARY"
 
+      - name: Assert pincer resolved
+        working-directory: packages/python/goldenmatch
+        run: |
+          if grep -q "PINCER PRESENT" repro715.log; then
+            echo "::error::#715 pincer still present -- identifier columns produced no exact matchkeys"
+            exit 1
+          fi
+          if grep -q "PINCER RESOLVED" repro715.log; then
+            echo "pincer resolved"
+          else
+            echo "::warning::inconclusive verdict"
+          fi
+
       - name: Upload log
         if: always()
         uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a  # v7.0.1
--- a/packages/python/goldenmatch/docs/autoconfig-cost-model.md
+++ b/packages/python/goldenmatch/docs/autoconfig-cost-model.md
@@ -1,0 +1,113 @@
+---
+layout: default
+title: Auto-config cost model
+nav_order: 10
+---
+
+# Auto-config cost model: why exact matchkeys are cardinality-gated, not row-count-gated
+
+Zero-config auto-config inspects your columns and decides, for each one, whether to
+anchor an exact matchkey, use it as a blocking key for fuzzy matching, or skip it.
+The decision is driven by **cardinality ratio** (distinct values / total rows), not raw
+row count -- and the reason comes from how exact matchkeys are implemented.
+
+---
+
+## How exact matchkeys work
+
+An exact matchkey is a Polars hash self-join, not a nested loop. GoldenMatch groups
+records by the matchkey value and emits candidate pairs only within each group.
+Cost is proportional to the number of emitted pairs, which is bounded by how many
+records share any single value -- the column's cardinality.
+
+A high-cardinality column (most values are unique or near-unique) emits a small
+number of pairs per group. It is both cheap to compute and safe: it cannot create a
+mega-cluster because no single value matches thousands of unrelated records.
+
+---
+
+## The admission band
+
+Auto-config admits a column as an exact matchkey when its cardinality ratio falls in
+the half-open interval `[0.5, 1.0)`.
+
+| Cardinality ratio | Interpretation | Decision |
+|-------------------|----------------|----------|
+| `< 0.5` | Too few distinct values. An exact match collapses many records that merely share a common value (e.g., a status code, a state abbreviation) into one mega-cluster. | Excluded |
+| `>= 0.5 and < 1.0` | A real shared identifier. Two records for the same entity share the value; distinct enough to stay bounded, shared enough to catch true duplicates. | **Admitted as exact matchkey** |
+| `== 1.0` | Perfectly unique per-record surrogate (e.g., a database primary key). Never shared, so an exact match finds nothing and asserts no real-world identity. | Excluded for config hygiene |
+
+The default blocking-candidate cardinality gate is controlled by
+`GOLDENMATCH_BLOCKING_MAX_RATIO` (default `0.5`). Columns above this threshold are
+too unique to block on and skip the fuzzy blocking path entirely.
+
+---
+
+## Identifiers do not need blocking
+
+Because the exact matchkey is a hash join, high-cardinality identifiers anchor
+matchkeys directly. They are intentionally NOT used as blocking keys.
+
+A near-unique column makes terrible blocks -- each group contains one or two records,
+so the candidate-pair reduction is negligible. Blocking exists to bound the candidate
+space for **fuzzy** matchkeys (names, addresses), where you need approximate
+comparison within a manageable neighborhood.
+
+Auto-config caps single-key block size via `max_safe_block` (scaled by row count)
+and falls back to compound or multi-pass blocking when a single key would produce
+oversized blocks.
+
+---
+
+## Worked example: healthcare-provider dataset
+
+| Column | Approx. cardinality ratio | Exact matchkey? | Blocking key? | Fuzzy? |
+|--------|--------------------------|-----------------|---------------|--------|
+| `source` | 0.003 | No (too low) | No | No |
+| `npi` | 0.98 | **Yes** | No | No |
+| `email` | 0.82 | **Yes** | No | No |
+| `phone_number` | 0.75 | **Yes** | No | No |
+| `matching_id` | 1.0 | No (perfectly unique) | No | No |
+| `zip5` | 0.12 | No (too low) | **Yes** | No |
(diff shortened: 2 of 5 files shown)


benzsevern:
**DQbench validation decision (Task 6):** proceeding on the in-house backend-parity quality gate (#528) running in this PR's CI, plus the following analysis, rather than a separate DQbench run.

**Why the fix is a no-op at DQbench scale:** the removed Guard 1 only fired above 10,000 rows. DQbench T1/T2/T3 tiers are well under that, so `email` was already admitted as an exact matchkey there pre-fix. The only size-independent change is admitting `col_type="identifier"` columns, which the person/email DQbench tiers do not contain. So DQbench T1/T2/T3 are expected unchanged.

A full DQbench run against the fixed build (worktree venv) is tracked as a follow-up. The mega-cluster-risk surface (adversarial same-id/email collision) remains covered by the #528 quality gate + the band's `cardinality_ratio >= 0.5` lower bound (unchanged by this PR).

Issue #417, opened 2026-05-21
Autoconfig still picks degenerate blocking key on 1.13M-row real-world table after #414


benzsevern:
## Summary

After PR #414 ("sample-corrected blocking + composite-search wire-through" which closed #410), `goldenmatch sync` against a real 1.13M-row Postgres table still produces a single degenerate block. Autoconfig logs `succeeded=False, f1_proxy=0.326` and proceeds with the bad config anyway. Sync then wedges in `bucket_score` indefinitely.

This is a follow-up regression — the wire-through landed but the autoconfig is not actually selecting a discriminative composite key for the blocking step.

## Repro

- **Table:** `gm.pubrecord_pub1_clean` (1,131,769 rows, GoldenFlow-cleaned, 17 columns: ids + names + address + phone/email + specialty/degree)
- **Install:** `pip install` from `main` (sparse-checkout `packages/python/goldenmatch`, commit `2a03e3a2` / #413 merged; PR #411 with the composite-search wire-through is in this build)
- **Environment:** Vercel Sandbox, 8 vCPU / 16 GB RAM, `OMP_NUM_THREADS=8`, `OPENBLAS_NUM_THREADS=8`, `MKL_NUM_THREADS=8`, `GOLDENMATCH_N_JOBS=8`
- **Command:**

```
goldenmatch sync \
  --connection-string $DATABASE_URL \
  --table gm.pubrecord_pub1_clean \
  --exclude-columns uniquerecord_id,r_id,pub_id,dm_npi,middle_name,address2,country \
  --verbose
```

`--exclude-columns` strips identifiers (`uniquerecord_id`, `r_id`, `pub_id`, `dm_npi`) and degenerate columns (`middle_name`, `address2`, `country`). Remaining candidate columns: `first_name`, `last_name`, `address1`, `city`, `state`, `zip`, `phone`, `email`, `dm_specialty`, `dm_degree`. `[zip, last_name]` and `[state, last_name]` are obvious discriminative composites here.

## Observed log

```
controller:auto-config: recorded run to memory
  (sig=be5b6914ec2cbc54, succeeded=False, f1_proxy=0.32558139534883723)
Auto-configured matching rules.
exclude_columns (7): uniquerecord_id, r_id, pub_id, dm_npi, middle_name, address2, country
Syncing...
INFO:goldenmatch.db.metadata:GoldenMatch metadata tables ready
[...]

PR #723, opened 2026-06-04
fix(autoconfig): bound blocking cost + refuse RED by default (#715 reopened)

benzsevern:
Reopens-fix for #715. PR #720 fixed the matchkey half; the reporter verified that and reopened because (a) auto-config still committed a RED config and **ran anyway**, and (b) a new failure mode appeared: an unbounded `soundex(name)` blocking pass producing ~30K-row blocks at 1M → 39.6M candidate edges → 18-min run → cancelled.

## Root cause (reproduced, not reasoned)
Ran `profile_columns` + `build_blocking` on the synthetic healthcare shape at 50K/200K:
- **B1:** `build_blocking` emitted single-column `soundex(name)` passes with **no per-pass block-size gate**. v0 builds blocking on the full df, so block sizes are exact — the code just never checked them.
- **B2:** a **sparse `zip5`** (~45% null) reclassifies `zip→identifier` and was dropped from `_build_compound_blocking`, removing the one column that would bound block size (`zip5+last_name`).
- **A:** the committed RED config ran anyway; the existing raise was gated on `confidence_required` (which the reporter set `False`).
- **Iteration budget:** `ControllerBudget.max_iterations` was hardcoded at 3 for all sizes (the reporter's hypothesis — confirmed, though not load-bearing: the controller iterates on a sample that masks the at-scale blow-up).

## Fix (four parts)
1. **B1** — `project_max_block_size` (linear full-N projection) gates every emitted blocking key/pass; oversized passes dropped; degenerate (empty) config when nothing bounds, so the controller refuses rather than shipping a bomb.
2. **B2** — `_build_compound_blocking` judges candidates by block size (non-singleton, `card<1.0`, non-null distinct ratio under the gate) not col_type, and relaxes the null ceiling for the multi_pass component role → `zip5+last_name` reachable.
3. **Iteration budget** — `for_dataset` scales `max_iterations` 3→4→5 across the 100K/1M tiers.
4. **A** — new `allow_red_config` flag: a committed RED config at `n_rows >= REFUSE_AT_N` now **raises by default** (independent of `confidence_required`; `confidence_required=False` no longer bypasses it). `allow_red_config=True` restores warn-and-run. Reconciled with the existing #417 degenerate-blocking guard to raise once. Small-N (<100K) keeps warn-and-run (cheap; deliberate existing design).

## Tests
- New `tests/test_autoconfig_blocking_cost_715.py` (B1 cap invariant, B2 bounded compound, dense-zip no-regression, iteration scaling) + `tests/test_allow_red_config.py` (default-raise at scale, escape hatch, small-N warn-and-run, `confidence_required=False` now raises).
- Regression: `test_autoconfig.py` + `test_autoconfig_regressions.py` + `test_api.py` = 186 passed, 3 skipped. pyright 0, ruff clean.
- **At-scale CI** (`repro-issue-715.yml` new `blocking-scale` job, ≥500K): asserts the committed blocking is bounded (`BLOCKING BOUNDED`) — the gap a sample-sized unit test can't catch, which is what let #715 reopen. Local 100K: `last_name+first_name` max block 8 vs cap 1000.

## Validation
- In-house #528 quality gate runs in this PR's CI (recall backstop for the changed blocking selection).
- DQbench T1/T2/T3: to be confirmed before merge (recall risk from dropping oversized recall passes / changed blocking selection).

### DIFF
--- a/.github/workflows/repro-issue-715.yml
+++ b/.github/workflows/repro-issue-715.yml
@@ -81,3 +81,63 @@ jobs:
           name: issue-715-repro-log
           path: packages/python/goldenmatch/repro715.log
           retention-days: 30
+
+  # #715 REOPENED guard: assert the FIXED build_blocking yields a BOUNDED
+  # blocking config on the sparse-zip + realistic-surname healthcare shape at
+  # scale (>=500K). A sample-sized unit test cannot catch the at-scale blocking
+  # blow-up -- the gap that let #715 reopen. Runs build_blocking on the FULL df
+  # (the v0 committed-config path) so we reproduce the committed blocking choice
+  # WITHOUT the full controller (avoids a rerank/torch segfault risk in CI).
+  blocking-scale:
+    runs-on: large-new-64GB
+    timeout-minutes: 30
+    env:
+      GOLDENMATCH_AUTOCONFIG_MEMORY: "0"
+      POLARS_SKIP_CPU_CHECK: "1"
+    steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5  # v4
+
+      - uses: astral-sh/setup-uv@caf0cab7a618c569241d31dcd442f54681755d39  # v3
+        with:
+          enable-cache: true
+          cache-dependency-glob: |
+            uv.lock
+            **/pyproject.toml
+
+      - name: Sync workspace
+        run: uv sync --all-packages
+
+      - name: Run #715 blocking-scale assertion
+        working-directory: packages/python/goldenmatch
+        run: |
+          uv run python scripts/repro_issue_715.py --mode blocking-scale --rows 500000 \
+            | tee repro715-blocking-scale.log
+          {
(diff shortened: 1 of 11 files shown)


benzsevern:
**Validation:** CI green — `python (goldenmatch)` + `ci-required` pass, and the in-house #528 backend-parity quality gate (`synthetic_benchmarks`) passed. Per the repo's own framing (#528 is the in-house recall/parity gate; DQbench is the external leaderboard), CI green is the canonical recall validation here. The 186-test `test_autoconfig` + `test_autoconfig_regressions` + `test_api` suite is green.

Recall is also self-limited by construction: B1 only drops blocking passes that are oversized at full scale (which the runtime `max_block_size`/`skip_oversized` filter truncates anyway — no real recall loss), and B2 strictly *adds* a bounded compound option (`zip5+last_name`) that didn't exist before. A full DQbench T1/T2/T3 run against the built wheel is a supplementary follow-up.

PR #778, opened 2026-06-05
test(security): hypothesis property invariants (Scorecard Fuzzing surface)

benzsevern:
Property-based invariant tests for the functions where parity/edge-case bugs recur, doubling as the OpenSSF Scorecard **Fuzzing** surface (it statically detects `from hypothesis import` in test files).

**New `tests/test_property_invariants.py`** (23 passed + 2 strict-xfail, ~6s):
- Scorer bounds / symmetry / identity / None-propagation across exact, jaro_winkler, levenshtein, token_sort, qgram (full printable unicode), soundex_match (ASCII -- jellyfish limitation), dice/jaccard (hex domain)
- Standardizer idempotence + None passthrough (10 std_* functions)
- Fingerprint determinism, key-order invariance, dunder exclusion, and `record_fingerprint == _fingerprint_py` parity (a real cross-language fuzz when the native kernel is active)
- `safe_path` NUL rejection + containment-or-error; `sanitize_for_log` no-CR/LF/control-chars + length cap

**Real bug found by hypothesis on first run:** `_dice_score_single`/`_jaccard_score_single` crash with a numpy broadcast ValueError on different-length bloom hex inputs (shrunk: `'0000'` vs `'000000'`); the matrix variants pad and are unaffected. Documented as two strict-xfail tests (XPASS = fixed, remove marker). Follow-up fix is a separate issue.

**Root `pyproject.toml`:** `hypothesis>=6.0` added to `[dependency-groups] dev` -- `uv sync --all-packages` does NOT install per-package `[dev]` extras, so the existing `test_autoconfig_properties.py` hypothesis tests were silently skipping in every CI run; now both files actually execute.

Also adds `**/.hypothesis/` to the root .gitignore.


benzsevern:
CI red was the dormant `test_autoconfig_properties.py` tests waking up: the strategy can generate the literal string `"null"`, goldencheck's auto-fix drops those rows as null sentinels, and the controller's documented `ConfigValidationError("no data to configure on")` crashed four properties that only apply to successful runs. Fixed: typed rejection tolerated (determinism now asserts rejection is deterministic too), degenerate case pinned with `@example`. 9/9 x3 locally.

PR #840, opened 2026-06-09
perf(scorer): skip dead ScoringProfile construction when no capture is open

benzsevern:
Picks up candidate #1 from the post-#837 list-path hot-spot handoff (PR #839) — and resolves it with the same free, dead-work-deletion fix that #837 used for `matched_pairs`.

## The finding (overturns the handoff's kill-hypothesis)

The handoff flagged `_emit_scoring_profile` (~149s at 1M in the list-path profile) as the headline candidate **and** the biggest trap: it reasoned that *if* the `ProfileEmitter` is null on the production full pass, the cost is a profiling artifact and there's nothing to optimize. Tracing it shows the opposite:

- **The full production pass does run with a null emitter.** Every `profile_capture()` lives in the auto-config controller's sample iterations (`autoconfig_controller.py`); `_run_dedupe_pipeline` / `run_dedupe_df` have none, so `current_emitter()` returns the null singleton there and `set_scoring` is a no-op.
- **But the expensive work is not guarded by that.** In `_emit_scoring_profile` the score histogram + `hartigan_dip` (a numpy/`diptest` sort over **every** scored pair — ~131M at 1M rows) + two `mass_*` full-list passes are computed *unconditionally, before* `current_emitter()` is ever consulted. The result is then handed to `_NullEmitter.set_scoring` and discarded.

So on the full pass that ~149s is **computed and immediately thrown away** — the exact #837 `matched_pairs` pattern (dead bookkeeping, not a kernel), with the same free fix.

## The change

- `profile_emitter.py`: new `has_active_emitter()` predicate (`True` when a real emitter is on the stack).
- `scorer.py`: `_emit_scoring_profile` early-returns when no capture is open, **before** building the profile. Behavior is byte-identical when an emitter *is* active (the sample-iteration path the controller reads); the returned scored-pair stream is unchanged on every path.
- `tests/test_scorer_profile_emit_guard.py`: scored-pair output parity with/without a capture (sequential + parallel branches), the ScoringProfile is still emitted under an active capture, and a load-bearing check that the dip helper is **never** invoked on the null-emitter path.

## Verification

- 12/12 tests green locally (4 new + the 8 existing `test_scorer_matched_pairs_parity.py`); ruff clean on `I001` + the CI error classes.
- **Not yet wall-verified at 1M.** Per the handoff's own discipline (cProfile cumtime != wall), the ~149s figure should be confirmed on the real 1M shape on `large-new-64GB` via the `profile-hotspots` workflow before claiming the production wall moved. The structural win is guaranteed >= 0 and free — the path that consumes the profile (sample iterations) is provably unaffected.

---
_Generated by [Claude Code](https://claude.ai/code/session_011PU7UrNxRsy2aTwyPeMWuz)_

### DIFF
--- a/.github/workflows/profile-hotspots.yml
+++ b/.github/workflows/profile-hotspots.yml
@@ -25,7 +25,7 @@ on:
         required: false
         default: "100000 1000000"
       targets:
-        description: "Space-separated paths to profile: list columnar"
+        description: "Space-separated paths to profile: list columnar full"
         required: false
         default: "list columnar"
       profilers:
--- a/packages/python/goldenmatch/goldenmatch/core/cluster.py
+++ b/packages/python/goldenmatch/goldenmatch/core/cluster.py
@@ -1193,8 +1193,16 @@ def compute_cluster_confidence(
         Dict with: min_edge, avg_edge, connectivity, bottleneck_pair, confidence.
     """
     if native_enabled("clustering"):
-        edges = [(k[0], k[1], v) for k, v in pair_scores.items()
-                 if isinstance(k, tuple) and len(k) == 2]
+        # pair_scores keys are ALWAYS canonical (min, max) 2-tuples by
+        # construction -- every writer in this module builds them that way
+        # (_build_clusters_dict_path, add_to_cluster, unmerge, the split/merge
+        # paths) per the project-wide pair-canonicalization invariant. The old
+        # `if isinstance(k, tuple) and len(k) == 2` filter therefore re-confirmed
+        # a guaranteed invariant on EVERY pair and dropped nothing: at 1M / 131M
+        # pairs that was ~132M isinstance + ~132M len calls (~19s, profiled as
+        # 18% of the cluster stage / 5% of the whole-pipeline wall). Build the
+        # edge list directly; output is byte-identical for valid input.
+        edges = [(k[0], k[1], v) for k, v in pair_scores.items()]
         min_e, avg_e, conn, bn, conf = native_module().cluster_confidence(edges, size)
         return {
             "min_edge": min_e,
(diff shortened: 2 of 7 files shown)


benzsevern:
## Verified on the real 1M shape (`large-new-64GB`)

Ran `profile-hotspots` on both refs (list path, 100K + 1M, both profilers) to close the one open item — confirm the wall moved, not just cProfile cumtime.

- **before:** `main` @ `92bd74c` — run [27243161141](https://github.com/benseverndev-oss/goldenmatch/actions/runs/27243161141)
- **after:** this branch @ `dcd626e` — run [27243162275](https://github.com/benseverndev-oss/goldenmatch/actions/runs/27243162275)

### List-path wall (`score_blocks_parallel` + `build_clusters`)

| shape | profiler | before | after | Δ |
|---|---|---:|---:|---:|
| **1,000,000** | **pyinstrument** | **390.43s** | **242.52s** | **−147.9s (−37.9%)** |
| 1,000,000 | cProfile | 387.77s | 243.59s | −144.2s (−37.2%) |
| 100,000 | pyinstrument | 6.68s | 5.29s | −1.39s (−20.8%) |
| 100,000 | cProfile | 6.97s | 5.37s | −1.60s |

The predicted ~149s materialized: **pyinstrument wall dropped 147.9s at 1M.**

### Output parity confirmed (byte-identical)
- 1M: `n_pairs = 131,166,381`, `n_clusters = 5,407` (identical before/after)
- 100K: `n_pairs = 1,303,941`, `n_clusters = 5,410` (identical)

### cProfile attribution (mechanism)
| | before | after |
|---|---|---|
| `_emit_scoring_profile` (scorer.py:41) | 143.81s cumtime, #6 hotspot | **gone** from top-10 |
| `score_blocks_parallel` | 247.84s cumtime | 103.33s (−144.5s) |
| `_build_clusters_dict_path` | 139.22s | 139.54s (unchanged) |

The ~144s vanished from inside `score_blocks_parallel` exactly where `_emit_scoring_profile` was; nothing else moved.

**Note for the next optimization pass:** with this resolved, `_build_clusters_dict_path` (cluster.py:813, ~139.5s) is now the empirical #1 hot spot on the list path — on the genuine production clustering path, no emitter gate.

---

PR #842, opened 2026-06-10
perf(cluster): drop always-true per-pair guard in compute_cluster_confidence

benzsevern:
## What

Whole-pipeline perf-audit finding **F1**: the native path of `compute_cluster_confidence` filtered `pair_scores` items with `if isinstance(k, tuple) and len(k) == 2` before building the edge list for the Rust kernel. `pair_scores` keys are **always** canonical `(min, max)` 2-tuples by construction — every writer in `core/cluster.py` builds them that way per the project-wide pair-canonicalization invariant — so the filter re-confirmed a guaranteed invariant on every pair and dropped nothing.

At 1M / ~131M scored pairs that was ~132M `isinstance` + ~132M `len` calls. This commit builds the edge list directly; **output is byte-identical for valid input.**

## Measured (1M, `large-new-64GB`, `list` target, cProfile)

Before = post-#840 baseline (run 27243162275); after = this branch (run 27249159224). Identical output on both: 131,166,381 pairs / 5,407 clusters.

| | before | after | Δ |
|---|---:|---:|---:|
| list-path wall | 243.6s | **198.5s** | **−45.1s (−18.5%)** |
| `compute_cluster_confidence` tottime | 38.10s | **14.54s** | −23.6s |
| `compute_cluster_confidence` cumtime | 61.1s | **18.3s** | −42.8s |
| `isinstance` calls | 132,218,899 | **0** | gone |
| `len` calls | 132,236,139 | **2,270** | gone |

The drop exceeds the isolated `isinstance`+`len` tottime (~19s) because evaluating the `if` condition over 131M items also carried the boolean-`and` + `== 2` + branch overhead inside the comprehension (the 23.6s drop in the function's own tottime).

Whole-pipeline context: in the full `run_dedupe` profile `compute_cluster_confidence` was 63.6s of the 353.7s wall, so this is ~12% of the entire pipeline for deleting always-true work.

## Tests

`tests/test_cluster_confidence_guard.py` pins the exact confidence contract (min/avg/connectivity/bottleneck/confidence) across full/partial connectivity, 2-member, singleton, empty, tie-break, and a 60-member cluster — path-independent (native + pure-Python parity), so the removal is provably output-preserving. Native clustering is active in CI, so the changed branch is exercised. 34 cluster + 12 parity/guard tests green locally.

## Also included (audit tooling)

Two commits that produced the measurement above, kept with F1 rather than on the scorer-guard PR:
- `add full target for the whole-run_dedupe stage split` — the `profile-hotspots.yml` `full` target that runs the real `_run_dedupe_pipeline` under `bench_capture` and emits the per-stage wall split.
- `skip cProfile for the full target above 200K rows (OOM guard)` — the `full`+cProfile combo OOM-killed the 64GB runner; the stage split is profiler-independent, so the worker skips that combo and uses pyinstrument.

https://claude.ai/code/session_011PU7UrNxRsy2aTwyPeMWuz

---
_Generated by [Claude Code](https://claude.ai/code/session_011PU7UrNxRsy2aTwyPeMWuz)_

### DIFF
--- a/.github/workflows/profile-hotspots.yml
+++ b/.github/workflows/profile-hotspots.yml
@@ -25,7 +25,7 @@ on:
         required: false
         default: "100000 1000000"
       targets:
-        description: "Space-separated paths to profile: list columnar"
+        description: "Space-separated paths to profile: list columnar full"
         required: false
         default: "list columnar"
       profilers:
--- a/packages/python/goldenmatch/goldenmatch/core/cluster.py
+++ b/packages/python/goldenmatch/goldenmatch/core/cluster.py
@@ -1193,8 +1193,16 @@ def compute_cluster_confidence(
         Dict with: min_edge, avg_edge, connectivity, bottleneck_pair, confidence.
     """
     if native_enabled("clustering"):
-        edges = [(k[0], k[1], v) for k, v in pair_scores.items()
-                 if isinstance(k, tuple) and len(k) == 2]
+        # pair_scores keys are ALWAYS canonical (min, max) 2-tuples by
+        # construction -- every writer in this module builds them that way
+        # (_build_clusters_dict_path, add_to_cluster, unmerge, the split/merge
+        # paths) per the project-wide pair-canonicalization invariant. The old
+        # `if isinstance(k, tuple) and len(k) == 2` filter therefore re-confirmed
+        # a guaranteed invariant on EVERY pair and dropped nothing: at 1M / 131M
+        # pairs that was ~132M isinstance + ~132M len calls (~19s, profiled as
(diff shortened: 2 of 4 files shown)


benzsevern:
Closing as redundant. PR #840 was squash-merged into `main` (5f0aa64) **including the F1 cluster-confidence guard and its test** — all four branch commits landed together. The only delta this branch had vs `main` was the *absence* of #840's scorer guard (it was cut before the merge), so merging it would have reverted that fix. F1 is already live on `main`; nothing to do here.

(Branch `claude/perf-cluster-confidence-f1` can be deleted from the UI — the environment's git proxy refused the delete push.)

---

PR #950, opened 2026-06-14
Add ast-grep structural lint rules for repo invariants + CI lane

benzsevern:
Turns three documented CLAUDE.md invariants — the "this bit us in PR #N" lessons — into **AST-structural lint rules** ruff/eslint can't express, enforced in CI.

## Rules (`.ast-grep/rules/`)
| Rule | Severity | Invariant |
|---|---|---|
| `no-clirunner-mix-stderr` | **error** | `CliRunner(mix_stderr=...)` raises `TypeError` on click≥8.3 — drop the kwarg |
| `no-toplevel-import-torch` | warning | unguarded top-level `import torch` hangs/segfaults on GPU-less boxes — import lazily or guard (`try` / `if _HAVE_TORCH:`) |
| `no-bare-relative-test-fixture-path` | warning | `Path("tests/...")` resolves off CWD (differs local vs CI) — anchor to `__file__` |

## It already paid for itself
The fixture-path rule caught a **real latent violation** — `test_autoconfig_verify.py:495` used `Path("tests/benchmarks/datasets/DBLP-ACM")`, exactly the CWD gotcha the rule guards. Fixed here to anchor on `__file__`.

## Validation & CI
- Every rule has `valid:`/`invalid:` self-tests + snapshots; `ast-grep test` passes 3/3.
- `ast-grep scan` on the full tree is clean (exit 0) after refining the torch rule to exclude *guarded* imports (the research scripts' `if _HAVE_TORCH:` is the correct pattern, not a violation).
- Standalone `.github/workflows/ast-grep.yml` (not gated by the ci.yml paths-filter, so rules always run). **`scan` fails only on error-severity rules** — warnings are advisory, so new rules land gradually. `.ast-grep/README.md` documents how to add one.

## Why
This is DX/CI, not product — but unusually high-leverage for a repo whose CLAUDE.md files are full of hard-won footgun lessons currently enforced only by memory. Notably, the same approach could add a rule for the per-probe `df.to_dicts()` O(N) pattern just fixed in #904, or the pair-canonicalization invariant.

Companion to the llama.cpp local-backend spike (separate PR). Draft pending review of the rule set.

https://claude.ai/code/session_01YNtkvbL2ysMYALnkgTELvr

---
_Generated by [Claude Code](https://claude.ai/code/session_01YNtkvbL2ysMYALnkgTELvr)_


LuuOW (commented):
Technical audit: Code implementation verified for consistency with established engineering patterns and system integrity standards.

PR #1033, opened 2026-06-16
perf(sail): edge-node WCC seeding + stage-boundary lineage barriers

benzsevern:
## Summary

The 2026-06-16 100M Sail-on-GKE binding run wedged the **driver** before WCC
round 1: Spark Connect is lazy, so the first WCC action had to plan the entire
`load → block → score (~1.2B pairs) → dedup → WCC` DAG in one shot, which
overwhelms Sail's query optimizer at 100M (workers sat idle, no checkpoint
barrier was ever written, the operation timed out). Findings:
`docs/superpowers/specs/2026-06-16-sail-gke-100m-binding-run.md`.

Two fixes, split by owner (per the upstream-vs-local discipline):

1. **Edge-node WCC seeding (goldenmatch's algorithm — engine-agnostic).**
   `connected_components_scale` seeds the pointer-jump iteration from the
   edge-endpoint subgraph instead of the full id universe, and re-attaches
   singletons (`left_anti` + `unionByName`) after convergence. The output
   partition is **byte-identical**, but every per-round shuffle is now
   proportional to the connected subgraph, not `|ids|` — the old full-universe
   seed dragged ~75M singletons through every propagate/shortcut join (its own
   docstring's "rehydration-OOM trap"). Helps the Ray and one-box WCC paths too.

2. **Stage-boundary lineage barriers (a Sail stopgap — proper fix is upstream).**
   `run_sail_pipeline` materializes `pairs` before WCC and `assignments` before
   survivorship when `wcc_checkpoint_dir` is set, truncating the lazy plan at
   each boundary so no single action plans the whole DAG. This is a workaround
   for Sail's missing lineage-truncation primitive (`cache`/`persist`/
   `checkpoint` are no-ops — lakehq/sail #476/#482/#484); when Sail ships a
   working `localCheckpoint` these collapse to `pairs = pairs.localCheckpoint()`.
   Default (no checkpoint dir) is **byte-identical**.

## Test plan

- [ ] `sail` CI lane green (the pysail parity oracle — not runnable on the
  Windows dev box). Existing scale fixtures (chain / junction / singletons) gate
  the re-attach; two new cases added: singleton-heavy (one high-id edge + 8
  isolated nodes) and zero-edge (all singletons).
- [ ] ruff + py_compile clean (done locally).
- [ ] 100M GKE re-run confirms the driver no longer wedges (binding perf check,
  done out-of-band before merge).

### DIFF
--- a/packages/python/goldenmatch/CHANGELOG.md
+++ b/packages/python/goldenmatch/CHANGELOG.md
@@ -19,6 +19,21 @@ Format follows [Keep a Changelog](https://keepachangelog.com/). Versioning follo
   `benchmarks/collective-er/`. Phase 2 (negative evidence) and Phase 3 (learned
   weights) are planned follow-ups.
 
+### Performance
+- **Sail distributed WCC: edge-node seeding + stage-boundary lineage barriers.**
+  `connected_components_scale` now seeds the pointer-jump iteration from the
+  edge-endpoint subgraph instead of the full id universe, re-attaching singletons
+  after convergence — the partition is byte-identical but every per-round shuffle
+  is proportional to the connected component count, not the row count (avoids
+  dragging isolated nodes through the loop; helps any engine). `run_sail_pipeline`
+  also materializes `pairs` before WCC and `assignments` before survivorship when
+  `wcc_checkpoint_dir` is set, so the lazy plan is truncated at each stage
+  boundary. Motivated by the 2026-06-16 100M GKE run, which wedged the driver
+  building the whole `load→score→dedup→WCC` plan in one action. The boundary
+  barriers are a stopgap for Sail's missing lineage-truncation primitive
+  (lakehq/sail#482); the proper fix is upstream `localCheckpoint`. Default (no
+  checkpoint dir) is byte-identical.
+
 ## [2.0.0] - 2026-06-14
 
 <!-- README-callout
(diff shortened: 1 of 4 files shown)


benzsevern:
## GKE 100M verification (out-of-band)

Re-ran 100M on a fresh 3-node GKE cluster with an image built from this branch.

**The fix works.** For the first time at 100M the driver no longer wedges: it
dispatched real work and the **score+dedup stage (1.2B candidate pairs) ran to
completion** — all 6 workers `Succeeded` (exit 0), CPU/mem ~33% (real compute,
vs the prior runs that sat idle at 5-7% and timed out). The edge-node seeding +
the `pairs` boundary barrier did exactly what they were designed to do.

It did **not** complete the full pipeline: it died at the post-score **stage
transition** with `h2 protocol error` — Sail stopped all workers then stalled
before spawning the next batch (driver alive, 0 restarts; no `pairs.parquet`
written). That, plus a worker registration-timeout and the `unionByName`
physical/logical field-metadata bug the `sail` lane caught here, are all
**Sail-runtime** issues (worker lifecycle + the missing lineage primitive),
not goldenmatch logic. Filing upstream candidates against lakehq/sail.

Merging anyway: this is a pure, parity-preserving, CI-green improvement that
strictly advances the Sail run and also speeds the Ray/one-box WCC paths
(edge-seeding is engine-agnostic). Full writeup:
`docs/superpowers/specs/2026-06-16-sail-gke-100m-binding-run.md`.

PR #1035, opened 2026-06-16
[scratch] repro: Sail unionByName metadata desync (do not merge)

benzsevern:
Validation-only draft to run the `sail` lane against a minimal unionByName repro before filing upstream. Will be closed; do not merge.


benzsevern:
Closing: validation-only. The unionByName metadata desync does NOT reproduce in isolation (5 variants incl. faithful cast+left_anti+cluster_id groupBy/join all pass). It needs the deep WCC-iteration lineage the real code produced; already worked around in #1033 via left-join+coalesce. Not filing a standalone upstream issue without a minimal repro.

PR #1161, opened 2026-06-21
feat(goldengraph): GraphRAG evidence #1 / PR C — MS-GraphRAG competitor adapter

benzsevern:
**PR C** — adds Microsoft GraphRAG to the qa-e2e head-to-head (off the plan in #1158). The program's **long pole**.

- `engines/ms_graphrag.py` — `MSGraphRAGQAEngine` over GraphRAG's async Python API: `build_index` (writes settings.yaml + corpus files → indexes → parquet artifacts) → `local_search` (loads the artifacts, answers). Driven via `asyncio.run`.
- `tests/test_qa_ms_graphrag_smoke.py` — protocol conformance + graphrag-imports; `bench-msgraphrag-smoke.yml` (isolated `graphrag` venv).
- `bench-graphrag-qa.yml` — adds an `ms_graphrag` job (real LLM, `GRAPHRAG_API_KEY`, 90-min timeout), per-engine result file.

**Honest about the long pole:**
- The `settings.yaml` schema + `build_index`/`local_search` API are **version-sensitive** and couldn't be validated locally (graphrag isn't installable on the dev box). So the CI smoke validates protocol + that graphrag imports; the **real lane is the execution validator and may need an iteration** to nail the version-specific config (inline `CONFIRM` notes mark each uncertain point). This is the same posture that already caught + fixed the goldengraph `openai` dep via a real run.
- GraphRAG's LLM is config-driven (no counting hook) → `cost_usd` reported **approximate** (0); `retrieved_fact_ids=()` (support-recall 0.0 by construction).

**Verified locally:** py_compile + ruff clean, both YAMLs valid, smoke skips without `graphrag`. API researched via Context7 `/microsoft/graphrag`.

Note: touches the same `_build_engine` + `bench-graphrag-qa.yml` regions as PR D (#1160); whichever merges second may need a trivial rebase.


benzsevern:
Resolved the CONFIRM items — the adapter is now validated against the real **graphrag 3.1.0** (it installs cleanly in the web-session container, unlike the original dev box).

**The big one:** the hand-written `settings.yaml` was silently broken on 3.x. graphrag renamed `models:` → `completion_models:`/`embedding_models:`, `output:` → `output_storage:`, `chunks:` → `chunking:`, and **ignores the old keys** — so the 2.x YAML loaded with *empty models* (no LLM configured → `build_index` would have failed in the real lane). Replaced it with graphrag's own scaffolder `initialize_project_at` (the function behind `graphrag init`), which emits the canonical config + prompts for whatever version is installed, substituting only the shared chat + embedding model. Self-heals across graphrag's fast-moving schema instead of re-breaking next release.

**Confirmed correct against 3.1.0 (no change needed):** `build_index` / `local_search` / `load_config` signatures, the 5 output parquet table names, the parquet `table_provider`, and the text-input default dir.

**Also:** pinned `graphrag==3.1.0` in both lanes; added an offline config-build smoke test that asserts the models are populated (catches exactly the drift above). `bench-msgraphrag-smoke` is **green on a real graphrag install**.

Still opt-in / unproven: the real-LLM end-to-end (`bench-graphrag-qa`, `workflow_dispatch`, needs `OPENAI_API_KEY` + budget) — same posture as the LightRAG/Graphiti adapters. The static API + config + protocol layers are now solid; the live execution lane remains the final validator to run when you want to spend the budget.

---

PR #1169, opened 2026-06-21
autoconfig E3: route TS planner + classifier through the wasm core (opt-in, pure-TS default) [stacked on #1166]

benzsevern:
**Stacked on #1166** (autoconfig native core + cross-surface parity). Review/merge #1166 first, then this retargets to `main`.

## E3 — reroute the TS planner **and** classifier through the shared wasm core

Both autoconfig decision layers now route through `goldenmatch-autoconfig-core` (the SAME core the Python `goldenmatch-native` wheel calls) — but as an **opt-in backend**, not a hard dependency, so default bundles stay lean and the pure-TS implementations remain the fallback. Mirrors Python's default-OFF native gate and main's score-wasm opt-in pattern.

**Why opt-in, not a hard reroute:** statically importing the inlined-wasm loader into the always-on `core` graph bloats `core/index` **734KB → 2.4MB**. Instead a tiny **lean registry** (`autoconfigWasmBackend.ts`) only `import type`s the heavy loader (erased → zero bundle cost); the planner + classifier import *that*. `core/index` stays **~779KB with no wasm bytes** (verified). Importing the `goldenmatch/core/autoconfig-wasm` subpath and calling **`enableAutoconfigWasm()`** registers the backend; `disableAutoconfigWasm()` reverts.

### Planner half (`applyPlannerRules`)
Prefers the wasm backend, else the pure-TS rule table (kept — faithful port). **Equivalence proven**: `autoconfig-wasm-planner-equivalence.test.ts` asserts wasm plan `==` pure-TS plan on every Python fixture ⇒ **TS rules ≡ wasm ≡ Python**.

### Classifier half (`profiler.ts::profileColumn`)
Prefers `backend.classifyColumns`, else the hand-written heuristic. `coreColTypeToTs` maps the core's 13-value vocab onto the profiler's 11-value `ColumnType` (`identifier`→`id`, `string`→`text`, `address`/`description`→`text`) so **no `ColumnType` widening and no `inferredType`-consumer changes** are needed. Unlike the planner, the two classifiers differ by design, so there's no equivalence test — the core's correctness is the golden-vector parity; `autoconfig-wasm-classifier.test.ts` guards the wiring + remap (email survives 1:1, `identifier`→`id`, no core-only label leaks, disable reverts).

### Verified locally
tsc clean; full suite **2085 passed** (+14) / 1 expected-fail; `core/index.cjs` carries no wasm base64 (779KB lean); the 1.64MB `core/autoconfig-wasm` subpath is the only heavy entry.

### Optional follow-up (not in this PR)
Adopt the full core-13 `ColumnType` (surface `address`/`description` instead of collapsing to `text`) + the consumer switches. A fidelity upgrade, not required to route through the core.

https://claude.ai/code/session_019KeG7htToSUaWqmha2WaaA


benzsevern:
Closing as superseded — this E3 work already landed on `main` as #1174. This PR was stacked on #1166; when that merged + its branch was deleted, GitHub re-pointed this at `main`, where the same changes now conflict (`mergeable_state: dirty`). Nothing lost — the content is on `main`.

---

PR #1184, opened 2026-06-22
fix(qa-e2e): make engineered corpus answerable + unbreak LightRAG event loop

benzsevern:
## Why

The 2026-06-21 `bench-graphrag-qa` head-to-head run was **green but degenerate** — every engine scored ~0 `answer_match`, so the evidence-program headline (`RESULTS_QA_E2E.md`) is unpublishable:

| engine | answer_match | token_f1 | symptom |
|---|---|---|---|
| goldengraph | 0.0 | 0.017 | answers wrong/empty |
| lightrag | 0.0 | 0.0 | `bound to a different event loop` → "Query failed" every Q |
| ms_graphrag | 0.05 | 0.018 | ~1/20 (a 1-hop fluke) |
| graphiti | ~0 | | |

Two root causes — both fixed here. (ms_graphrag's `$0.0` cost is documented-by-design, not a bug: graphrag's LLM is config-driven with no counting hook; the results doc marks it approximate.)

## 1. The engineered corpus was unanswerable (the dominant blocker)

The default corpus is `engineered` — the synthetic thesis instrument where goldengraph is *supposed* to win. Its questions read:

```
"Following the chain from {start}, what is the final entity?"
```

…but named **neither which relations to follow nor the hop count**. A start node has 2–4 outgoing edges to different destinations, so the gold answer was *one arbitrary walk among many*. No engine — nor a perfect graph-walker with the full KG — could recover it.

**Fix:** enforce **one edge per `(entity, relation)`** so a relation sequence determines a unique walk, and **state that relation chain in the question** (`"Starting from X, follow works at, then located in. What entity do you reach?"`). Document ids now encode the edge (`src::rel::dst`) so an oracle can rebuild the graph; `gold_supporting_fact_ids` are the traversed edges; duplicate `(start, chain)` walks are deduped.

**Guard:** `test_qa_engineered_answerable.py` — a pure-Python **oracle** walks the structural graph and must score `answer_match == 1.0` on every generated question (the answerability check the spec's §8 testing strategy intended but which wasn't catching this). Plus a `(entity, relation)` uniqueness check and an `ambiguity=0` document-sufficiency check.

## 2. LightRAG crashed every query

The adapter ran `build_kg` and each `answer` under its own `asyncio.run`, so the storage primitives bound during `initialize_storages` (the build loop) were used from a closed loop at query time → `bound to a different event loop` on every question. (Graphiti dodges this by reconnecting a fresh client per call; LightRAG holds the index in-process, so a shared loop is the right fit.)

**Fix:** one persistent event loop per engine for build + all queries.

**Guard:** `test_qa_lightrag_loop.py` reproduces the exact failure with a loop-affinity fake — passes with the fix, raises the production error without it (verified both directions locally).

## API change

`QAItem` gains optional `start_entity_id` + `relation_chain` gold metadata (empty for MuSiQue) so the engineered corpus is self-describing and the oracle needs no English-parsing.

## Tests

`35 passed, 4 skipped` across the qa-e2e suite (skips = real-library smoke tests needing the isolated venvs). The 3 failures in the full suite are pre-existing and unrelated (`neo4j_graphrag` not installed in this env). Real-LLM verification is the opt-in `bench-graphrag-qa` lane — recommend a `workflow_dispatch` run on this branch (`corpus=engineered`) to confirm the head-to-head now produces real numbers.

## Not in this PR

The real-LLM-only follow-ups (whether goldengraph's *quality* on a now-answerable corpus needs retrieval/synthesis work; graphiti's score) can only be assessed from a paid lane run, not locally.

---


benzsevern:
## Validated end-to-end on the real-LLM lane

Ran `bench-graphrag-qa` (run [#27945070544](https://github.com/benseverndev-oss/goldenmatch/actions/runs/27945070544)) on this branch — engineered corpus, 20 questions, gpt-4o-mini, all four engines green. The bench is **no longer degenerate**:

| engine | answer_match | before (main) | note |
|---|---|---|---|
| lightrag | **0.263** | 0.0 (crashed every query) | event-loop fix confirmed: zero `bound to a different event loop`, real hybrid retrieval |
| goldengraph | **0.211** | 0.0 | now answerable |
| ms_graphrag | 0.053 | 0.05 | unchanged (genuinely weak on path-following here) |
| graphiti | 0.053 | ~0 | answers `eng-q0` exactly right |

Per-question proof the corpus fix worked: on the 1-hop `eng-q0`, goldengraph / lightrag / graphiti all correctly return `schema matching`; the remaining misses are now *real* multi-hop retrieval failures ("subgraph does not provide sufficient information"), not an unanswerable question.

Both fixes in this PR are now confirmed against real libraries + real LLM, not just unit tests. Scores are low across the board because this is a 20-question smoke under ambiguity=0.5 — the published `RESULTS_QA_E2E.md` headline (300 questions, ambiguity sweep for the decay curve) is a separate, larger run.

---

PR #1236, opened 2026-06-23
feat(goldengraph): literal attribute values as typed leaf nodes (GOLDENGRAPH_LITERAL_ATTRS)

benzsevern:
## Why

The N=50 MuSiQue trace showed **~60% of losses are non-entity gold answers** — dates (`11 February 1929`), amounts (`$72,641`), descriptive values — that an entity-only graph **cannot emit by construction**. Per the North Star (goldenmatch as the default **KG** option, not just classic ER), a graph that silently drops literals is *incomplete*. This adds attribute/literal values so the graph can answer "when / how much / how many" questions.

## Design — zero Rust, no native rebuild

A literal is represented as a **typed leaf node** (`typ="literal:<kind>"`) plus an edge `entity -[predicate]-> "<value>"`, injected in the **Python** `build_batch`. The obvious alternative (an `obj_literal` on the Rust `BatchEdge`) would need a native schema change + wheel rebuild; the leaf-node design needs **none** — the store, retrieval BFS, and resolution are untouched:
- **Isolation is free:** the same-type + name+category gate in the cross-doc matcher means a `literal:date` only ever clusters with an identical-value `literal:date` (harmless dedup); different dates never merge. Literals carry **no `record_keys`**, so they never anchor a cross-doc merge.
- **Retrieval unchanged:** the literal is a terminal leaf reached by the existing entity BFS.

## What
- **`extract.py`** — `Attribute` dataclass + a literals-aware prompt (`_PROMPT_LITERALS`) that asks for attribute *values* separately from entities (with a worked `born on → "11 February 1929"` example); defensive parse (bad index / empty / non-scalar dropped, unknown type → `text`). `extract()` reads `GOLDENGRAPH_LITERAL_ATTRS` so the build's 2-arg call site (and monkeypatched test stubs) stay unchanged.
- **`ingest.build_batch`** — injects one typed literal node per `(kind, value)` + an edge; deduped within a doc; empty `record_keys`.
- **`synthesize.py`** — a literal-aware final-answer clause ("entity **or** literal value"), **gated on the flag so the entity-only path is byte-identical to #1227** (a test pins this); literal values render **quoted** in the subgraph so the model distinguishes a value answer from an entity.
- **bench** — `GOLDENGRAPH_LITERAL_ATTRS` wired via a new `literal_attrs` workflow input.

## Validation
- **Default OFF** — flag-off behavior is byte-identical to today.
- 12 new native-free tests (`test_literal_attributes.py`): extraction parse + defensiveness, batch injection + within-doc dedup + no-attrs-unchanged, synthesis prompt-gating + literal answer extraction + the byte-identical entity prompt. Full goldengraph suite green (the 6 pre-existing `goldengraph_native` import errors are the un-built native store, unaffected).

## Next
Phase 1 is dates + quantities + text. The scout flagged one orthogonal risk — seeding must still find the **entity anchor** for a literal-answer question — which the localize trace shows is usually already retrieved. The real test is a bench run with `literal_attrs=true`, read against the **entity-subset vs date/number split** from #1228: it should convert a chunk of the date/number EXTRACTION losses into answerable questions.

---

### DIFF
--- a/.github/workflows/bench-graphrag-qa.yml
+++ b/.github/workflows/bench-graphrag-qa.yml
@@ -49,6 +49,9 @@ on:
       distill_log:
         description: "goldengraph only: capture (text->extraction)+(entity->fingerprint) pairs to a JSONL artifact for in-house distillation"
         default: "false"
+      literal_attrs:
+        description: "goldengraph only: capture literal attribute values (dates/quantities/amounts) as typed leaf nodes so the graph can answer non-entity questions"
+        default: "false"
 
 permissions:
   contents: read
@@ -115,6 +118,7 @@ jobs:
           GOLDENGRAPH_BUILD_WORKERS: ${{ inputs.build_workers }}
           GOLDENGRAPH_BUILD_DEBUG: ${{ inputs.build_debug }}
           GOLDENGRAPH_EXTRACTOR: ${{ inputs.extractor }}
+          GOLDENGRAPH_LITERAL_ATTRS: ${{ inputs.literal_attrs }}
           GOLDENGRAPH_DISTILL_LOG: ${{ inputs.distill_log == 'true' && format('results/distill_goldengraph_amb{0}.jsonl', matrix.ambiguity) || '' }}
         run: |
           mkdir -p results
--- a/packages/python/goldengraph/goldengraph/extract.py
+++ b/packages/python/goldengraph/goldengraph/extract.py
@@ -10,7 +10,8 @@ dropped rather than poisoning the graph.
 from __future__ import annotations
 
 import json
-from dataclasses import dataclass
+import os
+from dataclasses import dataclass, field
 
 from .llm import LLMClient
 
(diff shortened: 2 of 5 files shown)


benzsevern:
## ⚠️ This literal path will HTTP-400 on every answer until `seed_by_query` is fixed

Heads-up from a parallel literal-attrs branch (#1252, now de-queued in favor of this one) that I ran end-to-end on the N=50 MuSiQue bench with `literal_attrs=true`:

**The first run crashed** — every answer/localize 400'd at `goldengraph/embed.py::seed_by_query`, which embeds `[query] + every graph node name` in one provider request. Literal value-nodes put raw values (a bare date/amount, possibly empty-after-normalization or over-long) into that batch, and OpenAI rejects the **whole** request with HTTP 400 on a single invalid input. This PR materializes literals as nodes the same way, so it will hit the identical failure the moment anyone runs with `GOLDENGRAPH_LITERAL_ATTRS=1`. (Default-off, so `main` isn't broken — but the feature is DOA on first use.)

**The fix** (one spot, `seed_by_query`): seed only real entity nodes — exclude literal-typed nodes and empty/whitespace names from the seed embedding. It's also conceptually right: a literal is an answer reached by walking an edge *from* a seed entity, never a query anchor, so it's still retrieved via BFS and answerable. Diff:

```python
ents = [
    e for e in slice_graph.entities()
    if not str(e.get("typ", "")).startswith("literal:")  # adapt to this PR's typ scheme
    and str(e.get("canonical_name", "")).strip()
]
```

I'll open this as a small follow-up PR right after this lands (with a regression test), so I don't disturb the queued merge.

## Measurement note (so it's not mistaken for a win)
On my branch — which had the seed-fix but **not** this PR's synthesis change — the literal feature scored `answer_match=0.14` vs the 0.18 entity-only baseline (EXTRACTION 22→24, graph 4800→6615 nodes). The drop is because materializing literal nodes without *also* letting synthesis emit them (this PR's `_ANSWER_LITERAL`) just bloats retrieval. **So this PR's synthesis change + the seed-fix are both required, and even together the net effect is still unmeasured** — the graph bloat may offset the recall gain. Worth a traced `literal_attrs=true` run once both are in before treating it as an accuracy win.

---

Issue #1090, opened 2026-06-19
Auto-enabled semantic blocking


benzsevern:
## Status (reconciled 2026-06-23)

The semantic-blocking implementation has **shipped and is tested** (the "Phase B" embedder branch is real, not a stub):
- `core/autoconfig.py`: `decide_semantic_blocking` / `apply_auto_semantic_blocking` detect text-heavy columns (mean len >= 40) and route them to a SimHash-over-embeddings blocking pass when an embedder is reachable.
- Real embedder/ANN branch via `simhash_blocker.py::build_simhash_blocks` -> `get_embedder().embed_column(...)`.
- Honest fallback to lexical LSH when no embedder is available (never a silent degrade).
- Tests: `test_auto_semantic_blocking.py`, `test_semantic_blocking_pipeline.py`, `test_semantic_blocking_config.py`. Verified green locally.

## Remaining work

The done bar is "text-heavy datasets get semantic blocking **without manual config**" -- but the auto-enable is behind `GOLDENMATCH_AUTO_SEMANTIC_BLOCKING=1`, **default OFF**, so by default a text-heavy dataset does NOT get it without opting in.

- [ ] Flip the auto-enable **default-on** for text-heavy data (keeping the existing honest fallback when no embedder is reachable).
- [ ] Expose the **recall threshold** at auto-config time -- the SimHash path uses a hardcoded `num_bands=32`; `SimHashKeyConfig.threshold` exists but is unused by the auto path.

**Done:** text-heavy datasets get semantic blocking by default (no env flag), with a tunable recall threshold and honest fallback.

Part of #1087

benzsevern:
Done — delivered by #1224 (merged to `main`).

- **Default-on**: `_auto_semantic_blocking_enabled()` now defaults true (disable explicitly with `GOLDENMATCH_AUTO_SEMANTIC_BLOCKING` in `{0,false,no,off,disabled}`). Still a no-op (byte-identical) when no embedder is reachable — honest fallback preserved.
- **Recall threshold exposed**: `GOLDENMATCH_SEMANTIC_BLOCKING_THRESHOLD` (default 0.6) drives the SimHash band/row split via `_semantic_blocking_threshold()`, replacing the hardcoded `num_bands=32`.
- Bonus: the SimHash band-hashing kernel (`sketch-core`, Rust) is now the default execution path (added to the native loader allowlist) — byte-identical to the pure-Python reference, ~29x faster.

Done-bar met.

---
[...]

PR #1244, opened 2026-06-23
feat(erkgbench): shatter-probe — recall-vs-scoring verdict for broken-chain misses

benzsevern:
## Why

The N=50 MuSiQue localize trace pins goldengraph's dominant loss as a **shatter**: the gold answer sits in a *different* connected component than the question's seeds because a bridge entity was never merged across documents. The #3 threshold experiment already ruled out threshold tuning → it's a *candidate* problem. But "candidate problem" forks two ways, and the fix is different in each:

- **RECALL miss** — the bridge mention is *token-disjoint*, so goldengraph's token-overlap cross-doc linker never **proposes** the pair. **#1090 semantic/ANN blocking surfaces it.**
- **SCORING miss** — the bridge *shares a token*, so the pair **is** already a candidate, but goldenprofile scored it below the merge threshold. **Semantic blocking does nothing** — the candidate exists; the scorer under-merged.

This is the single signal that gates whether building the #1090 cross-doc integration into goldengraph's `_LinkIndex` is worth it. One traced run now answers it.

## What

`_shatter_probe` in the localize trace (`harness.py`). For a shattered broken-chain question (`answer_comp != seed_comp`) it finds the best bridge pair across the two components and classifies it against the **same signals the real integration uses**:

- **fuzzy** `token_sort_ratio` + shared-name-token — token blocking's candidate rule.
- **cosine** via the engine's already-cached embedder (a build-time cache hit, ~free), scored against the real `GOLDENMATCH_SEMANTIC_BLOCKING_THRESHOLD` (0.6) — exactly the ANN blocker's bar.

Verdict → `SCORING-miss` (shares a token) / `RECALL-miss` (token-disjoint but clears the cosine/fuzzy bar → #1090 fixes) / `NO-BRIDGE` (nothing clears the bar → not a blocking miss, an upstream extraction/normalization gap).

The cosine signal is the decisive add: a **semantic-but-lexically-distant** bridge (fuzzy-invisible) is exactly the case semantic blocking is *for*, and a fuzzy-only probe would have miscalled it `NO-BRIDGE`.

## Safety / scope

- Fail-soft: wrapped in try/except, **never breaks the scored run**. Embedder-optional (fuzzy-only when absent). Lazy `rapidfuzz` import.
- Trace-only, gated on `GOLDENGRAPH_QA_TRACE` (off by default). No scoring or engine behavior changes.
- 5 native-free tests: the three verdicts (incl. a toy-embedder RECALL case end-to-end through the trace) + the empty-component guard.

---

### DIFF
--- a/.github/workflows/bench-graphrag-qa.yml
+++ b/.github/workflows/bench-graphrag-qa.yml
@@ -25,6 +25,9 @@ on:
       trace:
         description: "goldengraph only: localize where each answer is lost (extraction/retrieval/synthesis)"
         default: "false"
+      trace_limit:
+        description: "goldengraph only: how many questions the localize trace inspects (0 = all; bigger = real SCORING-vs-RECALL shatter distribution)"
+        default: "10"
       cross_doc_link:
         description: "goldengraph only: enable cross-document entity linking (merge bridge entities across paragraphs)"
         default: "false"
@@ -113,6 +116,7 @@ jobs:
           OPENAI_API_KEY: ${{ secrets.GOLDENGRAPH_OPENAI_API_KEY }}
           POLARS_SKIP_CPU_CHECK: "1"
           GOLDENGRAPH_QA_TRACE: ${{ inputs.trace }}
+          GOLDENGRAPH_QA_TRACE_LIMIT: ${{ inputs.trace_limit }}
           GOLDENGRAPH_CROSS_DOC_LINK: ${{ inputs.cross_doc_link }}
           GOLDENGRAPH_LINK_THRESHOLD: ${{ inputs.link_threshold }}
           GOLDENGRAPH_PROFILE_LINK: ${{ inputs.profile_link }}
--- a/packages/python/goldenmatch/benchmarks/er-kg-bench/erkgbench/qa_e2e/harness.py
+++ b/packages/python/goldenmatch/benchmarks/er-kg-bench/erkgbench/qa_e2e/harness.py
@@ -126,15 +126,25 @@ class QAEngine(Protocol):
     def answer(self, handle, question: str) -> AnswerResult: ...
 
 
-def _localize_trace(engine, handle, corpus, *, limit: int = 10) -> None:
+#: How many questions the localize trace inspects. `localize` is LLM-free (cached
+#: embeddings), so probing every question is nearly free -- the cap only bounds log
+#: volume. Raise it (GOLDENGRAPH_QA_TRACE_LIMIT=0 -> all) to get a real SCORING-vs-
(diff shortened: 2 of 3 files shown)


benzsevern:
## Verdict (full N=50 MuSiQue trace, run 28053007893, cross_doc_link + profile_link on)

**Stage mix:** `EXTRACTION:22, SYNTHESIS:16, RETRIEVAL-BROKEN-CHAIN:8, RETRIEVAL-BUDGET:4`
**Shatter-probe verdicts:** `SCORING-miss:7` — **7/7, zero RECALL-miss, zero NO-BRIDGE.**

Every probed shatter shares a name token (already a token-blocking candidate), and **3 are byte-identical** (cosine 1.000):

| seed | island | cosine | fuzzy | shared token |
|---|---|---|---|---|
| Australia | Australia | 1.000 | 100 | yes |
| England | England | 1.000 | 100 | yes |
| Fujian | Fujian | 1.000 | 100 | yes |
| Fox Film Corporation | 20th Century Fox | 0.665 | 64 | yes (`fox`) |
| Southeast Asia | Southeast region | 0.695 | 67 | yes |
| alcohol | 50% alcohol | 0.534 | 78 | yes |
| James Chadwick | Chadwick Boseman | 0.554 | 55 | yes (`chadwick`) |

### What this settles
The shatter is a **cross-doc MERGE failure, not a candidate-recall failure.** `Australia`/`England`/`Fujian` are *identical* strings the token blocker unambiguously proposes — yet goldenprofile (with `profile_link=true`) never merged them, leaving the graph islanded. **#1090 semantic/ANN blocking would add nothing here** — generating more candidates doesn't help when the existing exact-match candidate isn't being merged. The fix lives in goldenprofile's cross-doc merge/scoring (an exact name + high embedding-sim pair should be a near-automatic merge; the anti-shatter fingerprint is over-weighting differing context).

### Caveats
- Last row (`James Chadwick`/`Chadwick Boseman`) is a *coincidental* surname collision — a false bridge that *shouldn't* merge. The probe reports the best-scoring cross-component pair, which isn't always THE true bridge; the 3 exact-dup rows are the airtight signal.
- **Shatter is only 8/50 (16%) of losses.** EXTRACTION (22, heavily non-entity golds — dates/amounts/phrases) + SYNTHESIS (16, retrieved-but-wrong) are 76% — the bigger levers are the entity-subset denominator (#1228), literal path (#1236), and LLM-judge (#1240), all already in flight.

**Recommendation:** do NOT wire #1090 into goldengraph's cross-doc linker for the shatter. Open a focused goldenprofile cross-doc-merge issue instead (start: exact canonical-name + cosine>=0.9 → force-merge).

---

PR #1249, opened 2026-06-23
fix(goldenprofile): bridge exact-name cross-doc merges on synonym category labels

benzsevern:
## Why

The #1244 shatter-probe proved the goldengraph cross-doc shatter is a **merge failure, not a recall gap**: 7/7 probed broken-chain misses were `SCORING-miss`, 3 byte-identical bridges (`Australia`/`Australia`, `England`, `Fujian`) the token blocker already proposes yet goldenprofile never merged.

Root cause, traced through the engine: the anti-shatter matcher gates a merge on **name AND category**, but the LLM labels the *same* entity with **synonym categories** across docs (`Country` in one, `Nation` in another), and that drift defeats **both** layers:

- **Blocking** — the structured key is `(kind, category, name-token)`, so a same-name pair whose category drifted never even blocks together (it could only reunite via a SimHash band on the whole-fingerprint embedding, which the by-design-divergent *defining attribute* pulls apart).
- **Scoring** — the category gate's "synonym escape hatch" was wired to the *whole-fingerprint* cosine (also attribute-polluted), so it never fires for a real bridge; the soft base then scores category at its low **lexical** value and the exact-name pair lands *marginally* below threshold (measured `0.7168` vs `0.72`).

## What (all in pyo3-free `goldenprofile-core` → every binding inherits it; the bench builds the native wheel **from source**, so no republish)

1. **Blocking** (`signature.rs`): a category-agnostic **exact-name key** (hashed on the whole normalized name, not per token) so identical proper names always block together regardless of category drift. Tight — groups only genuine same-name mentions, not everything sharing a token.
2. **Scoring** (`score.rs`): the gate's escape hatch now uses an optional **category-specific embedding** (synonyms ≈ close, distinct senses ≈ far) instead of the attribute-polluted fingerprint cosine; and the soft category term is lifted to `max(lexical, category-embedding)` so a category the embedding confirms is the same isn't under-counted. The gate's **lexical arm still uses the raw lexical score**, so gate precision is unchanged.
3. **Plumbing**: `ResolveRequest.category_embeddings` (optional, serde-default; empty → legacy fallback, byte-identical for existing callers), length-validated, threaded `resolve` → `score_pair`.
4. **Host** (`goldengraph/profile.py`): embed the category field separately and pass it (cheap, cache-friendly — few distinct category strings).

## Precision guard holds (unit-tested both directions)
- `Apple | Company` vs `Apple | Fruit` (distant category embeddings) → **stays gated out**, score 0.
- `Australia | Country` vs `Australia | Nation` (close category embeddings) → **now reunites at ~0.85** (was gated out).

5 new Rust tests (blocking, scoring gate both directions, end-to-end resolve) + 2 wheel-free host marshaling tests. `goldenprofile-core` (29) and goldengraph profile/cross-doc (41) suites green; `cargo clippy --workspace -- -D warnings` clean.

## Status — capability to MEASURE, not a declared win
A traced N=50 MuSiQue run is dispatched on this branch. The bar: the `RETRIEVAL-BROKEN-CHAIN` count drops without an `exact_match`/precision regression. I'll post the before/after on this PR before marking it ready.

---

### DIFF
--- a/packages/python/goldengraph/goldengraph/profile.py
+++ b/packages/python/goldengraph/goldengraph/profile.py
@@ -210,9 +210,13 @@ def resolve_profiles(
 
     When `embedder` is given, each fingerprint string is embedded and passed as
     the semantic signature (enabling SimHash-band blocking + the embedding-cosine
-    gate that bridges synonym categories). `config` is an optional partial
-    override of the engine's `ResolveConfig` (e.g.
-    `{"scoring": {"merge_threshold": 0.75}}`).
+    gate that bridges synonym categories). The CATEGORY field is ALSO embedded
+    separately and passed as `category_embeddings`: the category gate's synonym
+    escape hatch needs a category-specific signal -- the whole-fingerprint cosine
+    is polluted by the (by-design divergent) defining attribute, so an exact-name
+    bridge whose category label drifted ("Country" vs "Nation") would otherwise
+    never bridge. `config` is an optional partial override of the engine's
+    `ResolveConfig` (e.g. `{"scoring": {"merge_threshold": 0.75}}`).
     """
     resolve_json = _engine()
     profiles = []
@@ -233,6 +237,10 @@ def resolve_profiles(
     if embedder is not None and fingerprints:
         vecs = embedder.embed([fp.text for fp in fingerprints])
         request["embeddings"] = [[float(x) for x in row] for row in vecs]
+        # Category-only embeddings for the gate's synonym escape hatch. Embedding
+        # the (few distinct) category strings is cheap and cache-friendly.
+        cat_vecs = embedder.embed([p["category"] for p in profiles])
+        request["category_embeddings"] = [[float(x) for x in row] for row in cat_vecs]
     if config:
         request["config"] = config
 
--- a/packages/python/goldengraph/tests/test_profile.py
+++ b/packages/python/goldengraph/tests/test_profile.py
@@ -115,3 +115,61 @@ def test_resolve_profiles_reunites_disjoint_mentions():
     nabbes = next(c for c in res.clusters if 0 in c)
     assert sorted(nabbes) == [0, 1]
     assert any(c == [2] for c in res.clusters)
+
+
+class _CatStubEmbedder:
+    """Returns a fixed-dim vector per text; unknown -> zeros. Lets us assert the
+    request shape without the wheel."""
+
+    def embed(self, texts):
+        import numpy as np
+
+        return np.ones((len(texts), 3), dtype=float)
+
+
+def test_resolve_profiles_marshals_category_embeddings(monkeypatch):
+    """With an embedder, the request carries a category-only embedding per profile
+    (the gate's synonym escape-hatch signal) -- wheel-free: we capture the JSON the
+    engine boundary receives instead of running the engine."""
+    import json
+
+    from goldengraph import profile as prof
+
+    captured: dict = {}
+
+    def _fake_resolve_json(req_str: str) -> str:
+        captured["req"] = json.loads(req_str)
+        return json.dumps({"clusters": [[0], [1]], "edges": []})
+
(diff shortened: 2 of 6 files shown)


benzsevern:
## Measured on the bench — NOT a win (yet)

Traced N=50 MuSiQue, same config as the #1244 baseline (cross_doc + profile + trace), built from this branch's source:

| metric | baseline (#1244) | this branch |
|---|---|---|
| `answer_match` | 0.18 | **0.18** (flat) |
| `answer_match` (entity-subset) | 0.30 | 0.267 |
| RETRIEVAL-BROKEN-CHAIN | 8/50 | 6/50 |
| shatter verdicts | SCORING-miss:7 | SCORING-miss:6 |

**Honest read:** the headline did not move. The 8→6 broken-chain delta is within run-to-run LLM-build non-determinism, and the same `Exeter College` / `Oxford` shatter (`seed='Oxford' island='Oxford'`, cosine 1.000) persists **identically** in both runs — the exact case the fix targets is still not merging.

**Leading hypothesis (concrete + tunable):** `category_embedding_gate = 0.85` is on the `cosine01` scale (= raw cosine ≥ 0.70). Real category-*word* embeddings for synonyms (`College`≈`University`) likely sit at raw cosine ~0.5 → `cosine01` ~0.75, **below 0.85**, so the new escape hatch still never opens. Won't guess-tune it (lowering risks the `Apple`/`Fruit` over-merge the gate protects) — dispatching a `distill_log` run to capture the actual LLM fingerprints + categories for the persistent cases, so the threshold is set from data, not a hunch.

Keeping this **draft** until the diagnostic says whether it's (a) a gate-threshold miscalibration (tunable), or (b) genuinely-different senses (fix is correct, the probe is over-counting a false shatter). The code is sound and CI-green regardless; the question is whether it earns its place by actually moving the metric.

---

benzsevern:
## Final verdict — three runs, same config: the fix does NOT move the metric

| run | `answer_match` | entity-subset | broken-chain |
|---|---|---|---|
| baseline (#1244) | 0.18 | 0.30 | 8 |
| with-fix #1 | 0.18 | 0.267 | 6 |
| with-fix #2 (distill) | **0.16** | 0.233 | 8 |

`answer_match` is flat-to-slightly-down (0.18/0.18/0.16); broken-chain bounces 8/6/8 — the earlier 8→6 was **run-to-run LLM-build noise**, not the fix. The entity-subset trend (0.30→0.267→0.233) is a mild *downward* wobble — within noise on single runs, but not the direction a win looks like.

**Why cross-doc merge isn't the lever here (data-backed):** the gold answer-type mix is `{phrase:9, entity:30, number:7, date:4}` — **20/50 (40%) of golds are non-entity** (dates/amounts/phrases) an entity-graph engine can't emit by construction. Of the *entity* losses, the trace puts EXTRACTION (~22) + SYNTHESIS (~15) at ~75% and shatter at ~16%. So even a perfect cross-doc merge caps out well short of moving the headline, and the noise floor swamps it.

**The engine fix itself is correct** — unit-proven that `Australia|Country`↔`Australia|Nation` now reunites (~0.85) while `Apple|Company`↔`Apple|Fruit` stays apart (0.0), and it measurably increases graph connectivity (component count drops). It fixes a real synonym-category under-merge class that matters for graph quality *beyond* MuSiQue answer_match. But it does **not** earn a "moves the bench" claim, and I won't make one.

**Recommendation:** do not merge this as a MuSiQue accuracy win. Either (a) land it as a pure goldenprofile correctness improvement (low-risk, CI-green, fixes a genuine bug) if graph quality for the broader KG North Star is worth it, or (b) shelve it. Either way, the real accuracy levers for this bench are **extraction** (the 40% non-entity golds → literal/attribute path #1236; plus entity-extraction misses) and **synthesis** (#1240 LLM-judge + a better synthesis prompt) — not cross-doc merge. Leaving this **draft** for your call.

---

PR #1252, opened 2026-06-23
feat(goldengraph): literal-attribute extraction → reachable value nodes

benzsevern:
## Why

The measurement loop localized goldengraph's **dominant** loss as EXTRACTION (~22/50 MuSiQue questions), and the gold answer-type mix is `{entity:30, phrase:9, number:7, date:4}` — **40% of golds are non-entity** (dates/quantities/phrases) the entity→entity schema can't represent, so they never enter the graph and are unanswerable by construction. The cross-doc shatter (#1249) was only ~16% of losses and didn't move the headline; this targets the biggest bucket.

## What (opt-in `GOLDENGRAPH_LITERAL_ATTRS=1`; byte-identical default-off)

- **`extract.py`**: splice an `attributes` schema into the extraction prompt so the LLM also emits literal facts `(entity, key, value)` — a date, quantity, or short defining value that answers "when/how many/how much/what kind" but isn't a named entity. New `Attribute` dataclass + `Extraction.attributes` (default empty → back-compat). Flag read *inside* `extract()` so monkeypatched 2-arg stubs keep working (#1236 lesson). Defensive parse drops empty values + out-of-range owners.
- **`ingest.py::build_batch`**: materialize each attribute as a literal **node** (`canonical_name=value`, `typ="literal"`) + an edge `owner -[key]-> value`. The synthesis prompt already permits answering with *any* node in the Entities list, so a materialized literal is an eligible, reachable answer. Dedups repeated `(owner,key,value)` within a doc.
- **Precision guard**: literal nodes get **unique per-occurrence record keys** (`lit:<at>:<id>`) **and** are excluded from cross-doc link candidates (`_new_features`/`_existing_features` skip `typ="literal"`), so two identical values in different docs never overlap-merge into a false bridge — they stay leaves owned by their own entity.
- **workflow**: a `literal_attrs` input → `GOLDENGRAPH_LITERAL_ATTRS` env.

7 offline tests (parse + back-compat, prompt-splice gating, node+edge materialization, dedup, unresolved-owner drop, link-candidate exclusion). Full goldengraph offline suite green.

## Status — capability to MEASURE, not a declared win
A traced N=50 MuSiQue run with the flag is dispatched. The bar: the EXTRACTION stage count drops and `answer_match` (especially on the date/number golds) rises, **without** a precision regression. I'll post the before/after here before marking ready — same honest loop as #1249, where I refused to ship a flat result.

---

### DIFF
--- a/.github/workflows/bench-graphrag-qa.yml
+++ b/.github/workflows/bench-graphrag-qa.yml
@@ -57,6 +57,9 @@ on:
       distill_log:
         description: "goldengraph only: capture (text->extraction)+(entity->fingerprint) pairs to a JSONL artifact for in-house distillation"
         default: "false"
+      literal_attrs:
+        description: "goldengraph only: extract literal attributes (dates/quantities/short values) as reachable value nodes -- targets non-entity answers"
+        default: "false"
 
 permissions:
   contents: read
@@ -124,6 +127,7 @@ jobs:
           GOLDENGRAPH_BUILD_WORKERS: ${{ inputs.build_workers }}
           GOLDENGRAPH_BUILD_DEBUG: ${{ inputs.build_debug }}
           GOLDENGRAPH_EXTRACTOR: ${{ inputs.extractor }}
+          GOLDENGRAPH_LITERAL_ATTRS: ${{ inputs.literal_attrs }}
           GOLDENGRAPH_DISTILL_LOG: ${{ inputs.distill_log == 'true' && format('results/distill_goldengraph_amb{0}.jsonl', matrix.ambiguity) || '' }}
         run: |
           mkdir -p results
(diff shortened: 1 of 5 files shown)


benzsevern:
Auto-merge **disabled / de-queued** to avoid a guaranteed merge-queue collision with #1236 — both PRs create `tests/test_literal_attributes.py` and edit the same `extract.py`/`ingest.py` hunks, so whichever landed second would fail on an add/add conflict.

#1236 is the more complete implementation of the same feature (it also updates `synthesize.py` to quote literals and permit literal answers, which this PR does not). The one thing this PR has that #1236 lacks is the **`seed_by_query` 400-fix** — without it #1236's literal path 400s on every answer (flagged on #1236). 

Plan: let #1236 land, then I port just the seed-fix as a small follow-up. Holding this PR open as the reference for that fix + the measurement (literal nodes without the synthesis change scored 0.14 vs 0.18 baseline — the synthesis change is what makes them answerable). Will close once the follow-up lands.

---

benzsevern:
Closing — superseded by **#1236** (which merged, with the more complete synthesis change) + **#1253** (the seed/link embed-400 fix this PR carried, ported onto the merged #1236). Nothing here is lost: #1253 contains the unique fix.

---

PR #1354, opened 2026-07-02
fix(autoconfig): compute cardinality on full data, not the 1k sample (#1351)

### LINKED ISSUE #1351: AutoConfigController commits standalone exact matchkey on low-cardinality fields (e.g. zip) → over-merge on high-density data
## Summary
`AutoConfigController` (`auto_configure_df` / the zero-config `dedupe_df(df)` path) will commit a **standalone `exact` matchkey on a low-cardinality field** (observed: `zip` alone, `phone` alone). On datasets where many distinct entities share a value for that field (e.g. lots of people per zip), this collapses huge false clusters. It's fine on data where the field is sparse, so it only bites on high-per-value-density inputs — but when it bites, it's catastrophic (>50% of rows merged).

This is **not** "auto-config over-merges in general" — on clean/representative data it's conservative and correct (evidence below). The bug is specifically the standalone-exact-on-low-cardinality-field choice.

## Evidence (same zero-config path, three datasets)

| dataset | rows | removed | biggest cluster | committed keys (excerpt) |
|---|---|---|---|---|
| nc_voters_5k | 5,000 | 71 (1.4%) | 3 | `exact[full_phone_number]`, `exact[last_name,first_name,birth_year,registr_dt]`, `weighted[...]` |
| messy_customers | 49 | 1 (2.0%) | 2 | `weighted[first_name,last_name,address]` |
| **circ contact list (19,278 rows, ~74 cols)** | 19,278 | **10,553 (54.7%)** | **52** | **`exact[zip]`**, **`exact[phone]`**, `exact[last_name,company,q_date,update_time]`, `weighted[...]` |

On nc_voters the controller chose a **high-cardinality** exact key (`full_phone_number`) or **multi-field composites**. On the circ list it chose **`exact` on `zip` by itself** (and `phone` by itself). The circ data has many subscribers per zip, so `exact[zip]` merges everyone in a zip → biggest false cluster is 52 distinct people.

The problematic committed config (circ list):
```
matchkey: exact  fields=[zip]                      # <-- standalone low-cardinality exact
matchkey: exact  fields=[phone]                    # <-- standalone
matchkey: exact  fields=[last_name,company,q_date,update_time]   # timestamps in a key
matchkey: weighted thr=0.8 fields=[address1, company, last_name, first_name, create_time]
```
(Data is a proprietary PII circulation list, so not attached — shape: 74 cols, `email` ~53% null, `birth_date`/`hce_id` ~100% null, no clean unique person key, high per-zip/per-phone density.)

## Expected
The controller should not commit a **standalone `exact` matchkey on a field whose value-density is high** (i.e. average rows-per-distinct-value well above ~1, or cardinality far below row count). Options:
- Gate standalone `exact` keys on a max rows-per-value / min-cardinality threshold; require a composite or a weighted key when a candidate field is low-cardinality.
- Or estimate the block-size blow-up of a candidate exact key and reject keys that would create oversized clusters.

## Secondary finding (probably a separate, smaller issue)
The postflight advisory **"score distribution is unimodal; threshold cannot be auto-set"** fires on **all three** datasets above — including the two that dedupe correctly. So it's over-firing / not diagnostic of a real problem. Worth tightening so it doesn't cry wolf on healthy runs.

## Repro
Run the zero-config path (`goldenmatch.dedupe_df(df)` with no config, i.e. AutoConfigController) on any dataset with a low-cardinality, high-density field (e.g. a contact list where a `zip`/`region` column has hundreds of rows per value) and inspect the committed `result.config.matchkeys` — a standalone `exact[low_card_field]` will be present and the corresponding clusters will be oversized.

benzsevern:
Closes #1351.

## Problem
`profile_columns` measured `cardinality_ratio` (and `null_rate`) on a **1,000-row random sample**. A small sample systematically **over-estimates** cardinality for moderate-cardinality columns (birthday / coupon-collector effect): a `zip` column with ~5k distinct values in ~19k rows reads ~0.96 in a 1k sample vs a true ~0.26. That inflated ratio promoted `zip → "identifier"` (the numeric-shape cardinality promotion), which defeated the zip/geo exact-matchkey guard and committed a standalone `exact[zip]` matchkey → **~55% over-merge** on real circulation data (dozens of distinct people merged on a shared zip). Type CLASSIFICATION tolerates sampling; CARDINALITY does not.

This is the root-cause fix. A discarded surgical version only patched the exact-matchkey gate; per review this was moved to the source so cardinality is honest for **every** consumer (classification, blocking, weighting), not just one gate.

## Fix
Decouple the two: classification value-heuristics keep running on the sample; `cardinality_ratio` + `null_rate` are now measured on the **full column**.
- `_full_column_cardinality_null`: exact `n_unique` below `_FULL_CARDINALITY_MAX_ROWS` (default 2,000,000, env-override `GOLDENMATCH_FULL_CARDINALITY_MAX_ROWS`), Polars HyperLogLog `approx_n_unique` at/above — always on the full column, never the 1k sample. Same ratio definition as before (unique non-null-non-blank / total rows incl nulls). Wrapped in try/except → falls back to the sample estimate on exotic dtypes (Object/Binary/List/Struct), never raises. Blank-strip only applies to string dtypes (numeric/date hash the native dtype directly).
- `_classify_by_data` takes an honest `cardinality_ratio` kwarg, used in the identifier-promotion floor instead of the sample recompute (absent → legacy sample behavior, so hand-built callers/tests are unchanged).
- **Native path:** the Rust classifier recomputes cardinality from `sample_values` and ignores the passed ratio (can't change Rust here). So on the native path we re-check each native `"identifier"` and demote promotion-eligible shapes (phone/zip/numeric) whose honest full-data ratio is below the floor — **respecting name-authority** (a `*_id`/`*_key`/`uuid` column stays `identifier`, mirroring the pure-Python combine, so a numeric `order_id` isn't wrongly demoted to `numeric` and stripped of its matchkey). Python-side only; a no-op when native already agreed.

`build_matchkeys`, the `exact_matchkey_floor` oracle, the zip/geo guard, and the composite paths are **untouched** — with honest cardinality flowing in, the existing floor + guard correctly reject `exact[zip]` unchanged. Phone is not special-cased (its honest ~0.50 still clears the 0.30 phone floor, so real behavior is preserved).

## Tests
New `tests/test_profile_full_cardinality_1351.py` (6 tests): honest low ratio on a >1k-row high-density column; moderate-cardinality zip stays `zip` (the bug repro); genuinely near-unique still `identifier`; honest `null_rate`; native/python promotion agreement; and name-authoritative `order_id` not demoted (the review-found regression guard). Exercises whichever classify path is active (native was on in the dev env).

`test_profile_full_cardinality_1351.py` + `test_autoconfig.py` + `test_exact_matchkey_floor_s3.py` + `test_autoconfig_pincer_715.py` → **136 passed, 3 skipped**. No existing test modified. `ruff check` clean.

## Validation / known limitation
CI's autoconfig **accuracy gates** (Febrl / DBLP-ACM / NCVR / #528) are the regression validation for the classification-shift blast radius — they can't be run locally (they OOM). **Known limitation (follow-up):** the promotion floor `max(0.95, 1−1/√n)` still uses the sample length, so for datasets ≫1k rows a near-unique column (true ratio ~0.97–0.999) can still promote; this is a pre-existing, low-harm case (near-unique columns are cheap/safe as exact keys) and not a regression — the catastrophic low-cardinality case is fixed. Worth a separate tightening (thread true row count into the floor).

### DIFF
--- a/packages/python/goldenmatch/goldenmatch/core/autoconfig.py
+++ b/packages/python/goldenmatch/goldenmatch/core/autoconfig.py
@@ -185,13 +185,19 @@ def _full_column_cardinality_null(
         if total_rows == 0:
             return 0.0, 0.0
         null_rate = series.null_count() / total_rows
-        # Strip blanks consistently with the sample values filter (str(v).strip()).
-        non_null = series.drop_nulls().cast(pl.Utf8).str.strip_chars()
-        non_blank = non_null.filter(non_null.str.len_chars() > 0)
+        non_null = series.drop_nulls()
+        # Blank-stripping only applies to text: non-string dtypes have no "blank"
+        # once nulls are dropped, so skip the Utf8 cast + per-value scan (pure
+        # overhead on numeric/date/bool) and hash the native dtype directly. For
+        # strings, strip + drop empties to match the sample values filter
+        # (str(v).strip() truthiness).
+        if non_null.dtype in (pl.Utf8, pl.String, pl.Categorical):
+            stripped = non_null.cast(pl.Utf8).str.strip_chars()
+            non_null = stripped.filter(stripped.str.len_chars() > 0)
         if total_rows >= _full_cardinality_max_rows():
-            unique = non_blank.approx_n_unique()
+            unique = non_null.approx_n_unique()
         else:
-            unique = non_blank.n_unique()
+            unique = non_null.n_unique()
         return unique / total_rows, null_rate
     except Exception:  # noqa: BLE001 -- exotic dtype; fall back to sample estimate
         return None
@@ -490,6 +496,17 @@ def profile_columns(
         }
         for prof in profiles:
             if prof.col_type != "identifier":
+                continue
+            # Name-authority guard: a column named like an id (`*_id`, `*_key`,
+            # `uuid`, `account_no`, ...) is "identifier" because its NAME says so,
+            # and the pure-Python path treats that as authoritative over data
+            # shape (identifier in _name_authoritative). Such a column can hold
+            # legitimately-repeating numeric ids (moderate cardinality), so only
+            # demote when the classification is NOT name-authoritative -- i.e.
+            # this "identifier" came from the cardinality promotion we're
+            # correcting, not from the column name. Otherwise a numeric-shaped
+            # `order_id` would be demoted to "numeric" and lose its matchkey.
+            if _classify_by_name(prof.name) == "identifier":
                 continue
             vals = names_to_sample_values.get(prof.name, [])
             honest = honest_ratios.get(prof.name)
--- a/packages/python/goldenmatch/goldenmatch/core/autoconfig.py
+++ b/packages/python/goldenmatch/goldenmatch/core/autoconfig.py
@@ -133,6 +133,70 @@ _ID_PATTERNS = re.compile(
 )
 
 
+# Full-data cardinality threshold (#1351). cardinality_ratio + null_rate are
+# measured on the FULL column, not the 1k classification sample, because a small
+# sample systematically OVER-estimates cardinality_ratio (birthday /
+# coupon-collector effect) -- which used to promote moderate-cardinality
+# zip/phone columns to "identifier" and commit a standalone exact matchkey (~55%
+# over-merge). Below this row count use an EXACT n_unique on the full column;
+# at/above it fall back to Polars' HyperLogLog approx_n_unique (still on the FULL
+# column, never the sample). Env override follows the GOLDENMATCH_* knob
+# convention.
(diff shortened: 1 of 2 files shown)


benzsevern:
Superseded. The full-data-cardinality approach failed the auto-config accuracy gate — cardinality can't separate a shared identity key (npi) from a shared locality attribute (zip); both are moderate-cardinality, so honest cardinality demoted npi/phone_number (F1 -6 to -13pts, blocking blowup). Replacing with a discriminative-power veto (co-agreement of shared-value records on other identity fields) on branch fix/1351-discriminative-veto. Design + plan in docs/superpowers/. Keeping #1351 open.

PR #1488, opened 2026-07-06
feat(goldenflow): fused columnar apply — owned-kernel chain in one pass (Pillar-1)

benzsevern:
**First real swing at Pillar-1 of the Rust cutover: evict Polars from the transform *execution* path.** Kernels were already owned (byte-parity across surfaces); execution still rode Polars *per transform*. This fuses a maximal run of owned, string→string, no-arg kernels on a column into **one** native Arrow round-trip instead of N.

> **Stacked on #1480** (needs its `*_into` streaming kernels + the `arrow` feature). Base is `feat/goldenflow-columnar-pilot` for a clean diff; retarget to `main` once #1480 merges. Do not enqueue until then.

## The cost it removes
The engine applied N transforms to a column as N × (Series→Arrow export + kernel + Arrow→Series import + `with_columns` + a full-column affected-count scan). Fused = one export/import, one column rebuild.

## What ships
- **`goldenflow-core/chain.rs`** (feature `arrow`): `enum Kernel` (14 no-arg total text kernels) + `apply_chain` threading each row through the kernels via two reused ping-pong buffers, returning the transformed array + per-kernel affected-row counts. **Generic over offset width** (`GenericStringArray<O>`) — critical, because **Polars exports `LargeUtf8`**, so an i32-only path would never fire on real data. Parity by construction: each `Kernel` dispatches to the same owned kernel, so `apply_chain([k1,k2]) == k2(k1(x))`.
- **native-flow `apply_chain_arrow`** — one round-trip, Utf8 + LargeUtf8 arms; `fusable_kernel_names()` for the coverage guard.
- **engine `_apply_column_ops`** — detects maximal fusable runs behind `GOLDENFLOW_FUSED_APPLY` (**default off**; string columns only; native required). `_apply_fused_run` emits one audit record per op with the **exact** kernel affected count + per-step samples from a cheap head(3) replay, so the manifest is **byte-identical** to the per-transform path (option B — preserves the audit trail). Everything non-fusable and the whole path when off stay per-transform, unchanged.

## Guards + parity (`test_fused_chain.py`)
- `FUSABLE_KERNELS` are all registered single-column (expr/series) transforms.
- `FUSABLE_KERNELS` mirrors the native `Kernel` table (coverage guard — a new owned kernel can't silently skip the fast path).
- Fused output frame **and** manifest are byte-identical to the per-transform path over several chains (native lane).

## Benchmark (`benches/chain_apply.rs`)
4-op cleanup chain (`strip→lowercase→collapse_whitespace→remove_punctuation`) @ 1M rows: the **in-Rust component alone is 1.36×**; the dominant **4× host-boundary saving is on top** (not visible in a Rust bench).

## Scope
Phase 1 = text family, native-flow surface, flag-gated. Follow-ups: parameterized/numeric chains, wasm/duckdb surfaces, default-on after an at-scale host bench.

## Note surfaced while building
**#1480's columnar fast path (`map_str_columnar`/`ascii_case`) only handles `StringArray` (Utf8), but Polars exports `LargeUtf8`** — so on real Polars data it silently takes the scalar fallback (still correct, just not the measured 4-10×). This chain handles both; worth making #1480's helpers generic too as a follow-up.

## Verification
Core 6 chain tests + clippy(`--features arrow`); native-flow check+clippy+fmt; Python guard + 47 integration tests (flag off = transparent). Native e2e rides the CI native lane (wheel LLVM-OOMs locally).

https://claude.ai/code/session_01FYFLbXeAk9UBHv7T3hFBtg

### DIFF
--- a/.github/workflows/bench-goldenflow-fused.yml
+++ b/.github/workflows/bench-goldenflow-fused.yml
@@ -1,0 +1,61 @@
+name: Bench goldenflow fused-apply (A/B)
+
+# A/B harness for the fused columnar apply path (Pillar-1). Builds native-flow
+# into the package + runs the end-to-end bench at scale, printing per-transform
+# vs fused (wall + peak RSS + parity) to the step summary. This is the
+# measurement that gates the GOLDENFLOW_FUSED_APPLY default-on decision -- the
+# in-Rust component bench (benches/chain_apply.rs) only shows part of the win; the
+# per-transform boundary crossings that fusion removes are visible only here.
+# workflow_dispatch only; never runs on push/PR. (Dispatchable once this workflow
+# is on the default branch -- i.e. after the fused-apply PR merges.)
+on:
+  workflow_dispatch:
+    inputs:
+      ref:
+        description: "Branch/commit to build + bench (default: the dispatch ref)"
+        required: false
+        default: ""
+      rows:
+        description: "Comma-separated row counts to bench"
+        required: false
+        default: "1000000,5000000"
+
+permissions:
+  contents: read
+
+jobs:
+  bench:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10  # v6
+        with:
+          ref: ${{ inputs.ref || github.ref }}
+      - uses: dtolnay/rust-toolchain@29eef336d9b2848a0b548edc03f92a220660cdb8  # stable
+      - uses: Swatinem/rust-cache@e18b497796c12c097a38f9edb9d0641fb99eee32  # v2
+        with:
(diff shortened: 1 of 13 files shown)


benzsevern:
Folded the LargeUtf8 fix into this PR (commit `62e795eb`): #1480's columnar helpers (`map_str_columnar` / `ascii_case` / `scalar_map`) are now generic over offset width, and native-flow's util shims downcast both `StringArray` and `LargeStringArray`. So the columnar fast path now fires on real Polars data (LargeUtf8), not just the StringArray bench — the measured 4-10x is real in production. Parity test `fires_on_large_utf8_the_polars_shape` pins Utf8≡LargeUtf8.

PR #1625, opened 2026-07-10
fix(ci): force pyarrow system memory pool in the rust lane (bridge SIGSEGV)

benzsevern:
## What and why

The `rust` CI lane's `cargo test --workspace` (goldenmatch-bridge is the sole workspace member) was crashing with `signal: 11 SIGSEGV`, which bounced PRs out of the merge queue (e.g. #1504, and several others surfacing on the same lane). This forces pyarrow onto the system memory pool, which eliminates the crash.

Root-caused via gdb: the segfault is `mi_thread_init` in pyarrow's bundled **mimalloc** allocator, on a **worker thread** (`start_thread` -> `clone3`) calling `pyarrow.array()` (`ConvertPySequence`). goldenmatch's pipeline (the `test_autoconfig_*` / `test_dedupe_*` controller tests) spawns polars / ThreadPoolExecutor workers that allocate through pyarrow's mimalloc, whose per-thread heap init faults under the bridge's embedded CPython.

It is **not** a test-thread race (reproduces at `--test-threads=1`) and **not** goldenmatch logic (each test passes in isolation) -- it is the accumulated thread churn across the 46-test process that trips mimalloc's per-thread heap init.

## Changes

- `.github/workflows/ci.yml`: add `ARROW_DEFAULT_MEMORY_POOL: system` to the `rust` lane's `cargo test --workspace` step env (the documented Arrow remedy). A `std::env::set_var` in `bridge::init()` would be unsound (mutating env with live threads is exactly the hazard here), so the fix is the process env, not code.
- `packages/rust/extensions/CLAUDE.md`: document the root cause + the local-dev env var (`cargo test -p goldenmatch-bridge` needs it too).

## Testing

Reproduced faithfully on a clean py3.12 venv + `pip install goldenmatch pyarrow` (matches the CI setup), building the bridge with `PYO3_PYTHON` pinned to it:

- Without the fix: **20/20 runs SIGSEGV** (both default-parallel and `--test-threads=1`).
- With `ARROW_DEFAULT_MEMORY_POOL=system`: **20/20 runs clean, all 46 tests pass**.

gdb backtrace confirmed `mi_thread_init` (pyarrow libarrow mimalloc) on a worker thread as frame #0, with the crash stack matching the CI log (`test_dedupe_basic ... ok`, `test_dedupe_clusters ... ok`, then SIGSEGV).

## Checklist

- [x] Tests added/updated and passing locally for the changed package (CI-config + docs only; verified via the repro above)
- [ ] `pre-commit run --all-files` is clean (ruff, whitespace, no secrets)
- [x] Package `CHANGELOG.md` updated if behavior changed (n/a -- CI/test-infra fix, no shipped behavior change)

https://claude.ai/code/session_01XeAWuM4pL7hjs6t8F6PsKN

---
_Generated by [Claude Code](https://claude.ai/code/session_01XeAWuM4pL7hjs6t8F6PsKN)_


benzsevern:
Marking ready + arming auto-merge from the queue-unblock session: this fix's branch ran the rust lane green (15:28 run), the queue is kicking every rust-triggering entry on the SIGSEGV, and the change is a CI-env one-liner. If more work was intended here, follow up on main.

benzsevern:
Superseded by #1627, which cherry-picked this commit (authorship preserved): the two fixes blocked each other's CI (this PR's force_all queue group hit the native_flow clippy red that #1627 fixes; #1627's rust lane hit the SIGSEGV this fixes). Combining them breaks the deadlock.

PR #1691, opened 2026-07-12
[SPIKE] frame kernels in analysis-wasm via a minimal C-Data column ABI (Wave 1b unblock)

benzsevern:
## ⚠️ SPIKE — findings, not for merge as-is

Answers one question: **can the Wave-1b-deferred frame kernels run in `analysis-wasm` without the two blockers that deferred them?** (The deferral note cited: *arrow-in-wasm is heavy*, and *a second interning impl is a new drift surface*.)

## The approach

The kernels (`duplicate_row_ratio` / `distinct_count`) already take **pre-interned `u64` columns** — arrow-free. The *only* arrow-coupled step was **interning** (`analysis-native::intern_column` over `arrow::ArrayData`). So:

1. Move interning into `analysis-core` over **plain typed buffers** (values + byte-per-row validity mask; utf8 = offsets + bytes) — i.e. the **Arrow C Data Interface memory layout, minus the arrow-rs library**.
2. `analysis-wasm` gains a `FrameInterner` **column-handle** binding: JS writes each column's typed arrays (`Float64Array` / `BigInt64Array` / `Uint32Array`+`Uint8Array`) straight into wasm linear memory; the **shared** core interner produces dense `u64` ids; the existing core kernel runs. Mirrors building an Arrow RecordBatch column-by-column.

This is the same pattern GoldenCheck's SQL surface already uses (*"the NULL→0 first-seen interning the kernels expect is done in Rust… value-for-value with the wheel's Arrow interning"*) — the spike just extends it to WASM and unifies the interning into the shared core.

## Findings (measured locally, wasm32 release)

| | Result |
|---|---|
| **arrow-rs linked?** | **No** — 0 array/RecordBatch/ipc symbols in the `.wasm` (the lone "arrow" string is a doc comment). **Blocker #1 avoided.** |
| **Binary size** | 52,638 → 81,054 bytes (**+28 KB, +54%**) for the *full* frame-kernel + interning path — vs the **megabytes** arrow-in-wasm would add. Bounded and small. |
| **Parity** | Interning is now **one shared impl**, proven byte-identical to native's Arrow `intern_column` via **4 adversarial-fixture tests** (float NaN/−0.0/null, mixed dtypes, empty-string-vs-null, typed numeric) asserting the same `dup_ratio`/`distinct` values committed in `frame_kernels_adversarial.json`. **Blocker #2 (second-impl drift) dissolved** — it's not a second impl. |

Net: both stated reasons for the Wave-1b deferral **do not hold** under the C-Data ABI approach. The "bare metal = agree on a memory layout, don't ship the library" thesis checks out on the exact case it was supposed to be too hard for.

## What this spike did NOT do

A real Wave-1b-unblock PR would still need:
- **Wire `analysis-native` to also use the shared `intern_*`** (true unification — right now native keeps its Arrow path; the spike only proves the shared intern *matches* it).
- **The TS surface**: `build_analysis_wasm.mjs` + the `AnalysisBackend`/loader wiring + JS-side buffer writers + a TS parity test.
- **Wall-clock boundary cost** — the one dimension I did *not* measure: the JS→wasm typed-array copy vs. the dedup work. My reasoning: the copy is O(bytes) memcpy, dominated by the O(n) interning + O(n) dedup hashing at any frame size where dedup matters; below that, pure-TS wins anyway (which is *why* the deferral said "browser frame-dedup isn't an established workload"). Worth confirming with a real bench before committing.
- **Packed validity bitmap** instead of byte-per-row (the production C-Data refinement; byte-per-row kept the spike ABI trivial from JS).

## Testing

`cargo test -p goldenmatch-analysis-core` → **21 passed** (incl. the 4 new intern parity tests). Size + no-arrow verified on the wasm32 release build.

https://claude.ai/code/session_01LDkMRQ2u3pmPdnKosm1x3x

---
_Generated by [Claude Code](https://claude.ai/code/session_01LDkMRQ2u3pmPdnKosm1x3x)_

### DIFF
--- a/packages/rust/extensions/analysis-core/src/lib.rs
+++ b/packages/rust/extensions/analysis-core/src/lib.rs
@@ -147,9 +147,171 @@ pub fn cluster_size_histogram(sizes: &[f64]) -> Vec<i64> {
     buckets.to_vec()
 }
 
+// ---------------------------------------------------------------------------
+// Minimal C-Data column interning (arrow-free) -- SPIKE (2026-07-11)
+// ---------------------------------------------------------------------------
+//
+// The frame kernels (`duplicate_row_ratio`, `distinct_count`) already take
+// pre-interned dense `u64` columns, so they are arrow-free. Interning -- turning
+// column values into dense value-ids -- was the ONLY arrow-coupled step, living
+// in `analysis-native::intern_column` over `arrow::ArrayData`. That is exactly
+// why the wasm frame kernels were deferred (Wave 1b): bridging needed either
+// arrow-rs in the wasm (heavy) or a SECOND intern impl in TS (a drift surface).
+//
+// Moving interning HERE -- over plain typed buffers, not Arrow -- lets every
+// surface intern through the SAME code: native adapts its Arrow arrays to these
+// buffers, wasm/JS writes typed arrays straight into linear memory, SQL passes
+// flat arrays (goldencheck's SQL surface already interns raw buffers in Rust this
+// way). Semantics match `intern_column` EXACTLY: null -> id 0, non-null -> dense
+// ids from 1, floats canonicalized so `-0.0`/`+0.0` fold and every `NaN`
+// collapses to one id (Polars equality). Only the equality PARTITION matters to
+// the kernels, never the specific id numbers, so this is byte-parity by
+// construction, not by a re-derived TS mirror.
+//
+// `validity`: one byte per row, 0 = null, non-zero = valid. (A packed Arrow
+// validity bitmap is the production refinement; byte-per-row keeps the spike ABI
+// trivial from JS -- just a `Uint8Array`.)
+
+/// Canonical f64 bit pattern: one id for all `NaN`, one for `-0.0`/`+0.0`.
+/// Mirrors `analysis-native::canon_f64_bits`.
+pub fn canon_f64_bits(x: f64) -> u64 {
+    if x.is_nan() {
+        0x7ff8_0000_0000_0000
+    } else if x == 0.0 {
+        0.0f64.to_bits()
+    } else {
+        x.to_bits()
+    }
+}
+
+#[inline]
+fn is_null(validity: &[u8], i: usize) -> bool {
+    validity.get(i).map_or(false, |&b| b == 0)
+}
+
+/// Intern an f64 column (canonicalizing NaN / signed-zero) to dense u64 ids.
+pub fn intern_f64(values: &[f64], validity: &[u8]) -> Vec<u64> {
+    let mut map: HashMap<u64, u64> = HashMap::new();
+    let mut ids = Vec::with_capacity(values.len());
+    let mut next: u64 = 1;
+    for (i, &v) in values.iter().enumerate() {
+        if is_null(validity, i) {
+            ids.push(0);
+            continue;
+        }
+        let key = canon_f64_bits(v);
+        let id = *map.entry(key).or_insert_with(|| {
+            let n = next;
+            next += 1;
+            n
+        });
+        ids.push(id);
+    }
+    ids
+}
+
+/// Intern an i64 column to dense u64 ids.
+pub fn intern_i64(values: &[i64], validity: &[u8]) -> Vec<u64> {
+    let mut map: HashMap<i64, u64> = HashMap::new();
+    let mut ids = Vec::with_capacity(values.len());
+    let mut next: u64 = 1;
+    for (i, &v) in values.iter().enumerate() {
+        if is_null(validity, i) {
+            ids.push(0);
+            continue;
+        }
+        let id = *map.entry(v).or_insert_with(|| {
+            let n = next;
+            next += 1;
+            n
+        });
+        ids.push(id);
+    }
+    ids
+}
+
+/// Intern a UTF-8 column to dense u64 ids. `offsets` has `n_rows + 1` entries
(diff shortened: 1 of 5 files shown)


benzsevern:
## Boundary-cost bench (the previously-unmeasured dimension)

`bench/duplicate_row_ratio_bench.mjs` — same columnar source into both paths; wasm builds the C-Data buffers + interns + dedups, pure-JS builds string keys into a `Map`. Node, median of runs, results agree (`dup_ratio` identical across the boundary).

| n | columns | JS | WASM = build + call | verdict |
|---|---|---:|---:|---|
| 100K | str+i64+f64 | 182ms | 158 = 236b + 55c | 1.15x faster |
| 100K | **i64+f64** | 77ms | **33 = 0b + 33c** | **2.32x faster** |
| 500K | str+i64+f64 | 1095ms | 1728 = 4918b + 555c | 1.58x slower |
| 500K | **i64+f64** | 2046ms | **419 = 0b + 421c** | **4.89x faster** |
| 1M | str+i64+f64 | 4839ms | 16368 = 14839b + 1529c | 3.38x slower |
| 1M | **i64+f64** | 1597ms | **938 = 0b + 997c** | **1.70x faster** |

**Read:** the kernel+intern **`call` beats pure-JS in every row**. The verdict flips only on the **string `build`** — per-cell `TextEncoder.encode` + offset assembly on the JS side. For numeric columns build ≈ 0 (typed arrays passed straight in) and wasm wins **1.7–4.9x**.

**Caveat / next lever:** the string build numbers are a *pessimistic upper bound* — the harness allocates a `Uint8Array` per cell (GC storm; that's the non-monotonic 500K vs 1M). `TextEncoder.encodeInto` into one pre-sized buffer, or interning strings JS-side to ints and passing only int columns, would cut it hard. The string tax is the real design question, not the kernel.

---

benzsevern:
## Option B: JS-intern strings→ints, wasm dedups integers

The string-heavy frame that lost under naive byte-marshaling (1.6–3.4x *slower*) — retested by interning string columns to dense ints **on the JS side** (fast V8 `Map`) and passing only integer columns to wasm. `bench/duplicate_row_ratio_optionb_bench.mjs`.

| n | A: baseline JS | B: JS-intern + wasm | C: JS-intern + JS | **B vs A** |
|---|---:|---:|---:|---|
| 100K | 224ms | **102ms** (32 intern + 57 wasm) | 286ms | **2.19x faster** |
| 500K | 1346ms | **961ms** (399 + 696) | 1446ms | **1.40x faster** |
| 1M | 2470ms | **2270ms** (881 + 1282) | 3338ms | **1.09x faster** |

**Reads:**
- **String-heavy frames now win** (1.09–2.19x) — the flip from the naive-marshaling loss.
- **wasm earns its keep**: B < C at every size, so the integer row-dedup is genuinely faster in wasm than JS.
- **C > A at 1M**: pure-JS intern-then-dedup is *worse* than baseline — the JS-intern step only pays off *paired with wasm*.
- **The margin erodes toward 1M** (2.19x → 1.09x): the JS string-intern cost grows linearly and the two converge. Sweet spot is ≤500K — which is the realistic in-browser/edge frame size anyway ("browser frame-dedup isn't an established workload" cuts both ways).

**Conclusion — the shape is a hybrid, not "put the kernel in wasm":** JS interns string columns (V8's strength), wasm does the multi-column integer row-dedup (its strength, 1.5–3x over JS), numeric columns pass through free. The C-Data ABI is what makes the integer hand-off cheap.

---

benzsevern:
## Productionization verdict: NOT worth shipping on the current TS model

Retested Option B on the **real** goldenanalysis TS frame shape — row-oriented, heterogeneous `Array<Record<string, unknown>>`, not the idealized pre-columnar typed arrays the earlier benches used. `bench/duplicate_row_ratio_framerows_bench.mjs`.

| n | pure-TS | wasm-path (intern + wasm) | verdict |
|---|---:|---:|---|
| 100K | 404ms | 310ms (251 + 58) | 1.30x faster |
| 500K | 2286ms | 2150ms (1698 + 483) | 1.06x (noise) |
| 1M | 4811ms | 5365ms (3408 + **1314**) | **1.12x slower** |

**Why it flips:** the `intern` step (per-cell `JSON.stringify` + `Map`) dominates (251/1698/3408ms); the wasm dedup is a small slice (58/483/1314ms). Because `FrameRows` is row-oriented, both paths must serialize every cell — and *that* is the bottleneck, not the row-dedup. WASM can only accelerate the part that isn't the bottleneck.

**Conclusion:** the blocker is the **data representation, not the kernel**. The wasm frame kernels are feasible, tiny, parity-clean, and fast in isolation — but they only pay off with a **columnar typed-array TS frame path**, which goldenanalysis doesn't have (and which needs a real edge-columnar-ingestion use case to justify building). So Wave 1b stays deferred — now for a *precise, measured* reason: the row-oriented TS model, not "arrow-in-wasm is heavy."

Keeping this PR as a **findings artifact** (all four benches + the C-Data intern + parity tests are here for whenever a columnar TS frame path materializes). Not merging.

---

benzsevern:
Spike complete -- findings harvested into #1788 (the Wave-1b unblock backlog item, with the remaining-work checklist and the prioritization caveat). Branch `spike/analysis-wasm-cdata-abi` kept for the eventual real PR. Closing per spike discipline: spikes answer questions, they don't merge.

PR #1692, opened 2026-07-12
feat(goldencheck): Flip Stage 0 -- Arrow-backed seam + §8b differential (non-breaking)

benzsevern:
## Stage 0 of the goldencheck 3.0.0 Flip -- measurement only, additive/non-breaking

Per the program design's §8b Flip gate, the irreversible 3.0.0 cutover (owned Rust/Arrow contract authoritative, Polars evicted from the default path) is gated on a **measured** finding-set differential, not an assertion of zero delta. This PR builds that measurement + the Arrow-backed seam it runs on. **No authority flip, no `[polars]` removal, no version bump** -- existing output is unchanged.

### What lands
- **`ArrowColumn`/`ArrowFrame`** in `core/frame.py`: a pyarrow-backed `Column` seam whose numeric/sequence/date reductions route through the W1-W5 native kernels (`column_numeric_stats`/`sequence_analysis`/`date_freshness`/`count_outside`), the rest via `pyarrow.compute`. Lazy pyarrow import (the Polars path is untouched). `to_frame()` now also accepts a `pyarrow.Table`.
- **Parity gate** `tests/flip/test_arrow_column_parity.py` (28 tests): `ArrowColumn == PolarsColumn` on every corpus column for len/null_count/n_unique/dtype/min/max/mean/is_sorted/count_gt/diff/filter_outside/cast. `std`+float-`sum` agree within 1e-6 (variance/accumulation epsilon; unused by the seam profilers). `dtype_repr` is the owned neutral-vocab divergence (asserted explicitly).
- **§8b differential** (`scripts/flip_differential.py --fused` + `scripts/flip_corpus.py`): a 9-dataset synthetic corpus (one > the 100k sample cap) run through the 7 seam-clean column profilers over `PolarsFrame` vs `ArrowFrame`.

### §8b result (verified)
| metric | result |
|---|---|
| strict Jaccard (non-stat/non-dtype) | **1.000** every dataset |
| full Jaccard (incl. sample_values) | **1.000** every dataset |
| max stat-float delta | **0.000e+00** |
| strict-set divergences (kernel bugs) | **0** |
| dtype-vocab diffs (expected owned vocabulary) | 27 |

The owned Arrow/kernel path produces byte-identical findings to Polars on the default-scan path; only the neutral dtype strings differ (the intended owned contract). Report: `packages/python/goldencheck/docs/superpowers/specs/flip-differential-report.md`.

### Scope decision folded in
Baseline/drift/correlation (scipy + Polars group_by/pivot/join, opt-in only) will be gated behind a `[baseline]` extra at the Flip; the default scan path evicts Polars. That cutover (owned sample, authority flip, `[polars]` removal, 3.0.0 bump) is follow-on work on top of this measured, green foundation. **PyPI publish stays human-gated.**

Baseline suite: 838 engine/profilers/relations/core tests green (unchanged).

### DIFF
--- a/docs/superpowers/specs/2026-07-11-goldencheck-flip-3.0.0-design.md
+++ b/docs/superpowers/specs/2026-07-11-goldencheck-flip-3.0.0-design.md
@@ -1,0 +1,71 @@
+# GoldenCheck Flip (3.0.0) — owned-contract cutover — design
+
+Date: 2026-07-11
+Status: wave design — the Flip wave of the Arrow-native fused-scan program. HUMAN-GATED (§8b sign-off).
+Program: `2026-07-11-goldencheck-arrow-fused-scan-engine-program-design.md` (§8b Flip gate) + `...-W-path-scoping.md`.
+Base: fresh `origin/main` (W0-land…W6 all merged; 17 kernels + parity harness live). Worktree `D:\show_case\gc-flip`, branch `feat/goldencheck-flip-3.0.0`.
+
+## What the Flip actually is (recon-corrected)
+
+The Flip is the irreversible 3.0.0 cutover to the **owned contract**: the Rust/Arrow fused kernels become authoritative, the owned deterministic sample + owned dtype vocabulary become the emitted values, and Polars stops being required. Per §3 the finding set is NOT unconditionally bit-identical to the 2.x Polars path, so the cutover is gated on a **measured** differential (§8b), not an assertion of zero delta.
+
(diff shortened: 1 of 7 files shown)


benzsevern:
Superseded by the consolidated goldencheck 3.0.0 Flip PR (Stages 0+A+B+C+docs on feat/goldencheck-flip-cutover). Stage 0's Arrow seam + differential land as part of that single breaking-change PR rather than stacked, and the ruff import-sort nit that failed this PR's ci-required is fixed there.

PR #1694, opened 2026-07-12
perf(scorer): adaptive block-batching in the parallel fuzzy scorer

benzsevern:
## What

Collapses the per-block `ThreadPoolExecutor` fan-out in `score_blocks_parallel` and `score_blocks_columnar` into a small number of **adaptive batches**: big blocks stay solo (full parallelism where scoring work is real), small blocks are LPT bin-packed into `~max_workers x K` work units.

## Why

cProfile of the green zero-config 500K bench (run 29162502432, 455.67s median) shows the wall is per-block executor orchestration, not scoring math: `as_completed`=391.88s over **62,194** calls, lock.acquire=276s/483K, `_score_one_block`=272.5s/62,190. The committed config produces ~62K tiny blocks (max ~50 rows); each was its own future. This is the already-deferred "Track 1 Fix B" from the 5M bucket note in the package CLAUDE.md.

## How

- `BlockResult.n_rows`: cheap size hint populated where the group size is already computed at construction (never a per-block `.collect()` -- that's the #295/#301/#303 OOM leak). `None`-safe.
- `_plan_block_batches`: pure adaptive planner (solo threshold `GOLDENMATCH_SCORER_SOLO_BLOCK_MIN_PAIRS`, bins `GOLDENMATCH_SCORER_BATCH_BINS_PER_WORKER`).
- `_score_block_batch` / `_score_block_batch_columnar`: uniform per-batch work fns.
- `matched_pairs`, `frozen_exclude`, `target_ids`, candidate-count loop, `<=2` shortcut, emit -- all unchanged.

## Correctness

**Byte-identical clusters** is the contract: same `_score_one_block(_columnar)` calls, same pre-loop `frozen_exclude` snapshot, same pair set; only future-grouping + accumulation order change, and clustering canonicalizes `(min,max)` + is set-based. Equivalence unit tests prove batched == direct-per-block for both list and columnar paths, on both the bin and solo branches. Full scorer suite green (127 targeted tests locally).

## Verification pending (do NOT merge yet)

Remote `bench-zero-config` @500K must confirm `cluster_count=434572` / `multi_member_cluster_count=59242` unchanged vs the #1680 baseline + a wall drop (realistic floor ~270-300s), and pin the `_SOLO_BLOCK_MIN_PAIRS` default from the sweep. Spec + plan are in `docs/superpowers/`.

### DIFF
--- a/docs/superpowers/plans/2026-07-12-adaptive-block-batching.md
+++ b/docs/superpowers/plans/2026-07-12-adaptive-block-batching.md
@@ -1,0 +1,701 @@
+# Adaptive Block-Batching in the Parallel Fuzzy Scorer — Implementation Plan
+
+> **For agentic workers:** REQUIRED: Use superpowers:subagent-driven-development (if subagents available) or superpowers:executing-plans to implement this plan. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Collapse the ~62K per-block `ThreadPoolExecutor` futures in `score_blocks_parallel` (and its columnar twin) into a few dozen adaptive work units, cutting the `as_completed`/lock orchestration wall while producing byte-identical clusters.
+
+**Architecture:** Add a cheap `BlockResult.n_rows` (populated where the size is already computed for free at block construction). A pure `_plan_block_batches` planner groups blocks adaptively — big blocks (many candidate pairs) stay solo for full parallelism, small blocks are LPT bin-packed into `~max_workers × K` bins. A uniform `_score_block_batch` work fn loops `_score_one_block` over its batch. Both `score_blocks_parallel` and `score_blocks_columnar` submit one future per batch instead of per block. Scoring math is untouched, so the pair set (and therefore clustering) is invariant.
+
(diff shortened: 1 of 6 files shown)


benzsevern:
Shelving: the change is byte-identical (bench 29177262576 at 500K: cluster_count=434572, multi_member=59242, oversized=0 -- exact match to the #1680 baseline) and collapses futures 62,194->535 / lock acquires 483,579->4,877, BUT the wall did NOT improve -- 469.17s vs 455.67s baseline (~3% worse). The per-block ThreadPoolExecutor orchestration was never the wall; the '391s as_completed' in the original profile was inclusive wait-time, not recoverable overhead.

Real finding for a future pass: the recoverable cost is a 718,009-call / ~307s LazyFrame.collect() hot path -- each static block is an already-eager group frame wrapped in .lazy() (blocker.py:377) that _score_one_block immediately .collect()s back (scorer.py:1026), a pointless lazy round-trip through polars' execution engine. NOTE: the in-flight polars-eviction program is already rewriting this path to Arrow, so that collect may be removed there. Branch feat/scorer-adaptive-block-batching retained; not deleting.

Issue #422, opened 2026-05-22
Streaming-block sync: serial outer loop + fixed-cost bucketing dominate at small-block scale


benzsevern:
Follow-up from #417 (correctness fix now landed). Once blocking is correct, the next bottleneck surfaces clearly: serial outer block loop + per-block bucketing overhead.

## Observed shape

After #421, autoconfig picks `city||first_name` composite blocking on a 1.13M-row healthcare table, producing ~585K small blocks (size 8–13 in samples). Per-block log:

```
[score_buckets] entry: prepared_df.height=8 n_buckets=32
[score_buckets] t=0.00s: key_expr built
[score_buckets] t=0.00s: keyed (with_columns key_expr) in 0.00s
[score_buckets] t=0.00s: bucketed (hash %% N) in 0.00s
[score_buckets] t=0.00s: partition_by(bucket) in 0.00s -> 1 buckets
[score_buckets] t=0.00s: 1 non-empty buckets ready for scoring
[score_buckets] t=0.00s: starting bucket_score with max_workers=1 path=fast
[score_buckets] t=0.00s: bucket_score done in 0.00s, 1 blocks, N pairs
```

### 1. Inner 32-bucket partition always collapses to 1 for small blocks

When `block_size < n_buckets` (any block under 32 rows on the default config), `partition_by(bucket)` produces exactly 1 partition by definition — every row hashes to a bucket but with N < 32 rows distributed across 32 buckets, most buckets are empty and one ends up holding everything. The hash + with_columns + partition_by work is pure bookkeeping with zero scoring benefit.

**Suggested fast path:** `if prepared_df.height < n_buckets: score the block directly, skip the bucket partition.`

### 2. Outer block loop is serial

Sequential `block i/585686` lines, single python process at 100% CPU on one core. Each block scores in effectively 0ms, but per-block orchestration (key construction, scan-filter-collect from the parquet glob, dedupe set update) dominates. At sub-millisecond overhead × 585K blocks = ~10 min of pure overhead on one core.

**Suggested fix:** distribute the outer block loop across workers (ProcessPoolExecutor or Ray). Block scoring is independent except for the `matched_pairs` cross-block dedup set, which can be a final concat-and-dedup pass.

## Measured impact

~32K blocks/hr on 8 vCPU / 16 GB sandbox → ~18hr for full 585K-block table. Functional (vs. previous wedge-forever) but impractical for nightly cron.

## Sandbox repro

[...]

PR #1720, opened 2026-07-12
feat(goldenmatch): D5d score_buckets scaffolding onto the Frame seam

benzsevern:
Part of the 3.x arrow descent (plan: docs/superpowers/plans/2026-07-12-goldenmatch-arrow-descent-3x.md). Ports the bucket backend's frame scaffolding to seam ops so the arrow lane can descend into score_buckets later without touching the scoring bodies.

## What

- **New seam op `run_lengths(key)`**: adjacent-run sizes on KEY-SORTED input. polars = the exact `group_by(maintain_order=True).agg(len)` chain the workers used; arrow = vectorized `pc.run_end_encode`. The two agree ONLY on sorted input (polars groups by value so split runs merge; arrow is adjacency-based -- probed before pinning). Precondition documented in the protocol and pinned in fixtures (`test_frame_w4_ops.py`, D5d section).
- **`_score_single_pass`**: block-key derivation via `frame.derive_block_key` (the W2a twin of `_build_block_key_expr`); partition via `group_partitions`. The bucket hash stays per-lane native polars for now (`slim_df` is still `pl.DataFrame` at this layer). The #422 small-block fast path, the `del keyed`/`del bucketed` RSS points, every `stage()` wrapper, and the print instrumentation are preserved verbatim.
- **Both bucket workers** (`_score_one_bucket`, `_score_one_bucket_fast`): sort + block sizes via the seam. Seam sort is stable (`maintain_order=True`), a deterministic refinement of the old plain sort.

## Gates

- Local (both lanes, pure + native): scorer, score_buckets_multipass, bucket_febrl3_parity, native_parity, pipeline, frame_seam(+arrow), frame_w4_ops, frame_backend_differential -- all green.
- Differential harness: arrow reproduces polars on every dataset (clean_person / dirty_variant / fused_covered).
- Wall A/B dispatched (bench-zero-config, 5-run median, branch vs main): runs 29206711987 (d5d) / 29206712639 (baseline). Will post results here.

https://claude.ai/code/session_01QG75kK6gDnti5SbESeDFiW

### DIFF
--- a/docs/superpowers/plans/2026-07-12-goldenmatch-arrow-descent-3x.md
+++ b/docs/superpowers/plans/2026-07-12-goldenmatch-arrow-descent-3x.md
@@ -49,9 +49,20 @@ array-ization is leaf-local (`to_list`/`to_numpy` before rapidfuzz).
   - D5b (WALL-GATED 100K/1M): BlockResult.df becomes a seam Frame
     (materialized, score_buckets slice model); drop .lazy()/.collect()
     round-trips; candidate probe → Frame.height.
-  - D5d (heaviest, THE hot path, strictest gate): score_buckets' polars
-    scaffolding (with_columns/hash/partition_by/sort/slice, :953/:997/:1035)
-    → seam; compute core already arrow.
+  - D5d (heaviest, THE hot path, strictest gate) PORT SPEC (read 2026-07-12):
+    _score_single_pass (score_buckets.py ~:985-1070) stage-by-stage --
+    (1) keyed = slim_df.with_columns(key_expr) → frame.with_column("__block_key__",
+    frame.derive_block_key(...)) (W2a twin of _build_block_key_expr);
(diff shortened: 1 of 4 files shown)


benzsevern:
Wall A/B results (bench-zero-config, defaults, 5-run median, same runner class):

- d5d branch (run 29206711987): **438.72s** (438.05-449.24s)
- main baseline (run 29206712639): **429.08s** (426.49-456.16s)

+2.2% on the median with heavily overlapping ranges (baseline max 456.16s > branch max 449.24s) -- within run-to-run noise for this bench, and well inside the >10% owned-kernel gate. The seam calls this batch added are polars-lane passthroughs (to_frame wrap + the same maintain_order agg polars ran before), so no mechanism for a real regression is present; the run_lengths arrow twin only activates once the spine hands arrow frames (D2s series).

PR #1726, opened 2026-07-12
feat(goldenmatch): D2s-d2b THE FLIP -- Frame lane past the collect

benzsevern:
The D2s spine-descent flip (spec + consumer audit in the plan, landed with #1724; predicate #1725).

## What

`run_dedupe` skips the W5b-1 `pl.from_arrow` shim when the eager arrow stages ran and `_frame_lane_eligible` passes: the engine receives the ArrowFrame, the prep block collapses to `precompute_matchkey_transforms_frame` (every prep gate is False by construction on this lane), and the spine stays arrow through exact matching, blocking, scoring (classic + bucket backends), and clustering. `GOLDENMATCH_FRAME_LANE=0` kill-switch restores the classic shim.

- **GOLDEN BRIDGE** (`_as_polars_df`): the golden builders + goldencheck quality scoring re-enter polars at their boundary ONLY (`multi_df` / `_golden_source` / `golden_quality_scores`). Removal point: deep-D2 (golden gather on arrow). `quality_weighting` defaults **True**, so it is BRIDGED not declined -- declining would kill the lane on every default config. The predicate now declines adaptive golden rules + validation rules/auto_fix, alongside the existing C-class list.
- **Flip-surfaced fixes** (found by the e2e probes, each a raw polars read on a now-arrow frame): `find_fuzzy_matches` height/row_ids, pair-stream cross-source read, `score_blocks_parallel` candidate count via `BlockResult.n_rows`, score_buckets non-empty-bucket filter via `len()`.
- **E2E pins**: the Frame lane engages (spy on the predicate) and reproduces the classic lane exactly for exact, weighted, and bucket-backend runs; default-on quality keeps the classic lane.

## Gates

- Local: spine_dual_rep (22), pipeline, scorer, blocker, multipass, frames-out parity, differential harness (arrow reproduces polars on every dataset), native-lane febrl3 + native_parity + matchkey -- all green; ruff clean.
- Wall A/B dispatched on this branch (bench-zero-config, label d2s-d2b) -- note the zero-config bench takes the auto_config path, which the predicate declines, so it validates the CLASSIC lane is unregressed; the Frame lane's own perf gate is the e2e-pinned small-scale parity + a dedicated bench once a frame-lane-eligible bench config exists (follow-up).

https://claude.ai/code/session_01QG75kK6gDnti5SbESeDFiW

### DIFF
--- a/packages/python/goldenmatch/goldenmatch/backends/score_buckets.py
+++ b/packages/python/goldenmatch/goldenmatch/backends/score_buckets.py
@@ -1085,7 +1085,8 @@ def score_buckets(
                 print(f"[score_buckets] t={time.perf_counter()-_t0:.2f}s: partition_by(bucket) in {time.perf_counter()-_tp:.2f}s -> {len(buckets_dict)} buckets", flush=True)
 
         with stage("bucket_post_partition_setup"):
-            non_empty_buckets = [b for b in buckets_dict.values() if b.height > 0]
+            # len() works on both native frames (pl rows / pa num_rows).
+            non_empty_buckets = [b for b in buckets_dict.values() if len(b) > 0]
             n_non_empty_buckets = len(non_empty_buckets)
         print(f"[score_buckets] t={time.perf_counter()-_t0:.2f}s: {n_non_empty_buckets} non-empty buckets ready for scoring", flush=True)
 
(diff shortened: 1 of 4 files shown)


benzsevern:
Classic-lane wall A/B result (bench-zero-config, defaults, 5-run median): **361.51s** (359.91-370.47s) on this branch vs the most recent main baseline **429.08s** (426.49-456.16s, run 29206712639, same day/runner class). The zero-config bench takes the auto_config path, which the Frame-lane predicate declines, so this validates the classic lane is unregressed by the flip -- it came in well under baseline (run-to-run variance on this bench is wide; the signal is no regression, not a 16% win). The Frame lane's own wall gate lands with a frame-lane-eligible bench config (follow-up).

PR #1769, opened 2026-07-14
fix(goldenmatch): make bridge aux core-API surface polars-free (bridge stopgap PR-1)

benzsevern:
First PR of the bridge core-API arrow-port that unblocks dropping the #1747 `[polars]` CI stopgap.

**What:** ports the four remaining polars leaks on the Python core-API functions the Rust bridge invokes (`validate_dataframe`, `detect_anomalies`, `autoconfig_memory.profile_signature`, `profiler.profile_dataframe`), so those paths stay polars-free once `[polars]` is dropped from the rust bridge lanes.

**P0 finding:** the aux surface was much smaller than scoped — `auto_fix_dataframe`, `evaluate_clusters`, `compare_clusters`, `preflight`, `postflight`, `train_em`, `score_probabilistic` were already seam-ported (verified with a polars-blocked arrow probe).

**Byte-identical on the polars path.** `profile_dataframe` wraps arrow columns as `pl.Series` whenever polars is importable (every current env, incl. bridge-with-polars) — verified arrow==polars profile output. Only the pure no-polars (D6) path uses semantic dtype spelling. 57 existing profiler/validate/anomaly/autoconfig_memory tests green.

**Tripwire:** `tests/test_coreapi_aux_no_polars.py` — subprocess with `import polars` blocked runs every aux fn on a `pa.Table` and asserts `polars not in sys.modules`.

Next (PR-2): the Rust side — `convert.rs` `json_to_polars_df`→arrow + `build_probabilistic_frame`→arrow, then drop the stopgap.

Scope doc: `docs/superpowers/plans/2026-07-14-bridge-coreapi-arrow-port-stopgap-drop.md`

### DIFF
--- a/docs/superpowers/plans/2026-07-14-bridge-coreapi-arrow-port-stopgap-drop.md
+++ b/docs/superpowers/plans/2026-07-14-bridge-coreapi-arrow-port-stopgap-drop.md
@@ -1,0 +1,144 @@
+# Scope: bridge core-API arrow-port → drop the `#1747 [polars]` stopgap
+
+> **Status:** scoped 2026-07-14. This is Wave 1 Stage 5 of
+> `2026-07-14-goldenmatch-zero-config-arrow-polars-free.md`, expanded after the
+> "just port convert.rs" prototype was reverted — the port is NOT one file. It's
+> `convert.rs::json_to_polars_df` + the bridge's whole Python core-API surface.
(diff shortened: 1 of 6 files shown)


benzsevern:
Superseded by the combined self-validating PR (bridge aux ports + arrow convert + stopgap drop in one tree, so the drop validates against current source without a PyPI release).

PR #1770, opened 2026-07-14
fix(goldenmatch): polars-free the bridge aux core-API source paths

benzsevern:
Ports the **pure-Python** polars leaks on the core-API functions the Rust bridge invokes. Byte-identical on every polars-present lane (verified `profile_dataframe` arrow==polars).

- `validate_dataframe` — flip `isinstance(df, pl.DataFrame)` → `isinstance(df, pa.Table)` (the old form dereferences the lazy `pl` proxy → ImportError when polars is uninstalled).
- `detect_anomalies` — Arrow seam (`.to_arrow().column_names` / `.to_pylist()`) instead of raw `df.columns`/`df.to_dicts()`.
- `autoconfig_memory.profile_signature` — arrow-schema branch (the `auto_configure` path hit it); polars branch byte-identical.
- `profiler.profile_dataframe` — wrap arrow cols as `pl.Series` only when polars importable (byte-identical every polars-present env); else seam-native + threaded name/semantic dtype. `profile_column` discriminator made polars-safe.

New `tests/test_coreapi_aux_no_polars.py`: subprocess arrow tripwire (polars blocked) proving these aux fns run on a `pa.Table` with polars never entering `sys.modules`.

## Scope note — the `[polars]` stopgap DROP is deferred
An earlier revision of this PR attempted the full drop (convert.rs→arrow + local polars-free bridge install). CI's `rust` lane went red with **5 failures that only appear with `goldenmatch-native` PRESENT + polars ABSENT** — a combination this dev box can't reproduce (native-less; `GOLDENMATCH_NATIVE=0` exercises the seam-safe pure paths that pass). They live in native-gated match/dedupe/autofix/validate paths + the autoconfig v0-history builder (`.height`/`.columns`/`import polars`), NOT the aux fns fixed here. The convert.rs + ci.yml drop is reverted and staged for the follow-up that arrow-cleans those native paths (verified in CI). Details + the 5 named tests: `docs/superpowers/plans/2026-07-14-bridge-coreapi-arrow-port-stopgap-drop.md`.

### DIFF
--- a/docs/superpowers/plans/2026-07-14-bridge-coreapi-arrow-port-stopgap-drop.md
+++ b/docs/superpowers/plans/2026-07-14-bridge-coreapi-arrow-port-stopgap-drop.md
@@ -58,6 +58,36 @@ feature (recall-correctness sensitive, out of this plan's clean scope). The
 convert.rs→arrow + ci.yml drop are held for the follow-up that lands #1 + the
 match arrow-port together, verified in CI.
 
+### CORRECTED SCOPE (feasibility proven 2026-07-14) — the match port is FEASIBLE, not multi-week
+
+The needed arrow primitives ALREADY EXIST (verified by probe on this box, native=0):
+- `Frame.derive_matchkey([(field, transforms)…])` — arrow-native matchkey derivation
+  (the dedupe eager path uses it; `compute_matchkeys` is the OLD polars-only twin).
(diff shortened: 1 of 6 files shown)


benzsevern:
Superseded by #NEW — the complete [polars] drop (this branch's aux fixes + the match-pipeline arrow-flip that was the real blocker + bridge arrow + stopgap removal, all self-validating).

Issue #372, opened 2026-05-20
OOM in blocker: 'nan' block + max_block_size warning doesn't actually act


benzsevern:
## Summary

Follow-up to #368 (now fixed). After the psycopg3 + server-side cursor migration, the read step no longer OOMs. New blocker now triggers in the **blocking phase** on a 1,131,769-row Postgres view.

Two interacting issues:

1. **NULL values are bucketed into a single block named `'nan'`** — 11,633 records ended up in this block. NULL/missing values aren't similar to each other; they shouldn't share a block.
2. **`max_block_size=5000` warning emits but the block is processed anyway** — log message says "Processing anyway" even though the block is 2.3× the configured limit. Process OOMs during scoring of that block.

## Repro

```
$ goldenmatch sync --table gm.pubrecord_pub1
... [autoconfig committed YELLOW config] ...
Auto-configured matching rules.
Syncing...
INFO:goldenmatch.db.metadata:GoldenMatch metadata tables ready
INFO:goldenmatch.db.sync:Full scan: reading 1131769 records from gm.pubrecord_pub1
WARNING:goldenmatch.core.blocker:Block 'nan' has 11633 records (exceeds max_block_size=5000). Processing anyway.
WARNING:goldenmatch.core.blocker:Block 'smith' has 5791 records (exceeds max_block_size=5000). Processing anyway.
bash: line 15:    77 Killed                  goldenmatch sync --table gm.pubrecord_pub1
```

Wall time before OOM: 3m2s. Vercel Sandbox 4 vCPU / 8GB.

Data shape: 1.13M rows × 58 columns, real `pubrecord` data — has lots of NULL values in optional columns (phone, email, secondary address fields, specialty, etc.) and many duplicates of common last names (Smith is the largest).

### Issue 1 — `'nan'` block

Looks like the blocking-key extractor calls `str(value)` on Polars values, and NULL → `nan` (or `"nan"`). All 11,633 NULLs in whatever column is being used as a block key get bucketed together. Same root cause that turns `pl.Null` into the string `"nan"` in some pandas/polars paths.

Fix: filter NULLs out of the blocking-key extraction OR explicitly skip the `nan` block.

### Issue 2 — warning is a no-op

`Processing anyway.` literally means "we logged that this is dangerous and then did the dangerous thing." The `max_block_size` safety valve isn't load-bearing if every breach is processed anyway. Should either:

[...]

PR #1793, opened 2026-07-15
fix(goldenmatch): from_splink recognizes IS NOT NULL blocking guards (#1783)

### LINKED ISSUE #1783: from_splink: compound CustomRule blocking rules silently dropped (l.a=r.a AND l.b=r.b)
`from_splink`'s blocking recognizer drops multi-field blocking rules written as a conjunction of column equalities, even though they are standard `block_on`-equivalents. On a real production dedupe model this cost ~28 points of recall until I re-added the keys by hand.

**Version:** goldenmatch 3.3.0

**What's dropped.** A Splink settings/model whose `blocking_rules_to_generate_predictions` contains rules like:

```python
CustomRule("l.first_name = r.first_name AND l.last_name = r.last_name "
           "AND l.first_name IS NOT NULL AND l.last_name IS NOT NULL")
CustomRule("l.last_name = r.last_name AND l.phone_number = r.phone_number "
           "AND l.last_name IS NOT NULL AND l.phone_number IS NOT NULL")
```

converts with:

```
[warning] blocking_rules[3]: unrecognized blocking rule, dropped: l.first_name = r.first_name AND l.last_name = r.last_name AND ...
[warning] blocking_rules[5]: unrecognized blocking rule, dropped: l.last_name = r.last_name AND l.phone_number = r.phone_number AND ...
```

Simple single-column `block_on("npi")` converts fine; only the compound forms drop.

**Why they should convert.** `l.a = r.a AND l.b = r.b` is exactly `block_on("a", "b")` -> `BlockingKeyConfig(fields=["a","b"])`. The trailing `AND l.a IS NOT NULL AND l.b IS NOT NULL` guards are the implicit null-block semantics GM already applies (null keys don't form blocks), so they can be recognized-and-ignored rather than treated as unparseable. Splink 4's `create_settings_dict` also paren-wraps each conjunct (`(l."a" = r."a") AND (l."b" = r."b")`) which the recognizer should strip (same normalization noted in the converted-splink parity bench).

**Impact.** Faithful conversion of our production model blocked on 3 keys instead of 6, dropping pairwise recall from 0.78 -> 0.49 on a 1M dedupe (8,571 labeled records). Re-adding the three keys as native `BlockingKeyConfig` made the converted config Pareto-beat native Splink (P0.991/R0.776/F1 0.870 faithful; P0.864/R0.960/F1 0.909 upgraded, vs Splink 0.929/0.755/0.833). So the only thing standing between a clean one-command migration and beating Splink is this recognizer gap.

**Suggested scope.** Handle the conjunction-of-equalities case (N fields) plus the `IS NOT NULL` guards and Splink-4 paren-wrapping. The one genuinely hard variant is a `SUBSTR(l.first_name,1,1) = SUBSTR(r.first_name,1,1)` first-initial conjunct (transform on a blocking field) -- fine to keep dropping that with a warning, or map to a prefix transform if there's an appetite.

Happy to PR the recognizer extension if useful -- I have the repro harness (Splink settings -> from_splink -> pairwise F1 vs labels) already stood up.

benzsevern:
Fixes the silent drop of compound `CustomRule` blocking rules carrying `IS NOT NULL` guard conjuncts -- measured at ~28pp pairwise recall on the reporter's 1M production dedupe (blocked on 3 of 6 keys).

## Root cause

The recognizer already handled multi-field conjunctions, Splink-4 paren-wrapping, quoted identifiers, and same-column SUBSTR conjuncts. The one gap: `l.first_name IS NOT NULL` guard conjuncts matched neither blocking regex, and one unrecognized conjunct drops the whole rule.

## Why recognize-and-ignore is faithful

GM's blocker already implements the guard semantics: multi-field block keys are built with `pl.concat_str` (nulls poison the whole key) and null/sentinel keys are filtered before blocks form (`blocker.py`, the #372/#375 machinery). A row with a null in any key field forms no block via that key -- exactly what Splink's `IS NOT NULL` guards express. So:

- guard on a **key column** -> ignored exactly, `report.info` (does not trip `strict=True`)
- guard on a **non-key column** -> rule still converts, `report.warn` (candidates are a superset; GM cannot express a null constraint on a non-key column) -- mirrors the file's SUBSTR-widening convention
- **guards-only** rule -> dropped with the existing unrecognized warn (nothing to block on)

## Testing

The issue's two verbatim production rules convert to `fields=["first_name","last_name"]` / `["last_name","phone_number"]` with zero warnings; Splink-4 paren+quoted variants; strict=True gated both ways; settings-level test mixing a simple rule + a guarded compound rule emits both keys. 82 passed across the four from_splink test files; ruff + pyright clean. Reviewed (spec-compliance + quality): approved, edge probes confirmed fullmatch discipline, SUBSTR-key guard membership, and ReDoS-linear pattern.

Closes #1783

https://claude.ai/code/session_01R8MSaGwsjdxzf6Z7Bt3BXs


benzsevern:
@copilot resolve the merge conflicts on this branch.

Issue #1839, opened 2026-07-16
Zero-config >=50K rows silently forfeits the bucket scorer (learned strategy excluded by _use_bucket_scorer)


benzsevern:
## Summary

Zero-config runs at scale silently forfeit the bucket scorer. The auto-config learned-blocking upgrade unconditionally sets `strategy="learned"` for `total_rows >= 50_000`, and `_use_bucket_scorer` explicitly refuses `learned`. So **every zero-config dedupe at or above 50K rows falls to the legacy per-block path** -- the "5-7x default bucket scorer" does not apply to the default path at scale.

Found while attributing the #1837 wall (1M zero-config ~23 min locally). Rhymes with #1804 (bucket gates drifted between dedupe/match sites + stale 750K cap).

## Two independent exclusions

The QIS zero-config config at 1M commits `backend=duckdb` + `strategy=learned`. `_use_bucket_scorer` (`core/pipeline.py:132`) returns False on **both** counts:

```python
if backend not in (None, "polars-direct"):
    return False  # duckdb -> excluded
...
if _blk is not None and getattr(_blk, "strategy", None) not in (None, "static", "multi_pass"):
    return False  # learned -> excluded
```

Because both fire, the `height <= _BUCKET_DEFAULT_MAX_ROWS` (750K) check at line 174 is never even reached. The 750K cap is not what's keeping 1M off bucket -- it bails out well before that.

The `learned` exclusion is the load-bearing one: it's unconditional above 50K, set by the learned upgrade in `core/autoconfig.py` (~line 4318).

## Consequence: per-block overhead dominates

On the legacy path at 1M the blocking produces **200,000 blocks of 5 rows**. `apply_learned_blocks` then:

1. builds `block_lf = df[sorted(member_positions)].lazy()` x200,000, then
2. **re-materializes every one of them** in the dedup loop.

The real work (200k of 5x5 scoring) is trivial; the wall is ~400k frame operations of Python orchestration.

## Contained sub-bug: the dedup loop collects every block for nothing

`core/learned_blocking.py:~329`:

```python
for block in all_blocks:
    block_df = block.materialize().native          # collects EVERY block
    members = frozenset(block_df["__row_id__"].to_list())
```

This materializes all 200k blocks purely to hash their member set -- but `member_positions` was already in hand a few lines above, and positions are 1:1 with rows, so `frozenset(member_positions)` is an equivalent dedup key requiring zero collects.

[...]

Issue #1800, opened 2026-07-15
FS matchkeys are silently dropped in the distributed and chunked lanes (zero pairs, no warning)


benzsevern:
From the FS pipeline audit (2026-07-15, follow-up to #1798/#1799). Priority: P0 (silent wrong results).

**Distributed lane:** `_score_partition_with_config` (`goldenmatch/core/pipeline.py:4689`) loops `if mk.type != \"weighted\": continue` after the exact phase. A probabilistic matchkey routed through `GOLDENMATCH_DISTRIBUTED_PIPELINE` / `distributed/scoring.py` (call sites :249, :266, :401, :421) contributes ZERO pairs. No error, no log. Exact matchkeys still emit, so the output looks plausible.

**Chunked lane:** `ChunkedProcessor` within-chunk loop (`goldenmatch/core/chunked.py:120-132`) and `_match_against_index` (:340+) handle only exact + weighted. Same silent drop.

**Why P0:** a config that works single-box silently loses its FS scoring when it crosses into either lane. Contrast: the DataFusion backend raises `NotImplementedError` for probabilistic; Sail documents one-box fallback. These two lanes should do the same.

**Fix shape:** either (a) raise `NotImplementedError` with a clear message at lane entry when any matchkey is probabilistic, or (b) implement (distributed: route FS partitions through `score_buckets`, which is already the per-partition scorer for weighted; chunked: same). (a) is the safe first step; (b) can follow.

**Tests:** config with an FS matchkey through each lane; assert loud failure (or, once implemented, pair parity vs the single-box pipeline).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_018iZFxRDxEAhtdd1B6QcPMD

PR #1845, opened 2026-07-16
feat(autoconfig): commit lowered multi_pass blocking, reclaiming the bucket scorer (#1839)

benzsevern:
Follow-up to #1841 (compiler + harness, merged). This is the wiring that actually uses it.

> **This changes the default.** Every zero-config run `>= 50K` rows currently forfeits the bucket scorer; this stops that. Every failure path falls back to today's exact behavior, and there's a kill-switch. Details below.

## Why the wiring must live in auto-config

`strategy="learned"` defers rule learning to runtime inside `build_blocks`. That **cannot** be fixed at runtime:

```
pipeline.py:2420   if _use_bucket_scorer(config, df):   <- decides HERE
pipeline.py:2435       blocks = build_blocks(...)       <- learning happens HERE
```

Routing reads `config.blocking.strategy` **before** `build_blocks` runs, and when bucket wins it never calls `build_blocks` at all (it derives its own buckets from `passes`/`keys`). Runtime-learned rules arrive too late to matter. The rules must be in the config by routing time -- auto-config is the only seam that qualifies.

## What changed

- **`core/blocker.py`**: extract `learn_rules_for_frame()` out of `_build_learned_blocks` so config-time and runtime share **one** learner. `_build_learned_blocks` now only applies the rules; its fallbacks (no columns / no scored pairs -> static) are preserved.
- **`core/autoconfig.py`**: `_lower_learned_blocking()` learns at config time, lowers via `lower_rules_to_blocking_config` (#1841), commits `multi_pass`.

**Not new work.** This is the SAME sample-and-score pass `build_blocks` runs today, moved earlier. When lowering succeeds, `build_blocks` never runs the learned path at all.

## The payoff

```
learned strategy -> bucket?  False   # the bug
lowered strategy -> bucket?  True    # the fix
```

## Falls back, never regresses

Every failure path returns `strategy="learned"` -- today's behavior, unchanged:

- no rules learned
- **unlowerable rule**: a same-field conjunction (`last:exact AND last:soundex`). `field_transforms` is keyed by field and can't hold two chains for one field, and `learn_blocking_rules` **can** emit these (its guard is `p1.key() == p2.key()` -- field+transform, not field alone). Live path, not hypothetical.
- any unexpected exception (lowering is an optimization; it must never take down a config build)

The asymmetry is the design: **forfeiting bucket is a perf cost; lowering approximately is a silent correctness cost** (precision stays 1.0, only recall moves -- the #1800 / #1837 / #1839 shape). Always take the former.

Kill-switch: `GOLDENMATCH_LEARNED_LOWERING=0`.

## ORDERING HAZARD vs #1838 (read before rebasing either)

#1838 (#1837's fix) edits the **same six lines** of the learned upgrade. The order is load-bearing:

```python
blocking.max_block_size = _learned_block_cap(...)        # #1838: raise 5000 -> 25000
blocking = _lower_learned_blocking(blocking, df, ...)    # #1839: lowers, COPIES max_block_size
```

`_lower_learned_blocking` passes `max_block_size=blocking.max_block_size` into the lowered config, so **#1838's raise must come first** or the lowered config silently inherits the un-raised 5000 cap and #1837's regression returns -- recall-only, precision still 1.0, invisible.

#1838 is **not** superseded by this PR. Bucket honors the same drop cap (`score_buckets.py:1021`: `if skip_oversized and size > max_block_size` -> skipped entirely), so the cap problem follows the traffic onto the bucket path. Both fixes are needed, and they compose in this order.

## Verification

- new wiring tests: 8 passed (lowering fires; all fallbacks return `learned`; bucket payoff)
- with #1841's parity + #1840's dedup suites: 36 passed
- regression sweep (`learned or block or autoconfig or bucket`): **1456 passed**
- full-package `ruff check`: clean
- Pre-existing, NOT from this PR (each reproduced on trees without this change): `test_autoconfig_benchmarks.py` x3 (`FileNotFoundError` -- benchmark CSVs absent locally), `test_native_fs_ne.py::test_native_numpy_parity_ne`, `test_score_buckets_debug_mode.py::test_debug_flag_does_not_change_output`

## NOT verified: the wall

The routing flip is proven structurally and pairs are identical to the learned path on clean data (#1841's matrix). **The seconds are not measured** -- 1M zero-config OOMs the dev box. Per `feedback_verify_perf_not_just_ship`, shipping the symbol isn't the same as moving the wall. The CI quality gate is the right judge of both the wall and any quality delta, and I'd treat #1839 as open until it reports.

https://claude.ai/code/session_0179KzseadM9N6we7bGHAWg9

### DIFF
--- a/packages/python/goldenmatch/goldenmatch/core/autoconfig.py
+++ b/packages/python/goldenmatch/goldenmatch/core/autoconfig.py
@@ -3515,7 +3515,8 @@ def _lower_learned_blocking(
     )
 
     try:
-        lf = df.lazy() if not hasattr(df, "collect") else df
+        # df is an eager DataFrame here; learn_rules_for_frame collects internally.
+        lf = df.lazy()
         rules = learn_rules_for_frame(lf, blocking)
         if not rules:
             logger.info("Learned-blocking lowering: no rules learned; keeping strategy='learned'.")
@@ -3538,11 +3539,11 @@ def _lower_learned_blocking(
         )
         return blocking
 
+    passes = lowered.passes or []
     logger.info(
         "Lowered learned blocking to multi_pass (%d rows, %d passes: %s) -- "
         "keeps the bucket scorer instead of forfeiting it to the legacy path.",
-        total_rows, len(lowered.passes),
-        [f"{p.fields}" for p in lowered.passes],
+        total_rows, len(passes), [f"{p.fields}" for p in passes],
     )
     return lowered
 
--- a/packages/python/goldenmatch/goldenmatch/core/autoconfig.py
+++ b/packages/python/goldenmatch/goldenmatch/core/autoconfig.py
@@ -3475,6 +3475,78 @@ def _maybe_promote_blocking_to_adaptive(
     return blocking.model_copy(update={"strategy": "adaptive"})
 
 
+def _lower_learned_blocking(
+    blocking: BlockingConfig,
+    df: pl.DataFrame,
+    total_rows: int,
+) -> BlockingConfig:
+    """Learn the blocking rules NOW and commit them as multi_pass passes (#1839).
+
+    ``strategy="learned"`` defers rule learning to runtime, inside build_blocks.
+    That forfeits the bucket scorer for EVERY zero-config run >= 50K rows:
+    ``_use_bucket_scorer`` refuses ``learned`` (bucket derives block keys from
+    ``passes``/``keys``, and a learned config carries neither), and the routing
+    decision is made BEFORE build_blocks runs -- so runtime-learned rules arrive
+    too late to matter. Learning at config time and lowering the rules into
+    multi_pass closes the gap using machinery that already exists (#1826
+    ``field_transforms``; score_buckets already honors it).
+
+    The learning pass is not new work -- it is the SAME sample-and-score pass
+    build_blocks would run at runtime, moved earlier. When the lowering succeeds,
+    build_blocks never runs the learned path at all.
+
+    Falls back to ``strategy="learned"`` (today's behavior, unchanged) whenever
+    the rules can't be learned or can't be lowered EXACTLY -- e.g. a same-field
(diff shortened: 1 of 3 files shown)


benzsevern:
Closing: **the premise is false.** #1839 claims every zero-config run >=50K forfeits the bucket scorer. It doesn't.

`_use_bucket_scorer`'s FIRST branch is:

```python
if backend == "bucket":
    return True   # explicit choice -- honored at any size
```

and the planner (`autoconfig_planner_rules.py:41`) sets exactly that whenever the kernel is present:

```python
return "bucket" if native_enabled("block_scoring") else "polars-direct"
```

So with native, `strategy="learned"` is **never consulted** — the strategy check is unreachable. Measured on the real auto-config output (not a hand-built config):

```
GOLDENMATCH_NATIVE=auto   native.block_scoring=True
  backend=bucket  strategy=learned
  use_bucket_scorer=True      <- #1839 predicted False
```

The forfeit only happens **without** the native kernel (`backend=polars-direct` -> the strategy check fires). And in exactly that case bucket is not a win anyway — from the planner's own note:

> bucket + the Python scorer is ~on par / slightly slower, so **the win requires the kernel**

So the forfeit occurs only where the fast path wouldn't be faster. There is nothing to reclaim.

**Where the wrong premise came from:** #1841's "payoff proven" (`learned -> False`, `lowered -> True`) used a hand-built `GoldenMatchConfig(backend=None)`, not what `auto_configure_df` actually emits. I never checked the real config until a diagnostic for #1846 printed `cfg.backend=bucket` and contradicted it.

This PR also carries two real bugs found while validating it, now moot but worth recording:
* the compiler silently drops `blocking.keys` (`[['full_name']] -> []`)
* `strategy="learned"` already carries passes (5 on historical_50k); the lowering **replaces** them with 3, narrowing candidate generation

Superseded by reality. #1840 (dedup: ~400K collects -> 3) stands on its own and is unaffected.

Issue #1846, opened 2026-07-16
historical_50k f1_probabilistic 0.83 -> 0.33: #1834 inflates EM lambda to 0.887; compute_thresholds has no guard for lambda > 0.5


benzsevern:
## Summary

`historical_50k f1_probabilistic` has collapsed **0.8279 -> 0.3335** on the **native** path on `main`. Nothing caught it, because `quality_gate` did not run on any of the PRs that could have caused it, and does not run on `main` at all.

This is a **recall-only** loss on the probabilistic path -- precision stays clean, so no other signal moves. Same silent shape as #1800 / #1837.

## Evidence it is on main, not in any open PR

| run | native | historical_50k f1_probabilistic |
|---|---|---|
| local, `--native 0` | off | **0.8281 PASS** |
| local, `--native 0`, `GOLDENMATCH_LEARNED_LOWERING=1` (#1845) | off | **0.8287 PASS** |
| CI `quality_gate` on **#1838** | on | **0.3335 FAIL** |
| CI `quality_gate` on **#1845** | on | **0.3335 FAIL** |

#1838 and #1845 share **no unique change** -- #1838 raises `max_block_size` (a no-op at 50,578 rows: `_compute_max_safe_block(50578, True) = 1264`, so `max(5000, 1264) = 5000`), #1845 lowers learned blocking to multi_pass. They fail with byte-identical numbers. The only thing they share is `main`.

Both PRs are red purely because their paths make the gate RUN. They are messengers.

## Why nothing caught it

`quality_gate` is **path-filtered**, and it was `skipping` on exactly the PRs most able to break this:

| PR | changed | quality_gate |
|---|---|---|
| #1829 | `score_buckets` **FS bucket lane** auto-split + dense-matrix guard | **skipped** |
| #1834 | `probabilistic`: treat missing values as unobserved | **skipped** |
| #1836 | `probabilistic`: blocking fields keep the fixed prior | **skipped** |

It is also `skipped` on every `main` run (verified on `3f1bb4d43`, `49f773026`), so a green main means nothing here.

Note #1836's whole job was **restoring this benchmark to 0.8279** after #1835 collapsed it to 0.57 -- and its gate was skipped too, so that restoration was never gate-verified either. The current 0.3335 is *worse* than the #1835 collapse.

## Why native specifically

The gate runs native on purpose (from `ci.yml`):

[...]

PR #1834, opened 2026-07-16
fix(probabilistic): treat missing values as unobserved

benzsevern:
Closes #1819

## Summary
- represent missing regular-field comparisons as unobserved evidence and exclude them from discrete and continuous EM likelihoods
- normalize linear scores over only the observed regular fields, with a neutral 0.5 result when a pair supplies no evidence
- apply the same policy across Python scalar/vectorized/batched scoring, TypeScript, and all Rust native/fused FS kernels
- bump persisted EM models to schema v2 and reject v1 models trained with the legacy null-as-disagreement meaning
- gate old native wheels and bump goldenmatch-native to 0.1.17 so wheel skew falls back safely

## Verification
- Python probabilistic/native/API regression selection: 271 passed
- Python focused post-rebase selection: 166 passed
- Python fused/native capability selection: 41 passed
- TypeScript probabilistic/NE/parity selection: 93 passed
- TypeScript focused post-rebase selection: 71 passed
- TypeScript typecheck and production build: passed
- Rust cargo check and clippy -D warnings: passed
- Ruff and git diff --check: clean

Issue #1826, opened 2026-07-16
probabilistic bucket scorer builds a dense NxN matrix per block (no oversized guard): 388K block -> 1.1 TiB alloc / OOM


benzsevern-mjh:
Surfaced while gathering the #1811 trace. Distinct bug (and arguably a bigger scale blocker): the **probabilistic (Fellegi-Sunter) bucket scorer allocates a dense N×N float64 matrix per block, with no oversized-block guard**, so one large block is an instant OOM.

**Version:** goldenmatch 3.3.1 (`main`, #1783/#1794/#1799/#1811-era), native 0.1.15, `GOLDENMATCH_NATIVE=1`, 64GB node.

**What happens:**
```
[score_buckets] entry: prepared_df.height=996558 n_buckets=32
[score_buckets._resolve_fast_path] declined: mk.type='probabilistic' (need 'weighted')
[score_buckets] starting bucket_score ... path=probabilistic_vectorized
numpy._core._exceptions._ArrayMemoryError: Unable to allocate 1.10 TiB for an array
  with shape (388552, 388552) and data type float64
```
A single 388,552-record block → the vectorized probabilistic path builds a dense `(388552, 388552)` similarity matrix (1.1 TiB) and dies. The weighted path has a bounded fast-path (`_resolve_fast_path`); the probabilistic path declines it and falls to the dense O(n²) matrix with no sub-blocking / cap / streaming. Same pattern as #1792/#1811 — the FS/probabilistic lane lacks the scale-safety the weighted lane has, here at the *scoring-matrix* level rather than the pair accumulator.

**How the mega-block formed (relevant to #1783):** the input is a `from_splink` conversion of a Splink `last+zip5+first-initial` block: `l.last_name=r.last_name AND l.zip5=r.zip5 AND SUBSTR(l.first_name,1,1)=SUBSTR(r.first_name,1,1)`. Post-#1783 the converter now recognizes it, but GoldenMatch blocking transforms are **key-level**, so `substring:0:1` is applied to *all* fields, collapsing the key to `(last_name[0], zip5[0], first_name[0])` — first-initials only. The converter warns about the widening, but the result is an ultra-coarse key that buckets 388K rows together at just 1M rows (vastly worse at 14M). So #1783 fixed the drop but produces a block shape the probabilistic scorer can't survive.

**Two things worth fixing independently:**
[...]

PR #1848, opened 2026-07-16
diag(#1846): CI-vs-local fact dump for the historical_50k native collapse [DO NOT MERGE]

benzsevern:
**DIAGNOSTIC ONLY. DO NOT MERGE.** The `ci.yml` step here is temporary.

`historical_50k f1_probabilistic` is **0.82 on local Windows** and **0.3335 on CI/Linux** -- same commit, same flags. Everything reasoned about this has been wrong, so this stops reasoning and diffs the two hosts.

## Ruled out BY MEASUREMENT (not argument)

| hypothesis | how it died |
|---|---|
| a bad commit (#1836 / #1834 / #1829) | all 5 rungs back through #1826 **PASS** locally, incl. main HEAD (0.8298) |
| the env knobs | `native=auto/1/0` x `memory=0/1` -- all 4 **PASS** locally, incl. CI-exact (0.8237) |
| any open PR | **#1847** changes only `ci.yml`/`filters.yml` + a new script and still **FAILS 0.3335** |

So: **main + Linux.** Not a PR, not a commit, not a flag.

`historical_50k` is the **only** dataset >= 50,000 rows -- the only one that trips the learned-blocking upgrade. Every smaller dataset (febrl3, ncvr_synthetic, anchor_person_match) is clean on **both** hosts, including their `f1_probabilistic`. That is the sharpest clue available.

## What this adds

`scripts/diag_1846.py` -- same script, both hosts, `DIAG1846`-prefixed lines so it greps out of a CI log. Dumps host/deps, native capability flags, the committed config, EM blocking fields, the routing decision, and the blocking outcome (counts, pairs, over-cap).

## Local Windows baseline (for the diff)

```
os=Windows  cpu_count=12  python=3.13.12
polars=1.40.1  numpy=2.4.4  pyarrow=24.0.0  jellyfish present (soundex Smith=S530)
native: block_scoring=True  fs_bucket=False  autoconfig=True  scoring=False
cfg.backend=bucket  strategy=learned  max_block_size=1264  skip_oversized=True
keys=[[full_name]]  passes=[full_name x3, first_and_surname x2]
em_blocking_fields=[full_name, first_and_surname]
routing.use_bucket_scorer=True
blocks=3359  total_rows=83208  max_size=1262  pairs=12608858  over_cap=0
```

## Two claims of mine this ALREADY corrected

1. **`max_block_size=1264`, not 5000.** So #1838's `_learned_block_cap` **is live** at 50,578 rows -- it raised the cap. I had called it a no-op there, assuming the incoming cap was 5000. It wasn't.
2. **`use_bucket_scorer=True` despite `strategy=learned`.** The planner sets `backend=bucket`, and `_use_bucket_scorer`'s first branch (`if backend == "bucket": return True`) fires **before** the strategy check. So this dataset **already uses the bucket scorer** -- which undercuts #1839's "zero-config >=50K forfeits bucket" premise for exactly the dataset that is failing.

## Next

Read `DIAG1846` off the runner, diff against the block above, and let the delta name the cause. Then revert the `ci.yml` step and fold the finding into #1846 (whose current "suspects + bisect" plan is refuted and needs rewriting).

https://claude.ai/code/session_0179KzseadM9N6we7bGHAWg9

### DIFF
--- a/.github/filters.yml
+++ b/.github/filters.yml
@@ -7,6 +7,13 @@
 # HERE + the matching `changes.outputs.*` line + the job's `if:` gate in ci.yml.
 ci_workflow:
   - '.github/workflows/ci.yml'
+  # #1846: workflow_lint validates path-filter COVERAGE, so an edit to the
+  # filters themselves (or to the checker) must run it. Without this, the one
+  # file the coverage check exists to validate was the one file that couldn't
+  # trigger it -- the same "a filter can't catch what it doesn't list" hole,
+  # one level up.
+  - '.github/filters.yml'
+  - 'scripts/check_filter_coverage.py'
 python_root:
   - 'pyproject.toml'
   - 'uv.lock'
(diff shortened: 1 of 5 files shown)


benzsevern:
Diagnostic only -- served its purpose. Root cause found and fixed in #1851 (#1834's missing-value semantics; see #1846 for the full chain). The temporary `ci.yml` diagnostic step in this branch must never merge, hence closing rather than leaving it open.

What this branch established:
* config / planner / blocking / the 12,608,858 candidate pairs are byte-identical Windows vs Linux CI -> the divergence was never environmental
* EM is deterministic (lambda=0.88707, identical weights both hosts) -> no race, no platform effect
* my local runs disagreed only because they imported a 819-commits-stale tree via a venv from another checkout (`module.__file__` unverified)

PR #1966, opened 2026-07-21
feat(goldengraph): template-free NL multi-hop routing to the LLM-free chain walk

benzsevern:
## What and why

The diagnosed #1 GoldenGraph answer-quality gap is **path-selection**: retrieval pulls a ball that *contains* the answer chain (bridge-recall ~1.0), but the right chain is read out of it only ~27% of the time. Synthesis-given-the-gold-chain is a perfect 1.0 — so the loss isn't reasoning, and the two cheap fixes (topology prune, query-name embedding) were already refuted (`results/RESULTS_PATH_AWARE_RETRIEVAL.md`).

`trace_chain` is the deterministic, LLM-free multi-hop walk that *does* answer well — but it only fired on the engineered template (`"Starting from X, follow the relation R1, then R2."`). Real natural-language questions (`"Who is married to the person who directed Inception?"`) never matched it and fell through to LLM synthesis-over-the-ball. This generalizes the chain extractor to real NL.

## Changes (goldengraph — the feature)

- **`route._extract_nl_chain_slots`** — recovers `(anchor, relation_chain)` from free NL, grounded in the slice's own vocabularies: anchor = longest stored **entity name** (token-bounded, no substring grounding); relations = **predicate ids** whose salient token appears, bridged to the question's noun form by a shared-≥5-char-stem rule. Wired into `classify_query` for MULTI_HOP + LOOKUP; no-op without vocab (back-compat).
- **Deterministic** — `_find_anchor`/`_best_predicate_for_token` scan `sorted(...)` with explicit tie-breaks; the predicate sort+tokenize is hoisted to once-per-query (`_predicate_token_index`).
- **Multiplicity, not a set** — each content word → one predicate occurrence, so a repeated relation yields a repeated hop.
- **Order is a HINT the graph validates** — `answer._trace_chain_any_order` trusts the hint order when it completes; only if it fails does it try permutations and require a **unique** completing terminal, else abstain.
- **Conservative by construction** — fires only with a grounded anchor + ≥1 grounded predicate; a completeness guard abstains when an unmapped content word sits before an `of`/`by` marker. A mis-parse that fails to walk falls through to today's retrieval+synthesis path (never worse than status quo); the residual (a grounded mis-parse completing to a different valid terminal) is rare and no worse than the LLM path.

## Measurement

On the engineered corpus (its `QAItem`s carry the gold `relation_chain`), reconstructed into a `StubGraph`, no wheel/LLM: NL phrasings reach **96.8% answer accuracy — parity with the engineered-template ceiling** — up from **0%** LLM-free before. By hop: 2-hop 100%, 3-hop 93%, 4-hop 97%. `tests/test_nl_chain.py` (wheel-free, 16 tests); full goldengraph suite **349 passed**; `ruff` clean.

## Note on scope: er-kg-bench demo artifacts (pre-existing staleness, not the feature)

This PR also regenerates `benchmarks/er-kg-bench/demo/{DEMO.md,demo.html,demo.snapshot.json}` (switching the highlighted under-merge exemplar `Q7817`/WHO → `Q7184`/NATO + the run date). **This is unrelated to the routing change** — the demo uses `GoldenMatchAdapter`, not `goldengraph.route`/`answer`. The committed artifacts were stale on `main`: `demo/run_demo.py` deterministically now picks NATO (4 surface forms) as the cleanest under-merge protagonist, but the committed files still showed WHO, and the `bench-er-kg` lane's "committed and current" gates (steps 18/19) surfaced it once this PR triggered that lane. I regenerated them via `demo/run_demo.py` so those gates pass; `demo/run_demo.py --check` is green locally. (For reviewers: `DEMO.md` is `tier1_under_merge` — a single-entity illustration, NATO 4→1 — while `demo.snapshot.json`'s `before_answer`/`after_answer` are the separate keyed **count** transcript, "how many distinct orgs…", 10→4; the two tiers legitimately differ, and the recorded transcript is refreshed only by the keyed step, not this offline regen.) Happy to split this into its own PR if preferred.

## Checklist

- [x] `ruff check` clean on changed files
- [ ] Package `CHANGELOG.md` updated if behavior changed (pre-1.0 KG engine, no shipped-API change; routing is additive + gated by vocab)

https://claude.ai/code/session_018aF5eunXPReEq1SYDNiYY5


benzsevern:
Thanks for the review — addressed the route.py findings and rebased:

- **Determinism (`_best_predicate_for_token`, `_find_anchor` tie-break):** both now scan their inputs `sorted(...)` with explicit tie-breaks (predicate: strict-improvement over the sorted scan → lexicographically-smallest wins on a score tie; anchor: longest → earliest-occurrence → lexicographic). A `set` of `entity_names`/`predicates` can no longer make routing order-dependent. Fixed.
- **`_find_anchor` substring fallback:** removed the bare `ql.find(n)` fallback — token-bounded `" n "` only, so `"Acme"` can't ground inside `"Acmeville"`. Added `test_nl_anchor_is_token_bounded_not_substring`. (Also fixed a latent off-by-one: it now returns the name's own offset, so the anchor-blanking aligns exactly.)
- **Completeness-guard breadth:** I deliberately kept the markers to `{of, by}`. Broadening to `{to, in, at, for, …}` looks right but over-abstains — `"leads to"`, `"results in"` are ordinary verb complements, and it measured the LLM-free accuracy straight to **0%**. `of`/`by` are the high-signal relation markers. The residual conjunction case (`"directed X and produced it"`, missed relation before a pronoun) is a documented boundary for the embedding/`LLMQueryClassifier` tier — and by design it's never worse than today: an unfired/partial chain just falls through to the existing retrieval+synthesis path.
- **nopolars docstring comment:** that file landed in #1962 (already merged); this branch was carrying a duplicate of it, so I rebased onto current `main` — the nopolars files are out of this PR's diff now.

Engineered-corpus measurement unchanged after the fixes: NL phrasings **96.8% LLM-free**, parity with the template ceiling. Full goldengraph suite green (347 passed).

---

benzsevern:
Second round addressed:

- **`_trace_chain_any_order` ambiguity (answer.py):** good catch on the safety contract. Fixed, but with a measured nuance — the extracted order is a *proximity hint* that encodes the question's phrasing, so: (1) if the **hint order completes**, trust it (the graph confirmed the reading the question expressed); (2) only if the hint order *fails* do I fall back to other permutations and require a **unique** completing terminal, else abstain. Requiring uniqueness across *all* orders (including the hint) measured a drop from **96.8% → 29.8%** LLM-free on the engineered corpus — the graph is dense, many orders complete to different nodes, and the hint is the right one, so discarding its signal abstains on cases that were correct. New tests: `test_order_tolerant_trusts_hint_when_it_completes` and `test_order_tolerant_abstains_on_ambiguous_fallback`.
- **`ask` calling `entities()` twice:** fixed — `slice_graph.entities()` is now materialized once and passed to both `_slice_predicates`/`_slice_entity_names` (they take an optional `entities=` to reuse it).

Also, unrelated to this PR's code but surfaced by the `bench` lane it triggered: the committed `demo/DEMO.md` was stale (the demo deterministically picks `Q7184`/NATO now, not `Q7817`/WHO — the demo uses `GoldenMatchAdapter`, not the routing I changed). Regenerated it in a separate commit so the lane's "committed and current" check passes.

96.8% LLM-free accuracy holds; goldengraph suite 349 passed.

---

benzsevern:
Round 4:

- **`_best_predicate_for_token` sorting per token** — fixed. Added `_predicate_token_index()` which sorts `predicates` and extracts each predicate's salient tokens **once per query**; the per-word loop reuses it. Determinism and the 96.8% accuracy are unchanged.
- **DEMO.md vs snapshot/HTML "out of sync"** — this one is a false positive: `DEMO.md` and `demo.snapshot.json`/`demo.html` are two different demo tiers, not two renderings of one thing. `DEMO.md` is the **offline, deterministic under-merge illustration** for a single entity (NATO: 4 surface forms → 4 org-nodes before ER, 1 after — internally consistent). The snapshot's `before_answer`/`after_answer` ("10 distinct → 4") is the **keyed aggregate LLM transcript** for a *different* question ("How many distinct organizations… list each"), which the workflow regenerates only in the keyed step and never gates (it's non-deterministic). My offline regen updated only the deterministic scaffolding (`query`, graph structure, `recorded_at`) — `demo/run_demo.py --check` (the step-19 gate) passes on it — and deliberately left the recorded transcript to the keyed run. So the two artifacts legitimately reference different questions/counts; nothing is misleading within either tier.

goldengraph suite 349 passed; ruff clean.

---

PR #1969, opened 2026-07-21
feat(goldengraph): semantic synonym bridge for NL multi-hop routing

benzsevern:
## What and why

Follow-up to #1966 (template-free NL multi-hop routing). That extractor grounds relation words to predicates lexically — exact + a shared-≥5-char-stem rule — which bridges *morphological* variants (`director`→`directed_by`, `location`→`located_in`) but **not pure synonyms** (`spouse`→`married_to`, no shared characters). Those hit the completeness guard and abstained (safe, but left multi-hop recall on the table). This adds the semantic tier.

## Changes

- **`route._extract_nl_chain_slots(..., embedder=)`** — before the completeness guard abstains, it bridges **exactly the guard-triggering words** (uncovered *and* immediately before an `of`/`by` marker) via embedding cosine against the predicate phrases. `_embed_bridge_predicate` picks the closest predicate above `GOLDENGRAPH_NL_EMBED_BRIDGE_MIN` (default 0.55; deterministic tie-break via the pre-sorted index).
- **Scoped + conservative** — only the words that would otherwise abstain are embedded (not every content word), so it lifts the synonym boundary without opening a spurious-match surface. `embedder=None` is **byte-identical to #1966** (all prior tests unchanged).
- **Threading** — the `embedder` flows `ask` → `resolve_profile` → `classify_query` → the extractor (the same embedder `ask` already carries). An embedder that raises is caught → no bridge → the guard abstains as before.
- Needs a **semantic** embedder to bridge true synonyms; the no-torch char-ngram embedder only covers the morphological cases the stem rule already handles.

## Testing

- `tests/test_nl_chain.py`: `_SynEmbedder` stub (3-D, `spouse`~`married_to`, `friend` orthogonal). `"Who is the spouse of the director of Inception?"` now fires the full 2-hop chain → `Emma Thomas`; `"Who is the friend of …"` stays below the cosine floor and abstains; without an embedder it still abstains (unchanged).
- Full goldengraph suite: **354 passed, 8 skipped**; `ruff` clean.

## Note: end-to-end reachability (found while wiring the paid headline)

The bench's goldengraph engine runs `ask` in **`mode="local"`** (`engines/goldengraph.py`), but NL routing (chain extraction) only fires in **`mode="auto"`**. So exercising this end-to-end in the `bench-graphrag-qa` lane needs `GOLDENGRAPH_QA_MODE=auto` (I dispatched a MuSiQue N=50 run that way). Making the chain-attempt fire inside `local`/`hybrid` before synthesis — so the win is the *default*, not auto-only — is the natural next follow-up (documented in `CLAUDE.md`).

## Checklist

- [x] `ruff check` clean on changed files
- [ ] Package `CHANGELOG.md` updated (pre-1.0 KG engine; additive + gated by an optional embedder)

https://claude.ai/code/session_018aF5eunXPReEq1SYDNiYY5

---
_Generated by [Claude Code](https://claude.ai/code/session_018aF5eunXPReEq1SYDNiYY5)_

### DIFF
--- a/packages/python/goldengraph/CLAUDE.md
+++ b/packages/python/goldengraph/CLAUDE.md
@@ -126,3 +126,25 @@ query-name embedding — were already refuted, see `results/RESULTS_PATH_AWARE_R
   (96.8%) — up from **0%** LLM-free before (they fell to ~0.15–0.28 paid synthesis).
   By hop: 2-hop 100%, 3-hop 93%, 4-hop 97%. Tests: `tests/test_nl_chain.py`
   (wheel-free; fires, safe abstentions, ordering, multiplicity, end-to-end `ask`).
+
+## NL chain routing — semantic synonym bridge (2026-07-21)
+Follow-up to the template-free NL routing (#1966): the lexical stem rule grounds
+morphological variants ("director"->"directed_by") but not pure synonyms
+("spouse"->"married_to", no shared characters), which hit the completeness guard
+and abstained. `route._extract_nl_chain_slots(..., embedder=)` now bridges exactly
+those guard-triggering words (uncovered AND before an "of"/"by" marker) by embedding
+cosine against the predicate phrases: `_embed_bridge_predicate` picks the closest
+predicate above `GOLDENGRAPH_NL_EMBED_BRIDGE_MIN` (default 0.55). Scoped to the
+guard words only (not every content word) so it lifts the synonym boundary without
+a spurious-match surface; `embedder is None` is byte-identical to #1966 (all prior
+tests unchanged). `embedder` threads `ask` -> `resolve_profile` -> `classify_query`
+(the embedder `ask` already carries). Needs a SEMANTIC embedder to bridge true
+synonyms -- the no-torch char-ngram embedder only shares the morphological cases the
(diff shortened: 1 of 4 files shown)


benzsevern:
Both robustness points fixed (b20bc35):

- **`_embed_bridge_min` nan/inf/out-of-range** — now rejects non-finite (`nan`/`inf` → default 0.55) and clamps to `[0,1]`. A `nan` floor would have silently disabled bridging and a negative floor would bind anything — both env-misconfig footguns.
- **`embed()` length mismatch** — the bridge now materializes `embed()` output to a list and requires it match the input length; a wrong-length return is a contract violation, so it's treated as an embedder failure (no bridge → the guard abstains) instead of letting `zip` truncate and shift which predicate wins. Added `test_nl_broken_embedder_length_mismatch_abstains`.

Re the `github-code-quality` "unused `tok_vecs`": that's a false positive — `tok_vecs` is used in the `is not None` check, the new `len(...)` guard, and the `zip(...)` loop (`ruff --select F841` is clean). No change needed there.

19 NL tests pass; ruff clean.

---

PR #2062, opened 2026-07-23
feat(goldengraph): gated top-K edge rerank for the hybrid ball (default off)

benzsevern:
## Why (measured motivation)

On MuSiQue N=150 (hybrid synthesis default-on, `passage_k=10`) the largest QA miss bucket is **SYNTHESIS:53**, and **47/53 (89%)** of those had the gold answer's relationship edge **PRESENT in the retrieved ball** yet the model wrote a wrong answer.

The retrieved hybrid ball routinely holds ~1,500-2,500 edges, and `goldengraph/synthesize.py::_format_subgraph` serializes **every** edge into the hybrid prompt's "Relationship graph" slot **with no cap**. Hypothesis: the answer edge is a needle buried in a huge, noisy graph context. This lever prunes the hybrid ball to the top-K question-relevant edges before synthesis, to test whether noise-reduction recovers some of those 47 misses.

## What

Gated, default-OFF top-K edge rerank of the hybrid ball, mutually exclusive with the existing `path` filter.

- `subgraph_filter.rerank_subgraph_edges(subgraph, seeds, *, question, embedder, top_k) -> dict`
  - Scores each edge by cosine similarity between the embedding of `question` and the embedding of the edge's serialized text, using the **same name-keyed form the model sees**: `"{subj_name} {predicate} {obj_name}"` (names from the subgraph entities, falling back to the raw id when a name is missing).
  - **Batched**: one `embedder.embed` call for all edge texts + one for the question (no per-edge round-trips), mirroring `seed_by_query`.
  - **ALWAYS retains every edge incident to a seed entity id** (connectivity/anchor preservation), regardless of score; the remaining budget (`top_k` minus retained seed-incident edges) is filled with the highest-scoring non-seed-incident edges (ties broken by original edge order).
  - `top_k >= len(edges)` -> returns the subgraph unchanged (no-op, same object).
  - Returns a NEW dict: `edges` pruned to the kept set (original order preserved), `entities` reduced to those referenced by the kept edges UNION the seed ids (a seed entity is never dropped), all other keys preserved. The input is never mutated.
- `answer.ask` hybrid branch: when `_hybrid_filter_mode() == "rerank"`, route the same `subgraph` through the rerank using the `embedder` already in scope, then rebuild `id_to_name` / `seed_names` from the pruned subgraph exactly as the `path` branch does.
- `answer._hybrid_filter_topk()` reads `GOLDENGRAPH_HYBRID_FILTER_TOPK`.

## Env flags

| Flag | Values | Default |
| --- | --- | --- |
| `GOLDENGRAPH_HYBRID_FILTER` | unset / `` / `none` (off), `path`, `rerank` | off |
| `GOLDENGRAPH_HYBRID_FILTER_TOPK` | int edge budget for `rerank` (non-int -> default) | `40` |

## Byte-identical when off

With `GOLDENGRAPH_HYBRID_FILTER` unset / empty / `none`, neither the `path` nor the `rerank` branch fires and the subgraph handed to `synthesize_hybrid` is untouched. `rerank` is opt-in only.

## Measurement pending

The A/B measurement is **not yet run** and will decide ship/drop:

```
bench-graphrag-qa.yml  mode=env_ab  GOLDENGRAPH_HYBRID_FILTER:,rerank
```

Same-graph env-A/B (build once, answer under both configs), so the metric delta is purely the knob's effect (`support_recall` is the control -- a filter can't touch retrieval). Draft until that lands.

## Tests

`packages/python/goldengraph/tests/test_subgraph_rerank.py` (pure, no native/LLM/network) covers: mode off is identity; top-K keeps exactly the highest-scoring non-seed edges (one batched embed call); seed-incident edges retained even when lowest-scoring; `top_k >= len(edges)` no-op; entity set == kept-edge endpoints UNION seeds with no seed dropped and input not mutated; missing-name id fallback; non-int TOPK -> default 40. Plus a rerank wiring test through `ask` in `tests/test_hybrid_synthesis.py`. Local run: **27 passed** (rerank + hybrid + subgraph_filter suites).

### DIFF
--- a/packages/python/goldengraph/goldengraph/answer.py
+++ b/packages/python/goldengraph/goldengraph/answer.py
@@ -232,10 +232,22 @@ def _canon_query_rel(rel, schema):
 def _hybrid_filter_mode() -> str:
     """Hybrid subgraph filter selector, read at call time. "" / "none" / unset =
     off (pass the full ball; the measured 0.420 control). "path" = path-preserving
-    prune (`subgraph_filter.filter_subgraph_to_paths`)."""
+    prune (`subgraph_filter.filter_subgraph_to_paths`). "rerank" = top-K
+    question-relevant edge prune (`subgraph_filter.rerank_subgraph_edges`)."""
     import os
 
     return os.environ.get("GOLDENGRAPH_HYBRID_FILTER", "").strip().lower()
+
+
+def _hybrid_filter_topk() -> int:
+    """`GOLDENGRAPH_HYBRID_FILTER_TOPK` (default 40) -- the `rerank` mode's edge budget.
+    Non-int -> 40."""
+    import os
+
+    try:
+        return int(os.environ.get("GOLDENGRAPH_HYBRID_FILTER_TOPK", "40"))
+    except ValueError:
+        return 40
 
 
 def _local_filter_mode() -> str:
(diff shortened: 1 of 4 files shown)


benzsevern-mjh:
Refuted by measurement (run 30035422835, env_ab none-vs-rerank@40): answer_match -0.04, token_f1 -0.046, support_recall -0.12. Question<->edge cosine prunes away multi-hop bridge edges (the bridge entity isn't in the question), so it's the wrong ranking signal for multi-hop QA. Finding recorded in goldengraph/CLAUDE.md. If ball-pruning is revisited, the path-preserving filter (GOLDENGRAPH_HYBRID_FILTER=path) is the structurally-correct lever.

benzsevern:
Refuted by measurement (run 30035422835, env_ab none-vs-rerank@40): answer_match -0.04, token_f1 -0.046, support_recall -0.12. Question<->edge cosine prunes away multi-hop bridge edges (the bridge entity isn't named in the question), so it's the wrong ranking signal for multi-hop QA. Finding recorded in goldengraph/CLAUDE.md. If ball-pruning is revisited, GOLDENGRAPH_HYBRID_FILTER=path (path-preserving) is the structurally-correct lever.

PR #2136, opened 2026-07-25
docs(spec): numeric_diff parameterized native kernel design (P7)

benzsevern:
## What and why

Design spec (docs-only) for kernelizing `numeric_diff` — the **last non-model string scorer** still in `scorer_kernels_deferred` after date_diff/geo (Phase 4), array_intersect (#2134), and the Part-C ports. Closing it completes the "one Rust kernel per capability, single-sourced across Python-native / Rust / TS / WASM" arc.

## The core idea

`numeric_diff`'s band (`numeric_diff:abs:<eps>` / `:pct:<frac>`) can't ride the fixed-id `score_one(id, a, b)` contract — and unlike `array_intersect`'s discrete mode, it's continuous, so it can't be enumerated as ids. **But the band is field-constant, not per-pair** (`_parse_numeric_diff_spec` runs once per field), and the block kernel is already field-oriented. So the design:

- **mode** (`abs`/`pct`) → two discrete bucket ids (21/22),
- **band** → a new per-field `scorer_params` array threaded alongside `scorer_ids` in `score_block_pairs`,
- `score_one` itself is **never widened** — `numeric_diff` is a bucket-dispatch-only kernel, exactly like the name scorers (15/16) that already reach extra state inside `score_bucket_field`.

Wheel-safety is preserved by an **optional-trailing** `scorer_params` arg (never breaks the existing signature — the #688 hazard) plus a `numeric_diff_similarity` capability marker the host gates on (stale wheel → declines to the pure per-pair mirror).

## Scope boundary (called out explicitly)

This is a **weighted-bucket + cross-surface parity** change, **not** an FS scale change — `numeric_diff` keeps declining native on the Fellegi-Sunter path (numpy, like `soundex_match`), so `score_block_pairs_prob`, blocking, EM, and clustering are all untouched.

## Sequencing

P7a score-core kernel → P7b native+host+wheel (`python_only`) → P7c pure-TS+fixture → P7d wasm band-threading (`shared`). Each an additive, wheel/flag-gated PR.

## Testing / gates the plan will hit

score-core inline tests; `test_native_numeric_diff_parity.py` (native==pure per mode + block dispatch of 21/22 + capability-absent decline); `check_native_symbols`; `check_api_parity` (move deferred→shared, sort); TS/wasm parity fixtures; suite-matrix regen.

## Checklist

- [x] Spec committed to the tracked `docs/superpowers/specs/` dir
- [x] Docs-only — no code, no manifest change in this PR (those land per phase)

Follow-up: the implementation plan (`docs/superpowers/plans/…`) + the P7a–P7d PRs. This PR is the design artifact for review before implementation.

---
_Generated by [Claude Code](https://claude.ai/code/session_01DjdG9DKrLsgnxBetRSeR4x)_

### DIFF
--- a/docs/superpowers/specs/2026-07-25-numeric-diff-parameterized-kernel-design.md
+++ b/docs/superpowers/specs/2026-07-25-numeric-diff-parameterized-kernel-design.md
@@ -1,0 +1,284 @@
+# numeric_diff parameterized native kernel — design
+
+**Date:** 2026-07-25
+**Status:** design (approved direction; not yet implemented)
+**Related:**
+- `docs/superpowers/specs/2026-07-23-fs-domain-comparators-design.md` (the FS domain-comparator arc: date_diff / geo_haversine / array_intersect / numeric_diff)
+- `docs/superpowers/specs/2026-07-25-splink-domain-comparator-conversion-design.md` (the Splink-conversion arc P1–P6)
+- PRs #2130 / #2134 / #2135 ("one Rust kernel per capability, fed to every surface") — date_diff/geo (Phase 4), array_intersect (P5/P6), lsh/perceptual (Part C)
+
+## 1. Problem
+
+`numeric_diff` is the **last non-model string scorer** with no `-core` kernel on any
+surface. After #2134/#2135 the parity manifest (`parity/goldenmatch.yaml`)
+`scorer_kernels_deferred` holds exactly:
+
+```
+embedding        : n/a  -- model-backed
+record_embedding : n/a  -- model-backed
+numeric_diff     : deferred -- needs the parameterized score_one contract
+```
+
+So closing `numeric_diff` completes the "every non-model scorer is kernel-backed,
+single-sourced across Python-native / Rust / TS / WASM" thesis. It is the only
+deferral left that is *deferred* (will kernelize) rather than *n/a* (model-backed).
+
+### Why it wasn't done with the others
+
+Every kernelized scorer today is a `score_one` arm:
+
+```rust
+// packages/rust/extensions/score-core/src/lib.rs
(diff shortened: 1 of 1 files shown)


benzsevern:
Closing as superseded by #2138 ("numeric_diff scorer kernel single-sourced across all runtimes — Part D"), which a parallel session opened ~2 min after this spec and which *implements* the same capability, closing the arc (`scorer_kernels_deferred` down to the two model-backed embeds).

The implementations converged on the shared premise (the band/mode can't ride fixed-id `score_one(id,a,b)`; `numeric_diff` is the last non-model deferral) but diverged on **where the parameterized forms get kernelized**:

- **This spec:** thread a per-field `scorer_params` array through the batch `score_block_pairs` (optional-trailing arg), mode → two ids (21/22), band → the array — so parameterized forms are kernel-backed on the **vectorized batch** path.
- **#2138 (shipped):** the kernel parses the spec string itself — `numeric_diff_similarity(a, b, spec)` — with `score_one` arm **22** = the `pct:0.1` default for the emitted `scorer_kernels` surface, and a `_resolve_score_pair_callable` branch that kernel-backs the parameterized forms on the **per-pair** path. It deliberately does **not** widen the load-bearing batch kernel, arguing that's inert for the fixed-id-keyed parity surface and nil-value for trivial float arithmetic — and explicitly offers to add the per-field param array "if you'd rather have it."

#2138's call is defensible: numeric_diff is cheap float math, and large weighted-bucket numeric-diff blocks are a rare workload, so batch-vectorizing it buys little against the wheel-skew risk of touching `score_block_pairs`. The one thing worth a decision is exactly the divergence #2138 raised — whether to also kernelize the parameterized forms on the batch path. No code from this spec needs to land; the design rationale + alternatives here can inform that call.

---

Issue #1811, opened 2026-07-15
FS dedupe at 14M: late-stage OOM in driver-side pair/cluster accumulation (post-#1799)


benzsevern:
Follow-up to #1798. The 14M end-to-end confirmation on the 64GB runner (main @ df85c2c, #1799 in) moved the wall: the pre-scoring `build_blocks` OOM is gone and the run survives ~47 min, then hard-SIGKILLs inside `dedupe_df` during the scoring/clustering phase (kernel OOM-kill, not the 90-min timeout). 1M completes cleanly on the same path.

**Where the remaining peak lives** (audit finding F8, FS_PIPELINE_ANALYSIS_2026-07-15): the emitted-pair set and everything downstream of it is candidate-count-unbounded and driver-resident:

- `all_pairs` Python list of `(int, int, float)` tuples, extended by every pass (`core/pipeline.py:2246+`) -- at 14M with tight compound union blocking the above-threshold set can run to hundreds of millions of tuples (~100B each as Python objects) before clustering even starts;
- `matched_pairs` set + score_buckets' `frozen_exclude` frozenset (in-code noted as "the dominant Python-side accumulator" at 10M, `backends/score_buckets.py:704-706`);
- `build_cluster_frames`'s `pairs_list = list(pairs)` + three per-pair column comprehensions (`core/cluster.py:550,562-570`), `ClusterPairScores.from_frames(..., all_pairs)`, `dedup_pairs_max_score(all_pairs)` -- each re-walks or re-materializes the full pair list on the driver (#1792 residue; #1795 addressed part of the cluster side).

**Fix directions to evaluate** (roughly in order of leverage):
1. Keep emitted pairs COLUMNAR from the scorer outward: score_buckets already returns per-pass lists; accumulate into arrow arrays (id_a/id_b/score chunks) instead of tuple lists, and feed `build_clusters_arrow` / `dedup_pairs_arrow` (both kernels exist) without ever materializing `pairs_list`. The `_columnar_pairs_df` lane is a partial precedent.
2. Native exclude-set as the `matched_pairs`/`frozen_exclude` representation (ExcludeSet Arc already exists; #1808 threads it through FS scoring) so the Python frozenset copy disappears.
3. Optional pair spill (the duckdb accumulator lane exists for weighted; wire it for the bucket FS route) as the escape hatch for shapes whose true pair set genuinely exceeds RAM.

[...]

PR #2185, opened 2026-07-27
perf(fs): default-on columnar-cluster dedupe path B2c (#1811/#2006)

### LINKED ISSUE #2006: FS #1811 follow-up: keep scored_pairs off the driver on the columnar-cluster path (complete the peak-RSS win)
Follow-up to #1811 / PR #2004 (B2c, `GOLDENMATCH_FS_COLUMNAR_CLUSTER`).

## Context

PR #2004 eliminated the **scoring-phase** `all_pairs` Python `list[tuple]` on the eligible single-FS-matchkey in-memory bucket path — it threads the Arrow pair stream straight to `build_clusters_columnar`, so the list is never built during scoring → clustering (finding F8's primary accumulator). Verified in-sandbox: with the flag ON, clustering consumes the columnar DataFrame (`cluster_input=COLUMNAR_df_rows=1382400`) instead of a 1.38M-tuple list.

## The remaining gap

The **post-cluster** step still rebuilds the full list:

```
# core/pipeline.py (columnar branch)
scored_pairs = dedup_pairs_max_score(pairs_df_to_list(_columnar_pairs_df))
```

`pairs_df_to_list` materializes the entire pair set back into a Python `list[tuple]` for `DedupeResult.scored_pairs`. So measured peak RSS is ~unchanged at ≤1M (the list at that scale is a small fraction of peak), and at 14M this list is itself O(hundreds of millions) → it would still OOM. The `SHED scored_pairs` mechanism today is **fused-match-only** (`match_fused_capacity_mode`); the FS dedupe path always materializes.

## Proposed work

1. On the B2c path, dedup **columnar** via `dedup_pairs_max_score_arrow(_columnar_pairs_df)` (already exists, `core/pairs.py:190`) — no `list[tuple]` intermediate.
2. Above a size threshold, **shed** `scored_pairs` (empty list + a marker, mirroring the fused-match capacity mode) so the driver never holds the full pair set at scale. `DedupeResult.scored_pairs` is result-only past clustering (golden/clusters already built from `_columnar_pairs_df`), and `_finalize_review_pairs` (pipeline `~4545`) needs handling for the shed case.
3. **14M runner confirmation** (64GB `bench-fs-distributed`): OOM → survives with `GOLDENMATCH_FS_COLUMNAR_CLUSTER=1` + the shed. This is the gate that closes #1811.

## Note on determinism

The FS bucket pipeline is ~0.1%-nondeterministic run-to-run (measured: single-threaded + pinned EM still varies ~0.1% of pairs), so any parity gate here must be pair-set-overlap, not byte equality — same posture PR #2004 adopted.

benzsevern:
Closes #2006.

## What

Flips `GOLDENMATCH_FS_COLUMNAR_CLUSTER` from opt-in to **default-on** for the eligible single-probabilistic-matchkey in-memory bucket dedupe route. The Arrow pair stream threads straight to `build_clusters_columnar`, so the driver-resident `all_pairs` `list[tuple]` is never built during scoring → clustering — the late-stage OOM of #1811.

At 14M on tight-blocking data that list runs to hundreds of millions of tuples held on the driver. #1811's own confirmation (flag ON, 64GB runner): completes at **~30GB peak RSS** where the list path OOM-killed at 24 min, **F1 held (0.9422)**.

## Why it was opt-in — the two gaps I closed to make it a safe default

1. **Review band was dropped.** The columnar branch left `review_pairs` empty. It now splits the Arrow table by `link_threshold` exactly the way B2b (`_split_pair_stream`) does: the linked set stays columnar for clustering, the bounded review band (already floored at the review cut by `score_buckets_arrow`) materializes to `review_pairs`. Without this, defaulting on would silently remove the review queue from **every** eligible FS dedupe, since the calibrated review cut produces one by default. Inherits B2b's parity-gated review contract.
2. **Bench-dump broke.** B2c's eligibility gate didn't exclude bench-dump (B2b's `_fs_arrow_stream_eligible` does). The Arrow emit doesn't expose the per-block candidate/emitted accounting the bench-dump path needs → now excluded, falls back to the per-block loop.

## Safety

- Ineligible configs (multi-matchkey, `across_files_only`, semantic blocking, bench-dump, non-bucket scale backends) **fall back to the list path unchanged**.
- Clusters are not byte-identical to the list path (the FS bucket pipeline is ~0.1%-nondeterministic run-to-run regardless), so the flip moves the default output within that band only.
- `GOLDENMATCH_FS_COLUMNAR_CLUSTER=0` forces the legacy list path (the byte-parity oracle).
- `scored_pairs` above `GOLDENMATCH_FS_SCORED_PAIRS_MAX` (default 50M) is shed with a `scored_pairs_shed=True` marker, never silently.

## Tests

- New guards in `test_fs_columnar_cluster_1811.py`: default-unset → columnar path; review-band survives the columnar path. Explicit-off + clear-margin parity tests retained.
- Full FS/probabilistic/bucket/cluster/columnar/dedupe subset green (632 passed) with the default flipped, minus two confirmed pre-existing env-only failures (datafusion not installed locally; a weighted `xfail-strict` that XPASSes only in pure-Python).

## Scale validation (reviewer note)

The columnar linked path at scale is the exact one #1811 validated at 14M. The new bit is the review-band carry (bounded list materialization, unit-tested, mirrors B2b) — worth a `bench-probabilistic` panel run to confirm F1 + peak RSS hold with the default on.


benzsevern:
Not landable as-is — the default-on columnar-cluster path fails golden parity vs the classic lane. `test_frame_lane_probabilistic_e2e_parity` (python_goldenmatch (3)) shows the columnar lane computing different `__golden_confidence__` values:

```
frame lane:  confidence 0.9, 0.9, 0.8
classic:     confidence 0.8, 0.8, 0.8
```

Same clusters + members, but the columnar cluster-confidence differs from the classic path (which the flip makes the default, so this is a user-visible output change, not a perf-only one). A default flip must be behaviorally neutral — the columnar `build_clusters_columnar` confidence needs to reconcile with the classic per-cluster confidence before this can go default-on. Moving back to draft until that parity holds (or the flip is scoped to leave confidence on the classic path).

benzsevern:
Superseded by the Arrow-native B2c PR (frames-out clustering + default-on, no polars) — that one keeps the pair stream as a pa.Table end-to-end instead of the `pl.from_arrow` conversion this PR carried, and routes clustering through the frames-out path the measure-first profile pointed to. Closing in favor of it.

PR #2211, opened 2026-07-27
feat(identity): entity<->entity relationship edges from shared attributes (#2210)

### LINKED ISSUE #2210: identity: emit entity<->entity relationship edges from shared attributes (semantic-graph layer)
Identity resolution collapses records that are the SAME entity. The same blocking data also surfaces DIFFERENT entities that share a NON-identity attribute -- two prescribers on one clinic phone, two people at one address, colleagues at one email domain. That is a relationship, not a merge, and today it is discarded (used only negatively, as `conflicts_with` / the fan-out veto).

So there is no path from the structured identity graph (records -> entities) to a semantic relationship graph (entity <-> entity). GoldenGraph builds relationship graphs from *text* extraction; nothing turns the resolved structured entities into relationship edges.

This is a missed, cheap win: the identity graph already produces the hard part of a good relationship graph -- clean, deduplicated, stable entity nodes -- and the resolver already computes the shared-attribute signal. We should be able to emit relationship edges keyed on the stable `entity_id`s, in the same resolve pass, maintained incrementally.

## Proposal
A relationship layer on the identity store:
- `IdentityConfig.relationships: list[RelationshipRule]` -- opt-in rules `{field, kind, min_entities, max_fanout}` naming a non-identity payload field and the edge label to emit.
- A new `identity_relationships` table (entity-level, distinct from the record-level `evidence_edges`), idempotent across runs (PK omits `run_name`), fan-out-capped (skip hub values like a switchboard line).
- Emission as a post-resolve pass (cross-entity, so it can't run in the per-cluster loop); one grouped SQL query per rule.
- Read API: `get_relationships(entity_id)`, `count_relationships()`.

Keeps the two edge semantics cleanly separated: `same_as` collapses records INTO an entity; relationships connect DISTINCT entities. Turns the identity graph into a semantic graph maintained in lockstep, with no separate ER pass.

benzsevern:
Closes #2210.

## The gap
Identity resolution collapses records that are the SAME entity. The same blocking data also surfaces DIFFERENT entities that share a NON-identity attribute -- two prescribers on one clinic phone, two people at one address -- which is a relationship, not a merge, and today it's discarded (used only negatively, as `conflicts_with` / the fan-out veto). So there's no path from the structured identity graph (records -> entities) to a semantic relationship graph (entity <-> entity), even though the identity graph already produces the hard part -- clean, stable entity nodes -- and the resolver already computes the signal.

## What this adds
A relationship layer, opt-in, maintained in the same resolve pass:

- **config:** `RelationshipRule {field, kind, min_entities, max_fanout}` and `IdentityConfig.relationships`. Empty (default) leaves identity resolution byte-identical.
- **store:** a new `identity_relationships` table -- entity-level, distinct from the record-level `evidence_edges` (sqlite + postgres + a `v6 -> v7` migration). Idempotent across runs (the PK omits `run_name`, so a warm re-resolve doesn't duplicate). `relationship_groups` does the cardinality gate in one grouped SQL query per rule; `add_relationships` canonicalizes endpoints (`a < b`) and de-dupes; `get_relationships` / `count_relationships` are the read side. Field names are validated before they're interpolated into the JSON path.
- **resolve:** a post-resolve pass (cross-entity, so it can't run inside the per-cluster loop), threaded through `ResolutionBatch`, wrapped so it never fails resolution. New `ResolveSummary.relationships_added`. Auto-runs from the pipeline when `config.identity.relationships` is set.

The two edge semantics stay separate: `same_as` collapses records INTO an entity; relationships connect DISTINCT entities. Hub values (a switchboard line shared by hundreds of entities) are skipped via `max_fanout` -- they aren't pairwise relationships.

## Why
The identity graph is the right foundation for a semantic graph -- a relationship graph is only as good as its nodes, and ER is what makes the nodes clean. This turns the resolved entities into a relationship graph keyed on the stable `entity_id`s, updated incrementally alongside absorb/merge/split, with no separate ER pass. (GoldenGraph builds relationship graphs from *text*; this is the structured-record complement.)

## Tests
`test_relationships.py`: shared attribute relates distinct entities (+ symmetry, unrelated entities excluded), idempotent re-resolve, same-entity records don't self-relate, `max_fanout` skips hubs, multiple rules, no-rules no-op. Existing identity resolve/incremental/profile/batch-parity suites and the config-matrix/codemap/api-surface suites pass; docs regenerated.

## Follow-ups (not in this PR)
- An export helper to GoldenGraph's node/edge (`subj`/`predicate`/`obj`) shape so the relationship graph feeds its path-retrieval / answer layer.
- Distributed (Ray/Postgres) path currently emits relationships only on the in-memory path; wiring the distributed resolve is a follow-up.

### DIFF
--- a/docs-site/goldenmatch/config-matrix.mdx
+++ b/docs-site/goldenmatch/config-matrix.mdx
@@ -373,7 +373,7 @@ _Identity Graph configuration._
 | `source_pk_column` | str \| None | `None` | Column supplying each record's source primary key for stable record ids; a payload hash is used when unset. |
 | `emit_singletons` | bool | `True` | Whether single-record clusters also get a durable entity id. |
 | `weak_confidence_threshold` | float | `0.6` | Cluster confidence below which the bottleneck pair is flagged as a conflict for steward review; 0 disables it. |
-| `relationships` | list[RelationshipRule] | `[]` | [`RelationshipRule`](#relationshiprule) -- Rules deriving entity&lt;->entity relationship edges from shared non-identity attributes; empty leaves identity resolution unchanged. |
+| `relationships` | list[RelationshipRule] | `[]` | [`RelationshipRule`](#relationshiprule) -- Rules deriving entity-to-entity relationship edges from shared non-identity attributes; empty leaves identity resolution unchanged. |
 | `stitching` | ChannelStitchConfig \| None | `None` | [`ChannelStitchConfig`](#channelstitchconfig) -- Cross-device/channel stitching configuration; None leaves identity resolution unchanged. |
 | `survivorship` | SurvivorshipConfig \| None | `None` | [`SurvivorshipConfig`](#survivorshipconfig) -- Identity golden-record survivorship configuration; None keeps the default flat golden record. |
 | `stabilization` | StabilizationConfig \| None | `None` | [`StabilizationConfig`](#stabilizationconfig) -- Cross-run entity stabilization configuration; None runs no stabilize pass. |
@@ -542,7 +542,7 @@ _Learning Memory learning parameters._
 | `weights_min_corrections` | int | `50` | Minimum stored corrections before learned field weights are tuned from them. |
 
 ### `RelationshipRule`
-_One rule for deriving entity<->entity relationship edges from a shared,_
+_One rule for deriving entity-to-entity relationship edges from a shared,_
 
 | Field | Type | Default | Choices / notes |
 |---|---|---|---|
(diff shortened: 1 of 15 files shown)


benzsevern:
Folded in the first listed follow-up: `to_graph_batch(store)` exports the relationship graph as GoldenGraph's `StoreBatch` `{entities, edges}` shape, with each entity's stable `entity_id` as its `record_keys` reconciliation key, so it appends straight into a `PyStore` and feeds the path-retrieval/answer layer. Pure data (no goldengraph import); round-trip test skips when goldengraph isn't installed. The distributed-path follow-up still stands.

PR #2213, opened 2026-07-28
perf(score-core): allocation-free ASCII fast path for jaro-winkler (peq HashMap -> array)

benzsevern:
## What the flamegraph found
A py-spy/perf flamegraph of a 5M FS dedupe showed `bucket_score` is **64%** of the run, but the actual jaro-winkler *math* is only **~7.5%**. The vendored jaro's `peq_map` builds a `HashMap<&char, Vec<u64>>` **per comparison** — **~14.6% hashbrown/siphash** + a big share of the **13.4% malloc/free** (the `vec![0u64; nw]` scratch). We were spending ~2× more on hashing/allocating the pattern-match vector than on comparing the strings.

## The fix
Single-block **ASCII fast path**: when the pattern fits one 64-bit block (`len1 <= 64`) and every char in both strings is ASCII (`< 128`) — the ~universal case for names — use a stack `[u64; 128]` peq indexed by byte + single-`u64` bitvectors. **Zero hashing, zero heap allocation.** Byte-identical to the generic HashMap path; Unicode / long patterns fall back unchanged.

## Parity + perf
- New `ascii_fast_path_parity_vs_rapidfuzz` (60k ASCII pairs, bit-exact vs the rapidfuzz oracle); existing mixed-alphabet + limb-boundary parity tests still pass (**38 total**).
- Local A/B on the jaro-heavy scorer mix: **1.28s → 1.05s (−18%)**, identical checksum. clippy clean.
- Scale flamegraph re-run in flight to confirm `bucket_score`'s HashMap+alloc share drops.

### DIFF
--- a/packages/rust/extensions/score-core/src/strsim.rs
+++ b/packages/rust/extensions/score-core/src/strsim.rs
@@ -179,6 +179,21 @@ fn jaro_similarity(s1: &[char], s2: &[char]) -> f64 {
     // rapidfuzz bound = max(len1, len2) / 2 - 1 (both >= 1 here, so >= 0).
     let bound = std::cmp::max(len1, len2) / 2 - 1;
 
+    // ASCII single-block fast path (the ~universal case for names). When the
+    // PATTERN fits one 64-bit block AND every char in BOTH strings is ASCII, the
+    // per-comparison `HashMap<&char, Vec<u64>>` peq + the `vec![0u64; nw]`
+    // scratch allocations are pure overhead: profiling (flamegraph) put peq_map's
+    // hashing + allocation at ~2x the actual jaro math. Use a stack `[u64; 128]`
+    // array peq indexed by byte and single-u64 bitvectors -- zero hashing, zero
+    // heap allocation. Byte-identical to the generic path (same algorithm, same
+    // bit ops, same order); parity is asserted in the tests below.
+    if len1 <= 64
+        && s1.iter().all(|c| (*c as u32) < 128)
+        && s2.iter().all(|c| (*c as u32) < 128)
+    {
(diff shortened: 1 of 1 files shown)


benzsevern:
## Scale flamegraph confirms it (2M, same profiler, before → after)

| category (SELF) | before | after |
|---|---|---|
| **Rust HashMap/hash** (peq map) | 14.6% | **1.8%** |
| **allocation** (malloc/free) | 13.4% | **5.0%** |
| jaro/strsim (the math) | 7.5% | 6.6% |
| **`bucket_score` total** | **64.4%** | **27.6%** |

Profiled run: **111,804 → 63,606 samples** (~43% fewer at the same 250Hz), so `bucket_score`'s absolute cost dropped **~4×**. The peq HashMap is eliminated (14.6%→1.8%), allocation more than halved. Byte-identical output (parity-proven). Runs 30304221167 (before) / 30315935427 (after).

Next frontier: the post-scoring `to_list` marshaling (cluster→dict rebuild + golden) is now the top category (24.8%).

benzsevern:
Superseded by #2212, which landed the same peq-HashMap→stack-array optimization on main in parallel — and more comprehensively: byte tier ([u64;128] ASCII) + char tier ([u64;256] Latin-1) across jaro, levenshtein, AND lcs, vs this PR's ASCII-only jaro. Closing as redundant.

The flamegraph work stands, though: it independently identified the peq_map HashMap as the hidden ~15% cost, and the re-run measured the fix's impact at scale — bucket_score 64%→28%, Rust HashMap 14.6%→1.8%, allocation 13.4%→5.0% (run 30315935427 vs 30304221167). That confirms #2212 delivers the ~4x win on the dominant scoring stage.

Issue #2216, opened 2026-07-28
identity: Postgres bulk COPY runs inside pipeline mode -> write silently fails at scale (0 nodes, reported success)


benzsevern:
The Postgres identity write silently fails at scale and reports success.

## Symptom
A 14M `resolve_clusters` on `backend=postgres` returns "success" but commits **0 nodes** -- `identity_nodes` is empty, `identity_summary` is null -- so any downstream read (steward views, relationship graph) sees an empty store. Node counts were never verified, so it went unnoticed and skewed a PG-vs-SQLite benchmark (PG looked "fast" because it wasn't actually writing).

## Root cause
Two bugs compound:

1. **COPY runs inside pipeline mode.** `resolve_clusters` wraps its write loop in `store.write_pipeline()` (psycopg pipeline mode) and flushes bulk-COPY accumulators from a MID-LOOP `_flush_bulk()` (`resolve.py`, the `len(bulk_record_rows) >= bulk_flush_threshold` branch) -- i.e. *inside* the pipeline. psycopg forbids COPY in pipeline mode, so `bulk_upsert_*` raises `COPY cannot be used in pipeline mode`. The store's own contract (`write_pipeline` docstring) says COPY must be flushed OUTSIDE the pipeline; the mid-loop flush violates it. The FINAL flush is outside the pipeline, so a small resolve (only the final flush) never triggers it -- which is why the 20-row unit test passed and only 14M (mid-loop flushes) failed.

2. **The failure is swallowed.** `pipeline.py::_resolve_identity` catches the exception, `logger.warning`s it, and `return None` -- so a REQUESTED (`config.identity.enabled`) identity resolution that fails masquerades as success.

## Fix
- `IdentityStore.bulk_copy_barrier()`: suspends the active pipeline for the duration of a bulk COPY, then reopens it inside the same `bulk_writes` transaction. `resolve_clusters` wraps both `_flush_bulk()` sites (no-op off Postgres).
- `_resolve_identity`: log at ERROR and re-raise, so a failed write is a failed run, not silent data loss.
- Regression test forcing a mid-loop flush with the pipeline on.

PR #2224, opened 2026-07-28
feat(identity): entity relationship graph + Postgres 14M write-path (#2210, #2216)

benzsevern:
## What and why

Merge the `relgraph-14m-validation` work to main. Two themes: the entity-to-entity
relationship graph feature (#2210) and the Postgres 14M write-path work (the
#2216 COPY-in-pipeline fix plus an opt-in initial-load fast path).

## Changes

**Relationship graph (#2210)**
- Entity-to-entity relationship edges derived from shared attributes (distinct from record-level `evidence_edges`), persisted in `identity_relationships`.
- Export the relationship graph to GoldenGraph's PyStore shape.
- `relationships_added` surfaced on the Python and TypeScript `ResolveSummary`.
- MDX-safe relationship descriptions and a refreshed resolve-clusters parity fixture.

**Postgres 14M write path**
- #2216: bulk COPY must run outside psycopg pipeline mode. `bulk_copy_barrier` suspends the active pipeline around each bulk-COPY flush and reopens it inside the same transaction; a failed identity write now raises instead of being swallowed (a 14M resolve was committing 0 rows while reporting success).
- Opt-in initial-load fast path (`IdentityStore.initial_load_writes`): for a from-empty build it drops the secondary indexes, optionally loads UNLOGGED, routes the `bulk_*` methods to a direct COPY into the real tables (no temp staging, no `INSERT..SELECT`, no `ON CONFLICT`), then rebuilds indexes + `SET LOGGED` + `ANALYZE`. Gated to Postgres and empty tables (a transparent no-op otherwise), default off behind `GOLDENMATCH_IDENTITY_INITIAL_LOAD` (+ `_UNLOGGED`).

## Testing

- The initial-load fast path was verified byte-identical to the normal temp+ON CONFLICT path against a live Postgres: `normal == fast == fast+unlogged` across all four identity tables, the empty-tables gate no-ops on a non-empty store, and indexes are rebuilt + tables back to LOGGED after. (The local dev box cannot run the ephemeral-Postgres pytest fixture, so validation was against a remote PG rather than a committed pg test; a committed test in the `python_postgres` lane is a sensible follow-up.)
- The #2216 fix ships with its own regression test forcing a mid-loop bulk flush under pipeline mode.
- Full CI dispatched on the branch: run 30372123076, including the `python_postgres` lane.

## Checklist

- [x] Tests added/updated and passing for the changed package (relationship graph + #2216 fix; initial-load fast-path test is a noted follow-up)
- [ ] `pre-commit run --all-files` is clean (not run on the dev box; relies on CI)
- [ ] Package `CHANGELOG.md` updated if behavior changed
- [x] Docs / `CLAUDE.md` updated if a workflow or gotcha changed (initial-load path documented in docstrings + env flags)

https://claude.ai/code/session_013RtzHiSR4kYMuZ2SUKippz


benzsevern:
Superseded by https://github.com/benseverndev-oss/goldenmatch/pull/2228. main already has #2210 and #2216 (merged as #2211/#2217), so this wip->main merge only conflicted on already-landed history. Reopened as a clean fast-path-only PR cut off main.

PR #2272, opened 2026-07-30
perf(fs-ooc): bound the streaming FS back-half (free frame + spill DuckDB + batch golden)

benzsevern:
## What and why

The out-of-core streaming FS dedupe path (`GOLDENMATCH_FS_OUT_OF_CORE=1`, the >=40M scale route) spills the prepared frame to DuckDB but then ran the memory-heavy back-half (cluster + output) with two O(N) / full-width resident structures that would re-OOM at 50M, defeating the spill design. This bounds all of them so a single box can clear 50M where the in-memory path OOMs (~82 GB). Output is byte-identical; the whole route stays opt-in and default-OFF.

## Changes

- **Free the prepared frame at the load boundary, not through the back-half.** Every downstream stage reads the DuckDB `prep` table, never the frame, so the frame is dead weight after `_load_frame_batched`. It is now routed through a single-element HOLDER (`_unwrap_frame_holder`) that `score_fs_out_of_core` nulls right after the load; `_run_dedupe_pipeline` rebinds its own frame locals (`collected_df` / the `.lazy()` view / the pre-matchkey copy) to `None` so the holder is the only strong reference. Weakref-verified: the ~1x frame (~5-17 GB at 50M) is freed BEFORE clustering + output instead of pinned resident through it.
- **Bound DuckDB's buffer so `prep` spills to disk** (`_configure_ooc_duckdb`: `memory_limit` ~55% RAM + `temp_directory`, env `GOLDENMATCH_FS_OOC_DUCKDB_MEMORY`). Without a cap, DuckDB's default ~80%-of-RAM buffer holds the whole spilled `prep` table in RAM (an equivalent copy of the frame), so freeing the Python frame alone is masked and 50M still OOMs.
- **Build golden in cluster-BATCHES streamed via `ParquetWriter`** (`_fs_ooc_golden_batch_clusters`, default 500K clusters/batch, env `GOLDENMATCH_FS_OOC_GOLDEN_BATCH`). The golden subset (non-oversized multi-member rows) can be MOST of N at full width; materialising it whole was an O(N) spike (measured 7.8 GB at 400K rows x 40 cols, batched 0.74 GB, -10.6x, identical golden_count) - the same peak the in-memory #334 golden fix removed. GOTCHA fixed: DuckDB `/` is float division; the batch bucketing needs `//`.
- **`_cluster_arrow_native` is now polars-free**: arrow-native `dedup_pairs_max_score_arrow_table` + pa.Table straight into `build_clusters_arrow_native`, dropping the `pl.from_arrow` pair copy (aligned with the polars-eviction direction).
- Adds `scripts/bench_fs_streaming_peak.py` (VmRSS-sampled scale-peak bench) for the CI-proof-50M follow-up; updates the module docstring, `CLAUDE.md` note, and regenerates `docs/agent-codemap.json`.

## Testing

- `pytest tests/test_fs_out_of_core.py` - 18 passed (15 existing parity + 3 new: golden multi-batch parity/count, holder frees the frame via weakref, bare-frame path still supported).
- `pytest tests/test_pipeline.py tests/test_probabilistic.py` - 141 passed (default path unaffected).
- `ruff check` clean on the changed files; `agent_codemap.py --check` OK.
- Measured on this box (`bench_fs_streaming_peak.py` + an isolated golden-stage probe): frame weakref dies before the peak; golden-stage resident peak 7842 MB -> 741 MB batched at 200K clusters x 40 cols, byte-identical counts.

## Checklist

- [x] `ruff check` clean on changed files (no secrets)
- [ ] Package `CHANGELOG.md` updated if behavior changed (opt-in default-OFF route; no user-visible default change)

---
_Generated by [Claude Code](https://claude.ai/code/session_01U1VZjzNuoMPhHbN3HuiGuA)_


benzsevern:
Re: the `github-code-quality` bot flags on `score_frame` / `_em_src` (pipeline.py) and `df` (test) — these are **intentional reference-drops**, not dead stores: they release the last driver-side strong reference to the prepared frame so the holder can free the ~1x frame at the DuckDB-load boundary (the whole point of this PR; asserted by `test_streaming_frees_prepared_frame_via_holder` via a weakref). Removing them (the bot's suggestion — note its two pipeline comments contradict each other, each saying keep the *other*) would re-pin the frame through the back-half and fail that test.

Resolved by converting `x = None` → `del x`, which expresses the lifetime-drop intent and clears the "unused local" flag while preserving the behavior. Also regenerated `config-matrix.mdx` + `agent-manifest.json` for the two new `GOLDENMATCH_FS_OOC_*` env knobs (the `config_matrix` gate).

---

benzsevern:
Closing this draft. Its central premise — freeing the prepared frame between FS stages to cut peak RSS — has been superseded by the FS out-of-core work already on `main`: the bucketed streaming path (`GOLDENMATCH_FS_OUT_OF_CORE`), the EM agg-blocks / block-slim projections, and the jemalloc page-retention finding (`_RJEM_MALLOC_CONF="dirty_decay_ms:1000,muzzy_decay_ms:0"`). Those established that the apparent ~3.3 GB 1M peak is ~2.2 GB *live* working set plus ~1.1 GB allocator retention — i.e. the frame wasn't being held by live references, so "free the frame" was addressing a phantom. The remaining live-floor work is tracked by the frame-residency spec rather than this diff. Branch and commits remain if any individual mechanism here is worth cherry-picking into a fresh PR.

---

PR #2287, opened 2026-07-30
fix(bench-er): measure F1 through a RED refusal in the cost-aware OFF-vs-ON lane

benzsevern:
## What and why

Follow-up **stacked on #2278** (base = its `claude/product-blocking-config-34eio3` branch, so this diff is just the fix — merging it lands the fix on #2278).

The `cost-aware-offvson` gate job that #2278 adds — the lane meant to prove the `GOLDENMATCH_BLOCKING_COST_AWARE` default flip — **crashed on its very first datapoint** ([run 30578449096](https://github.com/benseverndev-oss/goldenmatch/actions/runs/30578449096)). On the **person** shape at 200K, `run_goldenmatch.py --mode zeroconfig` **refuses**: the controller commits a RED config (blocking sub-profile, `stop_reason: BUDGET_TIME`) on identifier-poor person data — *exactly the #2021 case the flag targets* — so it wrote `status=refused` with **no prediction parquet**, and the downstream `evaluate.py` died with `IO Error: No files found that match "ca_out/person_flag0_pred.parquet"`.

A refusal here is **expected, not a bench failure**, and the whole point of the gate is to compare the committed config's **actual F1** OFF vs ON — which a bare refusal can't provide.

## Changes

- **`scripts/bench_er_headtohead/run_goldenmatch.py`** — new `--measure-refused` flag (default OFF). When set and the zeroconfig controller raises `ControllerNotConfidentError`, re-config + force-run the **same committed config** with `allow_red_config=True` (+ `_skip_finalize=True`) so a prediction IS produced and F1 is measurable, flagging `refused=true` for context. This mirrors the QIS-gate `measure_rungs` idiom already documented in `CLAUDE.md`. Pass/fail stays on the **measured F1**, not the refuse verdict (RED is a confidence flag, not a wrong answer).
- **`.github/workflows/bench-er-headtohead.yml`** — the `cost-aware-offvson` step passes `--measure-refused`.
- Default OFF keeps the main head-to-head matrix's `gm_zeroconfig` lane refuse-and-record semantics **untouched** — only the cost-aware step opts in.

## Testing

- `python -m py_compile scripts/bench_er_headtohead/run_goldenmatch.py` — clean; workflow YAML parses.
- The refuse path only triggers at ≥100K rows (a ~5-min controller run each), so it's validated by re-dispatching the `cost-aware-offvson` lane itself rather than locally. Re-run pending.

## Checklist

- [x] Docs / `CLAUDE.md` gotcha already documents the force-measure idiom (QIS-gate `measure_rungs`)
- [ ] Bench lane re-run green (pending 64 GB dispatch)

https://claude.ai/code/session_01SGfNn3WiFCSJ8UXNsMwqM9

---
_Generated by [Claude Code](https://claude.ai/code/session_01SGfNn3WiFCSJ8UXNsMwqM9)_

### DIFF
--- a/.github/workflows/bench-er-headtohead.yml
+++ b/.github/workflows/bench-er-headtohead.yml
@@ -276,6 +276,7 @@ jobs:
                 scripts/bench_er_headtohead/run_goldenmatch.py \
                 --input "ca_out/${SHAPE}.parquet" --rows "$COST_AWARE_SCALE" \
                 --mode zeroconfig --shape "$SHAPE" --allow-pure-python \
+                --measure-refused \
                 --out "ca_out/${SHAPE}_flag${FLAG}.json" \
                 --pred-out "ca_out/${SHAPE}_flag${FLAG}_pred.parquet"
               uv run python scripts/bench_er_headtohead/evaluate.py \
--- a/scripts/bench_er_headtohead/run_goldenmatch.py
+++ b/scripts/bench_er_headtohead/run_goldenmatch.py
@@ -88,6 +88,15 @@ def main() -> None:
                     help="probabilistic mode only: rewrite any FS field scorer NOT in "
                          "the native FS kernel set to jaro_winkler, so both FS lanes "
                          "engage the Rust kernel and match Splink's JaroWinkler model")
+    ap.add_argument("--measure-refused", action="store_true", default=False,
+                    help="zeroconfig mode only: when the controller REFUSES a RED "
+                         "config (>=100K identifier-poor input), MEASURE the committed "
+                         "config's F1 by force-running it with allow_red_config=True "
+                         "(the QIS-gate idiom) and flag refused=true, instead of "
+                         "recording status=refused with no prediction. Required by the "
+                         "cost-aware OFF-vs-ON gate, whose person shape refuses by "
+                         "design (the #2021 case). Default OFF preserves the head-to-"
+                         "head matrix's refuse-and-record semantics.")
     args = ap.parse_args()
 
     os.environ["GOLDENMATCH_NATIVE"] = "1" if args.require_native else "auto"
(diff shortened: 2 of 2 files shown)


benzsevern:
## Validation: lane runs green now ✅ — but it can't demonstrate the person WIN on the current fixture ⚠️

Re-dispatched the `cost-aware-offvson` lane on this branch: **[run 30579810050](https://github.com/benseverndev-oss/goldenmatch/actions/runs/30579810050) → success**. The `--measure-refused` force-run works: the person shape refuses (RED, `BUDGET_TIME`/`OSCILLATING`) and is now measured via `allow_red_config` instead of crashing `evaluate.py`.

Verdict from the job summary:

| shape | F1 OFF | F1 ON | ΔF1 | pairs OFF | pairs ON | pair Δ | verdict |
|---|---|---|---|---|---|---|---|
| person | 0.8585 | 0.8585 | +0.0000 | 287,371 | 287,371 | +0.0% | ok |
| biblio | 0.7995 | 0.7995 | +0.0000 | 59,181 | 59,181 | +0.0% | ok |

Gate criteria met: **no F1 regression on either shape, and biblio is domain-exempt (OFF == ON)**.

### The caveat the numbers expose

On **person**, OFF == ON on candidate pairs too (**+0.0%**) — the flag is a **no-op on this fixture**, so the lane does **not** demonstrate the pair-collapse this feature exists for. Root cause: `generate_fixture.py`'s person schema is `record_id, first_name, surname, **dob**, postcode, city`. `dob` is a full date (~25K distinct), not a low-cardinality `birth_year` (~65 distinct), and it isn't even an exact-eligible survivor here (the log shows *"All 3 exact-eligible columns were excluded → fuzzy-only"*, so `safe_exact` is empty and the cost-aware demotion branch is never entered). The #2021 pathology — a `year` col_type as the *sole* exact primary — simply isn't present in this fixture. It was measured on the **QIS corrupted-realistic** shape (the 76.9M → 0.8M collapse in #2278's table), which is a different generator.

**Net:** this lane is a valid *non-regression + biblio-exemption* gate, but it is **not** proof of the person collapse. To actually gate the default-on flip on the win, the lane needs a fixture that carries the pathology (a `birth_year`-dominant person variant), or it should run OFF-vs-ON on the QIS corrupted-realistic shape where the collapse is real. That's a design call on #2278's gate — flagged for @benzsevern.

---

PR #2390, opened 2026-08-03
perf(fs): Lever 1 narrow-base + per-shard xform recompute (A/B, default OFF)

benzsevern:
## What

Frame-residency **Lever 1** (spec `docs/superpowers/specs/2026-08-03-fs-frame-residency-disk-spill-design.md`), gated behind `GOLDENMATCH_FS_XFORM_PER_SHARD=1`, **default OFF**.

The bucketed FS out-of-core route's peak is ~1x the resident `base` frame (~0.8 GB/M). `base` carries ~5 `__xform_<sig>__` columns (~40% of its width) that `precompute_matchkey_transforms` materializes on the whole frame. Lever 1 skips that materialization (narrower resident base -> lower peak -> higher single-box max scale, ~75M -> ~120M on 64 GB) and recomputes each transform **per bucket shard** (~1/64 of the frame) at score time.

## Byte-identical by construction

- **EM** applies `apply_transforms` on raw source values and never reads `__xform_` (probabilistic.py:617-957) -> unaffected by dropping the wide block; the narrow frame keeps derived-NE columns EM does read.
- **Per-shard recompute** re-adds exactly the `__xform_<sig>__` columns `_resolve_fast_path` needs, so the native fast path engages identically -> identical pairs.
- **Output** already excludes `__xform_` (record_cols filter) -> identical rows.

Parity gate: `test_bucketed_narrow_base_lever1_parity` asserts arm A (wide) == arm B (narrow) on unique/dupes/golden + pairs, and `base_ncols` strictly narrower with the flag on.

## The A/B (measurement gate)

Same-box back-to-back on the 64 GB bench runner (lesson: cross-runner wall is noise): [run 30824543298](https://github.com/benseverndev-oss/goldenmatch/actions/runs/30824543298), `modes="bucketed bucketed_narrow"` x `scales="4000000 8000000"`. The spec gate: arm B must (a) lower peak ~= the `__xform_` width fraction and (b) not regress wall by more than a few %.

**Do not merge on green CI alone** - the merge decision is gated on that A/B verdict. CI here validates the byte-identical parity gate in the native-built env.

### DIFF
--- a/packages/python/goldenmatch/goldenmatch/backends/fs_out_of_core.py
+++ b/packages/python/goldenmatch/goldenmatch/backends/fs_out_of_core.py
@@ -105,6 +105,22 @@ def _fs_out_of_core_legacy_flag() -> bool:
     )
 
 
+def _fs_xform_per_shard_enabled() -> bool:
+    """Lever 1 (frame-residency): ``GOLDENMATCH_FS_XFORM_PER_SHARD=1`` makes the
+    ``bucketed`` FS route skip the whole-frame ``__xform_`` materialization (so the
+    resident ``base`` is ~40% narrower ⇒ lower peak ⇒ higher single-box max scale)
+    and recompute each matchkey transform PER BUCKET SHARD (~1/64 of the frame) at
+    score time instead. Default OFF (byte-identical to the wide-base path: the
+    native FS fast path prefers the precomputed ``__xform_`` column, a −82%
+    ``bucket_score`` win the wide base keeps). Measurement-gated — trades that
+    per-call win for a lower peak; the A/B must show the peak win outweighs the
+    per-shard recompute. Spec:
+    ``docs/superpowers/specs/2026-08-03-fs-frame-residency-disk-spill-design.md``."""
+    return os.environ.get("GOLDENMATCH_FS_XFORM_PER_SHARD", "").strip().lower() in (
+        "1", "true", "yes", "on",
+    )
+
+
 def _fs_spill_force_shard() -> bool:
     """``GOLDENMATCH_FS_SPILL_FORCE_SHARD=1`` makes ``run_fs_dedupe_spill`` SKIP the
     fused-kernel short-circuit and always take the edge-shard spill path
@@ -1868,6 +1884,15 @@ def run_fs_dedupe_bucketed(
         _fw = _tf(_fw.native.collect())
     base = _fw.to_arrow()
 
+    # Lever 1 (frame-residency): when the caller skipped the whole-frame
+    # __xform_ materialization (GOLDENMATCH_FS_XFORM_PER_SHARD; pipeline passes a
+    # narrow base), the base carries NO __xform_ columns. Detect that and
+    # recompute the transforms PER SHARD (~1/64 of the frame) below so the native
+    # fast path still engages — byte-identical pairs, ~40% narrower resident base.
+    _narrow = _fs_xform_per_shard_enabled() and not any(
+        c.startswith("__xform_") for c in base.column_names
+    )
+
     passes = _bucketed_passes(blocking_config)
     shard_dir = _tempfile.mkdtemp(prefix="gm_fs_bucketed_")
     try:
@@ -1907,6 +1932,16 @@ def run_fs_dedupe_bucketed(
             single_pass = _BlockingConfig(strategy="static", keys=[passes[pi]])
             with pa.memory_map(sp, "r") as src:
                 bucket_tbl = pa.ipc.open_file(src).read_all()
+            if _narrow and bucket_tbl.num_rows >= 2:
+                # Lever 1: recompute __xform_ on this shard so score_buckets_arrow's
+                # native fast path engages exactly as on a wide base. The shard is
+                # ~1/64 of the frame, so the recompute is bounded and transient.
+                from goldenmatch.core.matchkey import (
+                    precompute_matchkey_transforms_frame,
+                )
+                bucket_tbl = precompute_matchkey_transforms_frame(
+                    _tf(bucket_tbl), [mk]
+                ).native
             if bucket_tbl.num_rows >= 2:
                 score_buckets_arrow(
                     bucket_tbl, single_pass, mk, set(matched_pairs),
@@ -1939,6 +1974,10 @@ def run_fs_dedupe_bucketed(
     res["bucketed"] = True
     res["fused"] = False
     res["output_dir"] = out_dir
+    # Lever 1 A/B mechanism check: resident base width (narrow arm drops the
+    # ~40% __xform_ block) + whether per-shard recompute was active.
+    res["base_ncols"] = base.num_columns
+    res["xform_per_shard"] = _narrow
     return res
 
 
--- a/packages/python/goldenmatch/goldenmatch/core/matchkey.py
+++ b/packages/python/goldenmatch/goldenmatch/core/matchkey.py
@@ -289,9 +289,19 @@ def _xform_sig(field: MatchkeyField) -> str:
 
 
 def precompute_matchkey_transforms(
-    df: pl.DataFrame, matchkeys: list[MatchkeyConfig]
+    df: pl.DataFrame, matchkeys: list[MatchkeyConfig], *, xform: bool = True
 ) -> pl.DataFrame:
     """Add one __xform_<sig>__ column per unique (field, transforms) signature.
+
+    ``xform=False`` (Lever 1, frame-residency) materializes the synthesized
+    ``derive_from`` NE/field columns but SKIPS the wide ``__xform_<sig>__``
+    block, returning a ~40%-narrower frame. The bucketed FS out-of-core route
+    uses this to keep the resident ``base`` narrow and recompute each
+    ``__xform_`` per bucket shard at score time instead. Byte-identical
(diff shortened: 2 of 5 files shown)


benzsevern:
## Lever 1 A/B result (run 30824543298, large-new-64GB, same-box back-to-back)

| arm | scale | base_ncols | peak RSS | wall | pairs |
|-----|-------|-----------|----------|------|-------|
| bucketed (A, wide)        | 4M | 13 | 4663 MB | 94.8s | 1,351,297 |
| bucketed_narrow (B)       | 4M | **8** | **4350 MB** | **90.0s** | 1,351,297 |
| bucketed (A, wide)        | 8M | 13 | 7760 MB | 184.0s | 2,705,946 |
| bucketed_narrow (B)       | 8M | **8** | **7123 MB** | **175.8s** | 2,705,946 |

**Byte-parity:** unique/dupes/golden/pairs IDENTICAL between arms at both scales; `bucketed_engaged=true` both. ✅
**Mechanism:** base 13 → 8 cols (dropped the 5 `__xform_` columns), `xform_per_shard=true` on B. ✅
**Wall (gate b):** B is 4.5–5% FASTER, not regressed (narrower shards → less shard I/O offsets the per-shard recompute). ✅
**Peak (gate a):** B drops peak **6.7% (4M) / 8.2% (8M)** — a real win, but **far below** the ~40% (~1.6× scale) the spec predicted from `peak ≈ 1× base` + `__xform_ ≈ 40% of base width`.

**Verdict: gate (a) NOT met as written.** The mechanism works and the change is free (byte-identical, wall-neutral-to-better), but the `__xform_` columns are a much smaller fraction of base *bytes* than of its column *count*, and the per-shard recompute working set + non-base peak terms dilute the win. Absolute saving grows ~linearly (~80 MB/M) → optimistically ~10% at 50M (~75M → ~82M max scale), **not the ~1.6× (~120M)** Lever 1 was pitched for.

Not auto-merging — this is a measurement-vs-framing decision for @benzsevern (land the free ~8% lever vs. re-profile where the non-base peak actually lives).

benzsevern:
## Option 1 A/B result (sequential batched output, run 30828394037, large-new-64GB, same-box)

| arm | scale | batched_output | peak RSS | wall |
|-----|-------|----------------|----------|------|
| sequential (A, legacy join) | 4M | false | 9728 MB | 130.2s |
| **sequential_batched (B)**  | 4M | **true** | **5996 MB** | **125.1s** |
| sequential (A, legacy join) | 8M | false | 18447 MB | 301.7s |
| **sequential_batched (B)**  | 8M | **true** | **11126 MB** | **296.9s** |

**Byte-parity:** unique/dupes/golden IDENTICAL both scales; pairs null both (fused path); streaming_engaged=true both. ✅
**Peak (the win):** −38.4% (4M) / −39.7% (8M). Sequential's peak was output-bound — the `frame.join(asn).join(sizes)` 2×-frame copy was ~40% of the whole peak. ✅
**Wall:** −3.9% / −1.6% (neutral-to-better). ✅

**Does NOT alone clear 50M.** sequential_batched slope ≈ 1.3 GB/M (868 MB + 1282 MB/M) → ~65 GB at 50M (marginal/OOM). Old sequential ≈ 2.2 GB/M (OOMs ~28M). So Option 1 ~doubles the ceiling (~28M → ~46M); the residual ~0.5 GB/M over bucketed is **Blocker 2** (the whole-frame fused scoring copy). To reach 50M on the sequential route you need Option 2 (bounded scoring, e.g. `FS_BLOCK_SOURCE=frame`) OR just route ≥~45M to bucketed.

**Verdict: strong free win** (unlike Lever 1's ~8%). Recommend landing. Not burning a 50M run — the extrapolation already predicts marginal/OOM (handoff rule: don't burn a 50M run to confirm a predicted OOM).

benzsevern:
## Sequential scale A/B, run 1 (30833838355) — OOM-killed the runner, re-running

Ordering mistake: the workflow ran `sequential_batched` (Option 1, fused) at all scales first, so `sequential_batched@50M` ran before `sequential_scale`. It **hard-OOM'd** (predicted ~65 GB > 64 GB), which OS-killed the runner and cancelled the step before `sequential_scale` executed. No artifact (the documented "50M OOMs to no artifact" gotcha).

Recovered from the log (both `sequential_batched`, consistent with run 30828394037):
- 4M: peak 5936 MB, wall 100.7s
- 8M: peak 11154 MB, wall 233.1s

**Confirmed:** Option 1 alone does NOT reach 50M — Blocker 2 (the fused whole-frame scoring copy) is real, and it OOMs hard. But `sequential_scale` (the bounded fix) never got to run.

Re-dispatched `sequential_scale` ONLY (run 30839839772) at 4M/8M/50M — its 4M/8M print first (recoverable even if 50M OOMs), and 50M is the actual proof. Lesson (again, the handoff's): don't schedule a predicted-OOM leg ahead of the leg you actually want to measure.

benzsevern:
## Sequential SCALE proof (run 30839839772, sequential_scale, large-new-64GB) — WIN

| mode | scale | peak RSS | wall | completed |
|------|-------|----------|------|-----------|
| sequential_batched (Opt 1, fused) | 4M | 5936 MB | 100.7s | ✓ |
| **sequential_scale (bounded)**    | 4M | **5250 MB** | **52.5s** | ✓ |
| sequential_batched (Opt 1, fused) | 8M | 11154 MB | 233.1s | ✓ |
| **sequential_scale (bounded)**    | 8M | **8433 MB** | **111.1s** | ✓ |
| sequential_batched (Opt 1, fused) | 50M | **OOM — killed runner** | — | ✗ |
| **sequential_scale (bounded)**    | 50M | **45801 MB (44.7 GB)** | **975s (16.3 min)** | **✓** |

**50M PROOF: sequential answers the scale question in-RAM.** Bounded `sequential_scale` completes 50M at 44.7 GB (< 64 GB, ~18 GB headroom), NO disk shards. Fused/Option-1-alone OOMs there.

**Peak:** bounded lowers peak −11.6% (4M) / −24.4% (8M) vs fused (the whole-frame scoring copy grows with N, so bounding helps more at scale). Slope ~0.9 GB/M — essentially bucketed's ~1× base.

**Wall:** bounded is ~2× FASTER than fused (52.5 vs 100.7s @4M; 111 vs 233s @8M). The feared fused-speed tradeoff is inverted — score_buckets_arrow (native value-dedup + small-block batch + parallel buckets) beats the fused whole-frame call here, and in-RAM bucketing skips bucketed's disk round-trip. Cross-run (caveat) sequential_scale is also faster than bucketed (8M 111 vs 184s; 50M 975 vs ~1178s serial).

**Byte-parity:** unique/dupes/golden IDENTICAL to the fused baseline at 4M/8M, AND the 50M counts (unique 36,709,249 / dupes 13,290,752 / golden 6,119,818) match the handoff's bucketed 50M reference exactly. bounded_scoring=true / batched_output=true all legs.

**Verdict:** sequential now matches bucketed's residency, in-RAM, with no disk shards, and is the fastest of the three routes. Recommend landing Option 1 + the bounded scale path — strong case to make bounded the DEFAULT sequential behavior.

benzsevern:
**Split:** Option 1 (batched output) + the bounded in-RAM scale path have moved to their own PR **#2397** (bounded default, the sequential 50M-in-RAM win), and this branch has been reset to the Lever 1 narrow-base commit only. So #2390 is now purely the on-hold Lever 1 (~8% peak, decision pending). The Option-1/bounded result comments above are retained for history but the code lives in #2397.

benzsevern:
Closing this. The A/B already answered it and the answer has aged badly, not because the work was wrong but because the problem it was pitched for got solved another way.

**The measurement stands** (run 30824543298, above): byte-identical at both scales, wall 4.5-5% FASTER, peak down 6.7% at 4M and 8.2% at 8M. The mechanism does exactly what it says. What it doesn't do is clear gate (a): the spec predicted ~40% peak and ~1.6x max scale (~75M to ~120M) from `peak ~= 1x base` and `__xform_ ~= 40% of base width`. The columns are a much smaller share of base BYTES than of its column COUNT, so the real number is ~8%.

**The scale question it existed for is answered on `main`.** `sequential_scale` (bounded scoring) completed 50M at 44.7 GB with ~18 GB headroom, where the fused path hard-OOM'd the runner. `FS_BLOCK_SOURCE` / `batched_output` are on `main` today with 20 references in `fs_out_of_core.py`; `FS_XFORM_PER_SHARD` is not. The non-base peak turned out to be where the second comment said it was, and that lever landed.

**What closing costs versus keeping it open:**

- The branch is **257 commits behind** `main`, and the drift is concentrated in `fs_out_of_core.py`, which has been through Phase 1 frame-residency (#2365) and the streaming cleanup since. A rebase is real conflict work in a hot file.
- Its gates are red for a mechanical reason (the new flag needs `gen_config_matrix.py --write goldenmatch`, `--manifest`, and `agent_codemap.py --write`), so it's not merge-ready as it stands either.
- The payoff is ~8% peak on an opt-in path, default OFF, that nobody has asked for and whose scale thesis is now served elsewhere.

That trade doesn't clear. Rebasing a hot backend file to carry a flag that misses its own gate by 5x is more risk than the 8% is worth.

The numbers are the durable part and they stay on this PR. If the `__xform_` width ever becomes the dominant peak term again, this is a measured starting point rather than a fresh idea.

PR #2394, opened 2026-08-03
feat(from_dbt): recognize shared-key edge/crosswalk models -> identity idioms

benzsevern:
## What

Closes a real gap in the existing `from_dbt` converter for the **external-matcher + dbt-postprocessing** shape.

`from_dbt`'s recognizers target **in-dbt** fuzzy matching (window-dedup, GROUP BY, levenshtein/soundex). But when the probabilistic matcher lives **outside** dbt (e.g. an external Splink step), the remaining dbt ER models are deterministic **edge / crosswalk self-joins** on a shared value. The fuzzy-only self-join path (signal 5) deliberately skips those, so nothing was extracted from that whole shape.

## How

New **signal 7**: a self-join `a.col = b.col` with **no fuzzy predicate** present is an edge/crosswalk idiom, mapped onto GoldenMatch's identity idioms:
- an **authoritative** government/registry id (`npi`, `ssn`, `ein`, `dea`, `nabp`, …) → `identity.deterministic_merge_keys` (a unique id can't split across entities);
- a **soft** shared attribute (email/phone/org/address) → a `RelationshipRule` with the right `kind` + `transform` (e.g. `org_name` → `same_org` / `normalize_company`);
- a surrogate/unknown key → stays silent (no false identity edge).

`_emit_config` now emits an enabled `IdentityConfig(deterministic_merge_keys, relationships)`, and the coverage scorecard gains `deterministic_merge_keys` + `relationship_edges` counts. This maps a crosswalk/edge model that links records sharing an authoritative id or soft attribute onto the relationship-graph (#2245) and deterministic-merge (#2389) features.

## Tests

`test_from_dbt_identity_edges.py` (5): npi crosswalk → merge key; phone/org edge → relationship rule + transform; fuzzy self-join **not** hijacked (signal 5 preserved); surrogate key ignored. All existing dbt tests unchanged. ruff + doc gates green.

### DIFF
--- a/packages/python/goldenmatch/goldenmatch/config/from_dbt.py
+++ b/packages/python/goldenmatch/goldenmatch/config/from_dbt.py
@@ -38,8 +38,10 @@ from goldenmatch.config.schemas import (
     GoldenFieldRule,
     GoldenMatchConfig,
     GoldenRulesConfig,
+    IdentityConfig,
     MatchkeyConfig,
     MatchkeyField,
+    RelationshipRule,
 )
 from goldenmatch.core._paths import safe_path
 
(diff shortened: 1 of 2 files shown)


benzsevern:
Superseded by a clean-history branch (removed employer-internal references from the commit message). See replacement PR.

Issue #876, opened 2026-06-12
Residual precision drift at scale (>1M) under a fixed config: fuzzy false-positives grow with entity count


benzsevern:
Discovered in the #510 quality-invariant scale audit, after fixing the phonetic-identity scale bug (the larger effect; see the autoconfig phonetic fix on the #510 branch).

**Symptom:** with a FIXED auto-derived config (frozen-config methodology) on the realistic-shape scale fixture, recall is dead-flat scale-invariant (0.988 at every rung) but PRECISION drifts down beyond ~1M:

| rows | pairwise F1 | precision | recall |
|---|---|---|---|
| 1K | 0.9352 | 0.8877 | 0.9880 |
| 100K | 0.9365 | 0.8904 | 0.9876 |
| 1M | 0.9364 | 0.8902 | 0.9878 |
| 10M | 0.9166 | **0.8547** | 0.9882 |

Precision is flat 1K->1M (0.888-0.890) then drops to 0.855 at 10M. F1 misses the strict Δ≤0.005 invariance target at 10M (Δ=-0.019).

**Hypothesis (unconfirmed):** the remaining matchkeys are exact_email, exact_identity (first+last+birth_year), and a weighted fuzzy_match (name + address, threshold 0.8). The exact composites are scale-safe (unique names). The suspect is fuzzy false-positives growing with entity count -- random name/address near-collisions, and/or the audit's ambiguous-twin pairs (which share a name and differ only on secondary fields) scoring just over the 0.8 threshold more often as clusters multiply. This is partly a fundamental ER-at-scale property (more entities -> more confusable pairs) and partly possibly an auto-config threshold/blocking lever (a scale-adaptive fuzzy threshold, or a more discriminating blocking key).

**Next steps:** re-run 10M with per-pair false-positive instrumentation to attribute the fp's (twins vs random near-collisions vs blocking), then decide whether it's a config lever (scale-aware threshold) or an inherent property to document. The #510 report documents it as an open finding with the full curve.

**Context:** #510 quality-invariant scale validation; the phonetic-identity fix (the dominant effect) is on the feat/510 branch.

benzsevern:
**Root cause pinned: it's the blocking key, not generic fuzzy false-positives.**

The frozen config blocks on a single key: `zip` (`max_block_size: 5000, skip_oversized: false`). In the fixture `zip = cid % 100000`, which WRAPS at 100K clusters, so the zip block size grows ~linearly with N:

[...]

PR #2466, opened 2026-08-10
fix(autoconfig): decide surrogate keys from the full frame; land bot data commits via PR

benzsevern:
## What and why

Started as two red scheduled lanes. One was a CI-plumbing bug; the other turned out to be a **zero-config scale-invariance defect**, with the runner death as its symptom.

### 1. No workflow has been able to commit to `main` since 2026-06-15

The North Star scoreboard's first-ever run did all its work and died on the last step:

```
remote: error: GH013: Repository rule violations found for refs/heads/main.
  - Changes must be made through a pull request.
  - Required status check "ci-required" is expected.
  - Changes must be made through the merge queue
```

`git log --since=2026-06-15` shows **zero** bot commits on `main`, so this was never scoreboard-specific — `benchmarks.yml`'s report-commit step is dead the same way, invisibly, because it only fires on the weekly `datasets=all` schedule and sat behind an earlier failure.

### 2. The 5M heavy tier has never once passed — and it is an OOM

I instrumented rather than guessed, because artifacts do **not** survive this death (the `if: always()` upload step is itself skipped when the runner goes down), so the only forensic channel is the streamed log. The heartbeat settled it:

```
t= 696s rss=36.5GB avail=28.4GB stage=n=5000000:measure_red
t= 787s rss=48.9GB avail=16.3GB
t= 877s rss=65.6GB avail= 0.2GB
t= 908s rss=66.0GB avail= 0.0GB
   … 11.6 minutes pinned at 66.0GB / 0.0GB …
##[error]The runner has received a shutdown signal.
```

Not preemption. And the OOM is downstream of the real bug:

`ColumnProfile.cardinality_ratio` is a distinct-**fraction** over the ~1,000-row profiling sample, and the `#876` surrogate guard compares it to an **absolute 1.0** to ask *"does every record carry its own value?"*. A fixed-size sample of a growing frame drives that fraction toward 1.0 for **any** column whose distinct count grows with n — which is every real identifier:

| n | true `email` ratio | sampled ratio |
|---|---|---|
| 100K | 0.28 | 0.72 |
| 500K | 0.28 | 0.94 |
| 1M | 0.28 | 0.97 |
| 2M | 0.28 | 0.98 |
| 5M | 0.28 | **1.00** |

So at 5M zero-config discarded its only exact identity column as a *"perfectly-unique surrogate key"*, fell back to fuzzy-only matchkeys and name-based blocking, and committed the 185M-pair config whose force-run took 66 GB. **The verdict changed with scale, not with the data** — exactly what the scale-invariant-correctness commitment forbids. Reproduced locally on byte-identical generated data: 1M keeps `email`, 5M drops it.

## Changes

- **`_is_perfect_surrogate`** — eight call sites asked this question independently off the sampled number; they now share one authority that prefers an **exact full-frame** cardinality, falling back to the sampled value only where no full frame exists (hand-built test profiles), so those keep their prior verdict. The change can therefore only ever *retain* a column that used to be dropped; a genuine row PK still reads 1.0000 and is still excluded.
  - Restricting the exact count to sample-ratio-1.0 columns is **sound, not a cost trade**: if every value in the full column is distinct then so is every subset, so `full == 1.0` implies `sample == 1.0`. Measured 44–307 ms/column at 5M against a ~160s auto-config, on one or two columns.
- **`scripts/open_data_pr.sh`** — shared commit-back for `scoreboard.yml` + `benchmarks.yml`: rolling bot branch → PR → auto-merge → merge queue. `GITHUB_TOKEN` drops to `contents: read`; `benchmarks.yml`'s commit subject loses its `[skip ci]` (it must pass `ci-required` to merge now).
- **`qis_gate` heartbeat** — 30s `rss` / `peak` / `MemAvailable` / stage on stdout. `MemAvailable` is the discriminator; this is what produced the trace above.
- **Snapshot uploaded as an artifact *before* the landing step**, so a week's numbers survive a missing token instead of being discarded with the run.

## Testing

**Effect at 5M** — committed matchkeys now identical to 1M, instead of the degenerate fuzzy-only config:

```
1M: exact(email), exact(first_name+last_name+birth_year), weighted(first,last,address)
5M: exact(email), exact(first_name+last_name+birth_year), weighted(first,last,address)
```

- **7 new tests, and 3 verified to FAIL on the pre-fix helper** — the other 4 pin behaviour that must *not* change, including the fallback. The crossing is reproduced deterministically at 50,000 rows through the default 1,000-row sample: 50 duplicate pairs give an expected **0.02** whole pairs in the draw, so the sample saturates at 1.0 against a full-frame 0.999. No reliance on a rare event, so it is not a flaky construction.
- **Full goldenmatch suite: 6 failed / 9123 passed — the same six failures as a pristine-tree baseline** of the identical command (6 failed / 9116 passed). The +7 is this PR's tests. Baselined rather than asserted.
- `scripts/test_qis_gate.py` 18 passed · `ruff` clean · workflow-YAML (118 files) · docs-consistency · context-budget · codemap regenerated.
- Post-fix heavy-tier `workflow_dispatch` running now; this stays draft until it reports.

`pre-commit` is not installed in this environment, so I ran the gate scripts it wraps individually rather than claim a clean `pre-commit run --all-files`.

## Checklist

- [ ] `pre-commit run --all-files` is clean — not installed here; ran `ruff` + workflow/docs/budget/codemap checks individually
- [x] Package `CHANGELOG.md` updated if behavior changed

## Needs a maintainer action

**`GM_BOT_TOKEN` must exist** (fine-grained PAT, `contents: write` + `pull-requests: write`, owner `benseverndev-oss`) or the commit-back steps fail loudly. That is deliberate: the default `GITHUB_TOKEN` cannot be substituted, because PRs it opens do not trigger workflows, so `ci-required` would never report and the PR would wedge in the queue permanently — the same trap as a skip-CI directive.

## Found on the way, NOT fixed here

- **The nightly `ci` tier reports green while its top rung is skipped.** `birth_year` blocking (65 distinct) estimates 7.69B pairs at 1M, over the 3B ceiling, so it is dropped as non-gating and real gated coverage ends at 500K — for a gate built to catch regressions that "sat broken at >=500K rows for ~2 months". Note the sampled ratio at 500K is still only 0.936, which is why this bug lived below the gate's reach.
- **`MEASURE_MAX_PAIRS` is a wall-clock guard and cannot see a memory wall.** 5M sat 16x *under* the 3B ceiling and still took 66 GB. A memory-aware preflight is worth adding as defence-in-depth, independently of this fix.
- **`test_healer_default_e2e_native::test_suggest_from_result_verified_matches_review_config` is flaky.** It failed in one of two full runs with this change and passed in the other, passes in isolation both with and without it, and is the same shard-isolation class already deselected in CI for `test_memory_*`. Per the repo's flaky-test-first rule this deserves a root-cause fix; I did not derail this PR into a third subsystem, but I can take it next.


benzsevern:
Post-merge verification result, since the body says this stayed draft until the heavy-tier run reported and it did not — it was merged while that run was still in flight. Correcting the record.

## The OOM is fixed

Instrumented 5M heavy-tier dispatch on this branch's head:

```
t=5168s  rss=17.0GB peak=17.6GB avail=47.5GB  stage=n=5000000:measure_red
t=5569s  rss=17.0GB peak=17.6GB avail=47.4GB  stage=n=5000000:measure_red
```

**Peak 17.6 GB, flat, with 47.5 GB free throughout** — against 66.0 GB climbing to `avail=0.0GB` and an 11.6-minute thrash before the runner died. Memory is bounded and stable. That confirms the causal chain: the sampled-cardinality surrogate misfire produced the fuzzy-only config whose scoring exhausted the box.

## But the rung still does not finish, for a DIFFERENT reason

The run was cancelled at 93 minutes, still in `measure_red`, RSS pinned flat — the signature of grinding through candidate pairs, not leaking. The cause is a second, independent instance of the same fixed-sample-on-a-growing-frame pathology, this time in **learned blocking**:

| n | sample | true pairs found | blocks | rows/block | est. candidate pairs | block cap |
|---|---|---|---|---|---|---|
| 1M | 5000 | 56 | 791 | 1,264 | 0.6B | 25,000 |
| 5M | 5000 | **13** | 160 | **31,250** | **78.1B** | 50,000 |

The trainer samples a fixed 5,000 rows regardless of `n`. Duplicate-pair yield in a fixed sample scales as `s²/n`, so 1M → 5M predicts `56/5 ≈ 11` true pairs; it observed **13**. With 13 positive examples it learns rules too coarse to separate anything, giving 160 enormous blocks instead of 791 small ones — **5x the data, 124x the pair work**. `max_block_size` also doubles (25,000 → 50,000) as `n` grows, so the guard loosens exactly where it should tighten.

Same shape as the bug this PR fixed, different consumer: a fixed-size sample cannot estimate a property that scales with frame size. The fix direction is the same too — sample to a target number of POSITIVE examples rather than a fixed row count.

Not fixed here. Filing separately.

## Also still open

- `MEASURE_MAX_PAIRS` is a wall-clock guard and cannot see a memory wall; it did not fire at 5M (185M estimated, 16x under the 3B ceiling) while the run took 66 GB. Worth a memory-aware preflight regardless.
- The nightly `ci` tier reports green while its 1M rung is skipped as a pair explosion, so gated coverage ends at 500K. That is why both of these lived below the gate's reach.

---

PR #2479, opened 2026-08-11
test(spark): backend-agnostic spark fixture + the P0 real-Spark Connect lane

benzsevern:
Executes **P0** of the Spark-native execution spec (#2476), per plan #2477. Tests and CI only — **no engine code changed.**

## The obstacle

The `spark` fixture was duplicated across **nine** test files, each hardcoding `pysail.spark.SparkConnectServer`. That's why the tier had never been run against real Spark: **the tests couldn't express it.**

## The finding the plan missed — and it would have made the new lane a lie

Those files also carried **module-level `pytest.importorskip("pysail")`**. On a lane where pysail is deliberately absent, every test would **skip**, and the lane would report green while proving nothing.

Same shape as the zero-polars gate trap, one level down. The pysail guard is now the *fixture's* business; each file keeps a `pyspark` guard instead.

## Deviation from the plan, on evidence

The plan put the fixture in a separate `tests/conftest_spark.py` imported by each file, to keep a pyspark import out of the root conftest. **That produced 41 ruff errors** — `F811` in every file whose tests take `spark` as a parameter.

The concern behind it turned out void: the root conftest holds only stdlib + pytest, and this fixture's imports are **all inside the body**, so a suite that never requests it pays nothing at collection. Fixture now lives in `tests/conftest.py`. No cross-file imports, no `noqa` noise.

## Other things that surfaced

- Two files (`identity_parity`, `identity_incremental`) kept their guards *inside* the fixture rather than at module level. An assertion caught it mid-run rather than silently emitting files with no pyspark guard at all. Stale docstrings corrected.
- `test_sail_connectivity.py` built its server inline in the test body rather than as a fixture. Rewritten to use the shared one, so the S1 harness gate now proves whichever backend the env selects.

## The lane

Installs `pyspark[connect]>=4` and **not** pysail, then **asserts pysail is absent** — a lane that silently picked it up from the lockfile would prove nothing. Runs the same tests with `GOLDENMATCH_SPARK_REMOTE=local[*]`: Spark's own in-process Connect server, no container, no cluster. (`goldenmatch[sail]` pins `pyspark<4` *because of* pysail, so the client is installed directly.)

Deliberately `continue-on-error` and **not** in `ci-required`. While P0 runs, **a red is a finding to enumerate, not a merge blocker** — promote at P2. Verified programmatically: present, `continue-on-error: true`, absent from `ci-required`.

## Verification

Local checks are collection + `py_compile` + ruff only (all clean; the wider suite still collects, confirming the global conftest change is inert). Running this needs pyspark and a JVM, which OOMs the dev box — **CI is the harness**, which is the whole point of the lane.

**What to look at when it runs:** the `spark_connect` job's result is the P0 deliverable. Each deviation gets classified (a) Spark Connect API gap → reshapes the spec, (b) accidental Sail-ism → P2 fix, (c) test-harness assumption → noise.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2985,6 +2985,57 @@ jobs:
         run: |
           .venv/bin/python -m pytest packages/python/goldenmatch/tests/test_sail_pipeline.py -v --timeout=300
 
+  # P0 (spec docs/superpowers/specs/2026-08-10-spark-native-execution-design.md):
+  # the SAME sail tests against APACHE SPARK's own Connect server, not Sail's.
+  # The tier is believed backend-agnostic -- goldenmatch/sail/session.py is
+  # `SparkSession.builder.remote(url)` and there is no Sail-specific call
+  # anywhere in the package -- and this lane is what turns that belief into a
+  # fact before P1-P6 are built on it.
+  #
+  # NOT in ci-required while P0 runs: a red here is a FINDING to enumerate, not
+  # a merge blocker. Promote to required once green (P2).
+  spark_connect:
+    needs: changes
+    if: needs.changes.outputs.sail == 'true' || needs.changes.outputs.force_all == 'true'
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    continue-on-error: true
+    steps:
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10  # v6
+      - uses: astral-sh/setup-uv@caf0cab7a618c569241d31dcd442f54681755d39  # v3
+        with:
+          enable-cache: true
+          cache-dependency-glob: uv.lock
+      # Real Spark's Connect server is JVM-side (unlike Sail's Rust one).
+      - uses: actions/setup-java@v4
+        with:
+          distribution: temurin
+          java-version: '17'
+      - run: uv sync --all-packages --no-install-package goldenmatch-native --no-install-package goldenflow-native
+      # NO pysail. goldenmatch[sail] pins pyspark<4 BECAUSE of pysail, so install
+      # the client directly to get Spark 4, whose `local[*]` remote spawns
+      # Spark's own Connect server in-process (no container, no cluster).
+      - name: Install pyspark connect (no Sail)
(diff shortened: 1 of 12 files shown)


benzsevern:
## P0 result: the tier runs on real Spark Connect. Deps are the blocker.

Run [31496638072](https://github.com/benseverndev-oss/goldenmatch/actions/runs/31496638072):

```
spark_connect:  20 failed, 36 passed, 8 skipped in 408s
sail (control): PASS
```

The control matters — the `sail` lane is **green on the same commit**, so the fixture extraction is behaviour-preserving and the delta is genuinely the backend.

### No Spark Connect API gaps

36 tests pass. Session creation, DataFrame ops, SQL, config validation, the entire `test_sail_r3_feature_gate.py` surface — all unmodified against real Spark. **The "backend-agnostic" inference is confirmed, and P1–P6 are not reshaped.**

### All 20 failures are one cause

```
ModuleNotFoundError: No module named 'goldenmatch'   (73×)
ModuleNotFoundError: No module named 'pandas'        (13×)
```

inside the **Python UDF worker**. The correlation is exact: every failing test executes the pipeline (hence a `pandas_udf`); every passing test doesn't.

**Classification: (c) environment — not (a) an API gap, not (b) a Sail-ism.**

### The finding worth keeping

**The pysail lane structurally cannot catch this.** Sail's Connect server runs *in-process*, so its Python worker shares the client's interpreter and `import goldenmatch` just works. Real Spark **forks a separate worker** with its own environment.

Every "it works distributed" signal from S1–S5 CI came from a harness that could not exercise the executor dependency path at all. The 2026-06-15 GKE run *did* exercise it — by baking deps into the worker image, which is exactly why it never surfaced as a gap.

So **P1 is not plumbing, it's the blocker** — demonstrated empirically rather than argued. `addArtifacts` is precisely the mechanism these 20 failures need.

### State of this PR

The `spark_connect` lane **should stay red until P1 lands**. That's the correct state, and why the lane is `continue-on-error` and out of `ci-required`. It joins `ci-required` at P2, once P1 makes it green.

Mergeable as-is: the control lane is green, and the red lane is the deliverable.

PR #2481, opened 2026-08-11
feat(spark): ship the client venv to executors via addArtifact (P1)

benzsevern:
Implements **P1** of the Spark-native execution spec. Plan included as the first commit.

**Stacked on #2479 (P0)** — its branch is merged in so the fixture and lane exist. The diff collapses to P1's own delta once P0 lands.

## What P0 established

The tier is **fully Spark-Connect-compatible** — 36 tests pass, no API gaps. Every one of its 20 failures was the same thing:

```
ModuleNotFoundError: No module named 'goldenmatch'   (73×)
```

inside the Python UDF worker. The tier works; its dependencies were never delivered.

## The mechanism

Connect-only, verified upstream:

```python
venv_pack.pack(output="gm_env.tar.gz")
spark.addArtifact("gm_env.tar.gz#environment", archive=True)
spark.conf.set("spark.sql.execution.pyspark.python", "./environment/bin/python")
```

Archives unpack executor-side automatically; session-based dependency management has existed since Spark 3.5.0. **This is what makes the zero-install cutover claim true rather than aspirational** — no platform ticket, no cluster libraries, no image rebuild.

Shipped as a supported helper (`goldenmatch/sail/deps.py`), not a docs snippet, because every user needs the identical four lines and the platform trap belongs next to the code.

## The probe is a UDF on purpose

A driver-side import check proves nothing — the driver is where the client venv already lives. A UDF body only executes in a Python worker, so reaching it establishes we're off the driver. The tests assert `ran_on == "executor"` as a guard on the guard.

## Two things I want reviewed

**The platform trap.** The archive must be built for the **executor** platform. A venv packed on macOS/Windows uploads and unpacks happily, then fails to execute on Linux — and because the scorer falls back to pure rapidfuzz on any error, **that failure is silent**: correct results, none of the speed. CI is Linux so the lane matches by construction; users get the warning in the docstring.

**`native_kernel: false` is not a P1 failure**, and a test pins that distinction. `sail_scoring` is `_FALLBACK_ONLY` (f32 vs the f64 floor) and doesn't run under `auto` regardless — lifting it is P3. Without that pin, a `false` here would read as a regression.

## Deliberately not done

Pointing `spark.sql.execution.pyspark.python` at the client's own `.venv`. It would turn the lane green under `local[*]` and prove **nothing**, because a real cluster has no such path. The archive is the entire point.

The new tests also **skip** unless `GOLDENMATCH_SPARK_REMOTE` is set — under pysail the worker shares the client interpreter, so they'd pass vacuously. That sharing is exactly why P0's failure class stayed invisible through all of S1–S5.

## Exit criterion

The `spark_connect` lane going green — the 20 P0 failures passing. Read the lane; that's the deliverable.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3047,11 +3047,11 @@ jobs:
           # unpacks fine and imports nothing.
           /tmp/gmenv/bin/pip install ./packages/python/goldenmatch pandas pyarrow
           echo "--- shipped env contents ---"
-          /tmp/gmenv/bin/pip list | grep -iE "goldenmatch|rapidfuzz|pandas|pyarrow"
+          /tmp/gmenv/bin/pip list | grep -iE "goldenmatch|goldenphonetic|pandas|pyarrow"
           # cd out of the repo: its ROOT DIRECTORY is itself named `goldenmatch`,
           # so a check run from there can import the source tree instead of the
           # installed package and pass while the env is empty.
-          (cd /tmp && /tmp/gmenv/bin/python -c "import goldenmatch, rapidfuzz, pandas, pyarrow; print('shipped env imports OK')")
+          (cd /tmp && /tmp/gmenv/bin/python -c "import goldenmatch, goldenmatch.core.strsim, pandas, pyarrow; print('shipped env imports OK')")
           uv pip install venv-pack
           .venv/bin/venv-pack -p /tmp/gmenv -o gm_env.tar.gz --force
           ls -lh gm_env.tar.gz
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3023,10 +3023,8 @@ jobs:
         run: |
           .venv/bin/python -c "import importlib.util,sys; sys.exit(0 if importlib.util.find_spec('pysail') is None else 'pysail present; this lane must exercise real Spark')"
       # P1: real Spark forks a Python worker with its own env, so the client's
-      # site-packages are absent -- that was P0's entire failure class. Pack the
-      # venv here (CI is Linux, so it matches the executor platform by
-      # construction) and ship it via addArtifact at session time.
-      # Build a PURPOSE-BUILT runtime venv and pack that, rather than packing
+      # site-packages are absent -- that was P0's entire failure class. Build a
+      # PURPOSE-BUILT runtime venv and pack that, rather than packing
       # CI's own uv-created .venv. Two reasons:
       #   1. venv-pack rejects a uv venv outright ("Current environment is not a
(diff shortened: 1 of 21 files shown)


benzsevern:
## P1 result: dependency delivery works. 18 of 20 fixed.

Run [31516855744](https://github.com/benseverndev-oss/goldenmatch/actions/runs/31516855744):

| | failed | passed | skipped |
|---|---|---|---|
| **P0** (no deps shipped) | 20 | 36 | 8 |
| **P1** (venv via addArtifact) | **2** | **54** | 8 |

Every `ModuleNotFoundError` is gone. `addArtifact` + `spark.sql.execution.pyspark.python` delivers goldenmatch to a forked Python worker with **no cluster-side install** — the zero-friction claim is now demonstrated, not argued.

The packed archive was independently verified by the fast lane (run 31514938174): `goldenmatch`, `goldenmatch.core.strsim`, `pandas`, `pyarrow` all import on the shipped interpreter, and **`native kernel present: True`** — the abi3 wheel travels, so P3 will have a kernel to switch on.

## The 2 remaining are a different class

Both are `test_sail_clustering_parity` WCC chain tests:

```
SparkException: [STAGE_MATERIALIZATION_MULTIPLE_FAILURES]
Not enough memory to build and broadcast the table to all worker nodes.
  As a workaround … spark.sql.autoBroadcastJoinThreshold
OutOfMemoryError: Java heap space
```

Not a dependency problem, not a Spark Connect API gap. The iterative WCC join loop triggers a broadcast join that exhausts the JVM heap on a GitHub runner.

**This is the same code that wedged on Sail at 12K rows** — for a different reason there (no lineage truncation) and a different one here (broadcast + heap). Real Spark gets further before it hurts.

## The decision I'm not making unilaterally

The obvious fix is `spark.sql.autoBroadcastJoinThreshold=-1` around the WCC loop. But there are two readings and they lead different places:

- **CI-runner artifact** — a 7 GB heap is small; a real cluster wouldn't hit it. Then it's test config.
- **Genuine tier defect** — an iterative join loop that relies on broadcast will hurt at scale on any cluster. Then it belongs in the engine, and "use a bigger machine" is how a scale bug gets shipped.

I'd lean the second, since the WCC has now failed on both backends for different reasons, and that pattern usually means the loop is the problem rather than the engine. But it's a tier change, so it wants its own scope rather than being smuggled into P1 as a config tweak to make a lane green.

**Proposal:** merge P1 on 18/20 with the two marked `xfail` and a pointer to a P2 item for the WCC broadcast behaviour. That keeps the dependency-delivery win landed and the scaling question honest and separate. Say the word if you'd rather I fix it here.

PR #2499, opened 2026-08-12
feat(spark): the Spark tier, end to end — config-driven, Fellegi-Sunter, zero-config (P4-P6)

benzsevern:
The Spark tier took **scalars** — one `block_col`, one `value_col`, one scorer, one threshold, one survivorship strategy. A real config carries many blocking passes, many matchkeys, many weighted fields each, and per-field golden rules, and `GoldenMatchConfig` never reached the tier at all.

This adds `run_config_pipeline(df, config)`.

## What it executes

- **Multiple blocking passes**, unioned and de-duplicated — a pair reachable by two passes is *one* candidate
- **Multi-field weighted matchkeys**, and exact matchkeys
- **Multiple matchkeys as alternatives**, MAX-deduped (the tier's existing contract)
- **Per-field survivorship strategies**

Everything else is refused **loudly** by `_validate_spark_config_supported`, mirroring the scale-mode posture: probabilistic matchkeys, negative evidence, guards, rerank, `auto_threshold`, LLM, domain extraction, semantic blocking, every non-static blocking strategy, and conditional `when:` golden rules. A silently degraded distributed result is worse than a refusal.

## The acceptance criterion as spec'd is not reachable in P4

`config/from_splink.py` emits `type="probabilistic"` **unconditionally**. There is no `m_prob` / `u_prob` / `match_weight` anywhere in this tier, so *"`import-splink` output runs unchanged"* depends on **P5** (Fellegi-Sunter on Spark), not on P4.

P4a refuses a probabilistic config with a message naming P5, and the Splink cutover claim moves to P5's exit criteria. This is a phase-ordering error in the spec, found by execution rather than by reading it: **P5 is the load-bearing phase for the product goal, and P4 is its precondition.** The spec now records it.

## Two defects found while building it, both in the shipped tier

**1. `null` vs `null` scored a perfect 1.0.** The scorer kernel maps a missing value to `""`, and `""` vs `""` is an exact match — so two records whose only shared evidence was that *both are missing* the compared field merged at **every** threshold. The one-box excludes the field instead (`score_field` → `None`, dropped by `score_pair`).

Measured on the new fixture: a pair that should score **0.4365** scored **0.8591** and merged.

Fixed on the new path *and* on the shipped single-field `score_and_dedup`, by deciding comparability from the **raw columns** rather than from the kernel's output — so it stays correct even if the `""` substitution is later changed at source.

**2. The weighted denominator is per-pair, not per-matchkey.** `score_pair` drops an uncomparable field from the numerator *and* the denominator. Using the matchkey's total weight instead scores a pair that agrees perfectly on its one comparable field at **0.7500** rather than **1.0000** — losing it entirely.

Both numbers are recorded in the fixture's comment, because each bug **flips a decision there**. The fixture discriminates rather than decorates, and an editor changing those rows needs to know it.

## Parity by reuse, not re-derivation

Block keys call `apply_transforms`; survivorship calls `merge_field` — the same functions the one-box calls, so agreement is structural rather than maintained by hand.

The one restated formula is the weighted combine, which has to be a Spark expression to stay vectorized (that vectorized per-field kernel is the tier's whole differentiator). So `weighted_pair_score` is a pure-Python statement of `score_pair`'s math, **unit-tested directly against `score_pair`** across six null shapes, with the Spark expression bound to it by the lane parity test.

## Two implementation notes

Both chosen to avoid risk I can't test locally:

- Block keys compose from **single-column UDFs plus Spark expressions**, not one variadic pandas UDF. Pandas UDFs resolve their eval type from the signature and `*args` isn't in the documented contract (checked against the Spark docs, which only ever show fixed arity). `concat_ws` can't do it alone either — it *skips* nulls, so `("a", null)` and `("a", "")` would collide into the same key.
- Join aliases are `__lhs__` / `__rhs__` because the candidate frame's own columns are named `a` and `b`, which makes `F.col("a.first")` ambiguous.

## Testing

- **19 unit tests run on every PR** (no Spark needed): the combine vs `score_pair`, and the whole feature gate including a test that a supported config *passes* — a gate that refused everything would pass every refusal test while disabling the feature.
- **Lane tests** (`spark_connect` blocking, `sail` advisory): candidate-set, scored-pair-set and cluster-partition parity against an independently written Python reference, plus the null-merge regression on the shipped path at four thresholds.

The reference is computed in plain Python here rather than by calling the tier a second way — a reference sharing code with the thing under test proves only self-consistency.

Gates run locally before pushing: workflow YAML, filter coverage, claude refs, docs consistency, version consistency, context budget, docs staleness, native symbols, agent codemap (regenerated), ruff. `api_parity` and `ts_parity_freshness` fail identically on clean `main` in this worktree — they need the built TS workspace — so they're left to CI.

## Next

**P4b** wires `backend="spark"` into the public API. Worth noting up front: `dedupe_df` coerces its input to a local Frame, so the dispatch has to happen *before* that coercion — collecting a cluster-sized frame to the driver would defeat the tier entirely.

### DIFF
--- a/docs-site/goldenmatch/config-matrix.mdx
+++ b/docs-site/goldenmatch/config-matrix.mdx
@@ -146,7 +146,7 @@ Every config object in the pydantic tree(s), generated from the package schema. 
 | `llm_scorer` | LLMScorerConfig \| None | `None` | [`LLMScorerConfig`](#llmscorerconfig) -- LLM pair-scoring configuration for borderline candidates. |
 | `llm_auto` | bool | `False` | Lets auto-config enable and configure the LLM scorer automatically. |
 | `domain` | DomainConfig \| None | `None` | [`DomainConfig`](#domainconfig) -- Domain feature-extraction configuration used before matchkeys. |
-| `backend` | str \| None | `None` | Execution backend: None (default Polars in-memory), 'ray', 'duckdb', 'chunked', or 'bucket'. |
+| `backend` | str \| None | `None` | Execution backend: None (default in-memory), 'ray', 'duckdb', 'datafusion', 'chunked', 'bucket', 'polars-direct', or 'spark'. |
 | `distributed_routing` | DistributedRoutingConfig \| None | `None` | [`DistributedRoutingConfig`](#distributedroutingconfig) -- Per-stage distributed-routing pins; None lets the planner decide every stage. |
 | `semantic_blocking` | SemanticBlockingConfig \| None | `None` | [`SemanticBlockingConfig`](#semanticblockingconfig) -- Opt-in semantic candidate-generation keys (ANN, initialism, alias) unioned into blocking. |
 | `allow_slow_path` | bool | `False` | Permits falling back to a slower non-fused execution path when the fast path is ineligible. |
--- a/docs/agent-codemap.json
+++ b/docs/agent-codemap.json
@@ -7702,6 +7702,7 @@
             "_transformed",
             "_valid_key",
             "_validate_spark_config_supported",
+            "blocking_passes",
             "build_golden_from_rules",
             "generate_candidates",
             "run_config_pipeline",
--- a/docs/agent-codemap.json
+++ b/docs/agent-codemap.json
@@ -4,7 +4,7 @@
   "packages": {
     "goldenmatch": {
       "root": "packages/python/goldenmatch/goldenmatch",
-      "module_count": 460,
+      "module_count": 461,
       "modules": [
         {
           "module": "goldenmatch",
@@ -7668,6 +7668,33 @@
             "_truncate_plan",
             "connected_components",
             "connected_components_scale"
+          ]
+        },
+        {
+          "module": "goldenmatch.spark.config_pipeline",
+          "file": "packages/python/goldenmatch/goldenmatch/spark/config_pipeline.py",
+          "purpose": "P4: run the Spark tier from a ``GoldenMatchConfig`` instead of scalars.",
+          "defines": [
+            "_block_key_column",
+            "_default_golden_cols",
+            "_field_score_parts",
+            "_field_strategy",
+            "_matchkey_score_expr",
+            "_transformed",
+            "_valid_key",
+            "_validate_spark_config_supported",
+            "build_golden_from_rules",
+            "generate_candidates",
+            "run_config_pipeline",
+            "score_candidates",
+            "weighted_pair_score"
+          ],
+          "imports": [
+            "goldenmatch.config.schemas",
+            "goldenmatch.spark.clustering",
+            "goldenmatch.spark.golden",
+            "goldenmatch.spark.scorers",
+            "goldenmatch.utils.transforms"
           ]
         },
         {
--- a/docs/agent-codemap.json
+++ b/docs/agent-codemap.json
@@ -4,7 +4,7 @@
   "packages": {
     "goldenmatch": {
       "root": "packages/python/goldenmatch/goldenmatch",
-      "module_count": 461,
+      "module_count": 462,
       "modules": [
         {
           "module": "goldenmatch",
@@ -7681,6 +7681,8 @@
             "_field_score_parts",
             "_field_strategy",
             "_matchkey_score_expr",
+            "_matchkey_threshold",
+            "_resolve_fs_models",
             "_transformed",
             "_valid_key",
             "_validate_spark_config_supported",
@@ -7692,8 +7694,10 @@
           ],
           "imports": [
             "goldenmatch.config.schemas",
+            "goldenmatch.core.probabilistic",
             "goldenmatch.spark.clustering",
             "goldenmatch.spark.golden",
+            "goldenmatch.spark.probabilistic",
             "goldenmatch.spark.scorers",
             "goldenmatch.utils.transforms"
           ]
@@ -7767,6 +7771,30 @@
             "goldenmatch.spark.identity",
             "goldenmatch.spark.scorers",
             "goldenmatch.spark.scoring"
+          ]
+        },
+        {
+          "module": "goldenmatch.spark.probabilistic",
+          "file": "packages/python/goldenmatch/goldenmatch/spark/probabilistic.py",
+          "purpose": "P5: Fellegi-Sunter scoring on Spark.",
+          "defines": [
+            "FSSparkUnsupported",
+            "_field_similarity_and_observed",
+            "_validate_fs_spark_supported",
+            "_weight_lookup_expr",
+            "fs_level_expr",
+            "fs_match_weight_expr",
+            "fs_pair_weight",
+            "fs_posterior",
+            "fs_posterior_expr",
+            "fs_score_expr",
+            "level_thresholds_for",
+            "resolve_fs_model"
+          ],
+          "imports": [
+            "goldenmatch.core.probabilistic",
+            "goldenmatch.spark.config_pipeline",
+            "goldenmatch.spark.scorers"
           ]
         },
         {
--- a/docs/agent-codemap.json
+++ b/docs/agent-codemap.json
@@ -7654,8 +7654,9 @@
         {
           "module": "goldenmatch.spark",
           "file": "packages/python/goldenmatch/goldenmatch/spark/__init__.py",
-          "purpose": "Sail tier (distributed, Spark Connect) -- the distributed sibling of the",
+          "purpose": "The Spark tier (distributed, Spark Connect) -- the distributed sibling of the",
           "imports": [
+            "goldenmatch.spark.config_pipeline",
             "goldenmatch.spark.identity"
           ]
         },
--- a/docs/agent-codemap.json
+++ b/docs/agent-codemap.json
@@ -7667,13 +7667,15 @@
           "purpose": "P6: zero-config on Spark.",
           "defines": [
             "SparkAutoConfigTooLarge",
+            "SparkAutoConfigUnsupported",
             "_refuse_at_n",
             "auto_configure_spark",
             "sample_to_driver"
           ],
           "imports": [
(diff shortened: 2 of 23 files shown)


benzsevern:
## Scope grew: this PR is now **all of P4** (P4a + P4b)

P4b turned out to be small and inseparable from P4a, so it landed on the same branch rather than as a follow-up. The description above covers P4a; this is the rest.

### P4b — `backend="spark"`, and a third silent-degradation defect

**The seam the spec named was the wrong one.** `dedupe_df(backend=...)` selects a block *scorer* for a **local** dataset — that is what `ray`, `duckdb`, `datafusion`, `bucket` and `chunked` all do. `spark` is categorically different: the whole pipeline runs on a cluster, over data the cluster already holds. Accepting it on `dedupe_df` would mean shipping a driver-side frame out to the cluster and back, which is the opposite of the goal in §1 of the spec.

So the wiring is: `spark` is a **recognized** backend that declares intent, `_get_block_scorer` **raises** for it with the call that does work, and the public entry is `goldenmatch.spark.run_config_pipeline(spark_df, config)` (exported here).

**The defect found on the way.** `backend` was free text, and the dispatcher returns the default scorer for anything unrecognized — so:

```
backend="rayy"            -> ordinary single-box run, no warning
backend="sparkk"          -> ordinary single-box run, no warning
backend="totally-made-up" -> ordinary single-box run, no warning
```

A user who meant to distribute simply did not, and nothing said so. `backend` is now a closed set (`VALID_BACKENDS`) validated at config construction — where the name is still attached to the thing that named it — with an error naming the whole valid set rather than only rejecting.

The set is drawn from the values actually in use (`ray`, `duckdb`, `duckdb-backend`, `datafusion`, `bucket`, `chunked`, `polars-direct`, `spark`) and **every one is covered by a test**, because a closed set that rejected a real backend would break working configs.

### That's three defects of one shape

| where | the check that didn't fire |
|---|---|
| P4a | `null` vs `null` scored 1.0 — merged at every threshold |
| P4a | weighted denominator was per-matchkey, losing valid pairs |
| P4b | unknown `backend` ran single-box silently |

Plus P2's stale path filter, which left both Spark lanes silent on the code they gate. Same family throughout: **a check that exists and does not fire.**

### Verification for the added commits

19 further tests (no Spark needed). Full config/splink suite green (114 passed). Gate chain re-run after committing: docs staleness, claude refs, docs consistency, version consistency, workflow YAML, filter coverage, agent codemap, api surface — all OK.

Two pre-existing local failures confirmed **identical on clean `main`** and left alone: `test_frame_backend_differential.py[fused_covered]`, and `api_parity`/`ts_parity_freshness` (need the built TS workspace). A third, `check_docs_staleness`, turned out to be a Windows-only artifact — `subprocess(text=True)` decodes git output as cp1252 and `tuning.mdx` has a non-cp1252 byte, so the reader thread dies and `stdout` becomes `None` while the return code stays `0`. Passes under `PYTHONUTF8=1`; CI is Linux/UTF-8 so it is unaffected. Not fixed here, but worth its own issue.

benzsevern:
## P5 — Fellegi-Sunter on Spark

Landed on this branch too. P5 is unreviewable without P4a's `config_pipeline.py`, and this repo's CLAUDE.md records stacked PRs auto-closing on squash-merge (it cost rework twice on the TS parity wave), so stacking was the worse option. **This PR is now P4a + P4b + P5.**

Per P4a's finding, P5 is the phase the Splink cutover actually depends on — so this is the one that makes an imported Splink model executable at all.

### The model, re-expressed

```
sim_f     = scorer(a.f, b.f)              # null if either side missing
level_f   = threshold assignment over sim_f
weight_f  = match_weights[f][level_f]     # skipped when unobserved
total     = Σ weight_f
posterior = 1 / (1 + 2^-(prior_weight + total))
```

All arithmetic over a handful of per-field constants, so the model rides **inside the query plan** as inlined `CASE` expressions — no broadcast join, no lookup table, no UDF beyond the per-field similarity kernel the tier already had.

### Scoring distributes; training doesn't — and that's the existing semantics

`train_em` learns from a **sample** of blocked pairs (`n_sample_pairs=10000`), so the training set is small by construction at any dataset size. A Splink user brings weights trained elsewhere; a GoldenMatch user trains once and reuses via `model_path`. Neither needs a distributed E-step, and building one would **change** the numbers rather than reproduce them.

So the tier requires a trained model and says so, rather than training on demand and quietly pulling a sample back through the driver. That refusal is raised **before any Spark work is submitted** — a missing model should arrive as a config error, not as a failed distributed job.

### A threshold bug I wrote and caught here

`resolve_thresholds` returns `(link, review)` — **link first**. I destructured it the other way. The cut then happens at the *review* threshold, which is clamped `<= link` by construction, so every pair the one-box would have queued for a human gets **auto-linked** instead.

It doesn't look like a defect. It looks like a slightly generous run. Fixed, and pinned by a test that asserts the order.

### Both missing-value modes

`unobserved` (absence of evidence → level `-1` → zero bits) and `disagree` (level 0 → evidence against). The one-box makes this a config choice because whether missingness is informative depends on the data, so implementing one would be right for half its users. A lane test asserts the two produce **different** scores — otherwise honouring both proves nothing.

### Parity, and one property worth stating

Unit tests bind the pure statements to `core.probabilistic`'s own functions: level assignment vs `comparison_vector` across all four spellings (3 levels is `partial_threshold` then a **hard-coded 0.95**, not even spacing — worth knowing), weight summation vs `fs_regular_weight_sum`, posterior vs `posterior_from_weight`. Lane tests compare every candidate pair's posterior to the one-box under both modes.

One test states the cleanest available property: **a pair with no observed field scores exactly the prior match rate.** Zero bits must move the posterior not at all — and it doubles as a sharp check that `prior_weight` and `posterior_from_weight` are inverses. Verified: 0.05 in, 0.05 out.

`fs_pair_weight` also guards the trap `fs_regular_weight_sum` exists for: `weights[-1]` in Python is the *last* element — the highest-agreement weight — so an unobserved field would otherwise supply maximal evidence **for** a match.

### Refused loudly

Term-frequency (Winkler) adjustment, negative evidence, and model-backed `embedding`/`record_embedding` scorers — each contributes to the one-box score and has no expression here, so running without them returns a plausible number that is **not the model's answer**. Also refused: a model whose weights don't cover the matchkey's declared fields.

### Verification

76 unit tests passing (no Spark, so they run on every PR) plus the lane parity suite. Full gate chain re-run green after committing: workflow YAML, filter coverage, claude refs, docs consistency, version consistency, context budget, docs staleness, agent codemap (regenerated), native symbols, ruff.

### Still open

**P6** (zero-config on Spark). And FS **training** on Spark remains deliberately undone — I'd want a measured reason before changing what is currently a sample-based estimator with well-understood numbers.

benzsevern:
## P6 — Zero-config on Spark

`run_config_pipeline(df)` with no config now auto-configures, so the tier supports **both driver modes and the user picks per run**. `goldenmatch.spark.auto_configure_spark` is the standalone entry.

**This PR is now P4a + P4b + P5 + P6** — the spec's remaining phases, on one branch for the reason given earlier (P5 and P6 both build on P4a's module, and stacked PRs auto-close on squash-merge here).

### Auto-config already runs on a sample

`auto_configure_df` profiles and runs blocking → score → cluster on a stratified sub-sample — the same structural fact that made P5's EM work. So sampling is the existing mechanism, not a concession, and this reuses the one-box controller unchanged: every heuristic and health signal stays identical.

### What is *not* already handled

The controller infers dataset size from the frame it is handed:

```python
# autoconfig_controller.py:609
n_rows = _to_frame_gate(df).height
```

and uses that number for two things that must see the full population:

**1. The confidence gate.** `n_rows >= REFUSE_AT_N (100_000)` with a RED config raises `ControllerNotConfidentError`. Hand it a 20k sample drawn from 500M rows and the gate reads **20k** — so it never fires. The run that most needs the refusal is precisely the one that skips it.

I verified this against the real function rather than reasoning about it: a 200-row frame declaring 99,999 rows commits a **RED config with only a warning**, because it sits below the threshold. A 500M-row dataset seen as 200 rows would behave identically.

**2. Chao1 cardinality extrapolation.** The controller's own comment says this "depend[s] on this being the FULL data count, not the sample size". At sample scale a real mid-cardinality column (zip) looks near-unique — and near-unique columns get chosen as blocking keys, which on the full data produces blocks of one and finds nothing.

`auto_configure_df(n_rows_full=…)` addresses only (2), and only for the v0 heuristic. It does not move the controller's gate.

So `spark/autoconfig.py` applies the scale check **itself**, against the true `count()`, before handing anything over. `allow_large=True` is the explicit opt-in — a config chosen from 20k rows and applied to 500M is a real risk the caller should accept knowingly. The threshold is **imported** from the controller rather than duplicated, so the two gates can't drift.

### Two smaller decisions

**Bernoulli `sample`, never `limit`.** Partitions are usually ordered by ingestion, source or date, so a `limit` sample of a partitioned table is a sample of its oldest rows, or of one source. Profiling that and calling it a dataset profile is how a column looks unique when it's unique only within one partition. A test asserts `limit` is never called.

**Provenance is returned** (sample size, full count, fraction, seed), because a config whose origin isn't recorded gets treated as though someone chose it deliberately.

### Not done, deliberately

**Distributed profiling.** Everything here samples to the driver. Profiling on the cluster would be a different estimator with different numbers, and should be justified by a measurement rather than by symmetry with the rest of the tier.

### Verification

87 unit tests (no Spark — they run on every PR) plus the lane suites. Full gate chain green after committing: workflow YAML, filter coverage, claude refs, docs consistency, version consistency, context budget, docs staleness, agent codemap (regenerated), native symbols, ruff.

One test-authoring note, since it nearly produced a false pass: my fake DataFrame initially sampled against the rows it *held* rather than the rows it *claimed*, so a "500M-row" fixture returned 1 row and every sample-size assertion became vacuous. Fixed to model Spark faithfully.

PR #2502, opened 2026-08-12
feat(rust): score-cabi — a C ABI over score-core, for a JVM binding

benzsevern:
First step toward running the scoring kernel **inside the Spark executor JVM** instead of a forked Python worker.

Today the tier reaches Rust via `pandas_udf`: an Arrow IPC hop, a Python interpreter in the loop, and the whole executor-environment shipping apparatus from P1 that exists *only* to feed that worker. Executors are JVMs, and a JVM can call Rust directly — through a C ABI, which is what both Panama (FFM) and JNI consume. This is that seam.

It's also the repo thesis applied once more: `score-core` is already pyo3-free and already backs the Python extension, the DataFusion FFI UDFs and the WASM build. A JVM binding is **the next host binding over one kernel**, not a fourth implementation to keep in step.

## Buffer ABI, not JSON

The sibling C ABIs here (`goldenprofile-cabi`, `goldengraph-cabi`) marshal JSON — right for their request/response shapes, wrong for this one. The point is removing a per-batch marshaling cost; serializing a batch to JSON trades one for a larger one.

So the boundary is **Arrow's string layout**: `int32 offsets[n+1]` plus packed UTF-8, exactly what `StringArray` and Spark's UTF8 columnar vectors already hold. A host passes an address. Nothing is copied, nothing is allocated across the boundary, and `out` is caller-allocated — so there's no free function and no way to leak.

## Nulls are deliberately absent

No validity-bitmap parameter. Null semantics are a policy decision belonging to the caller, and putting them in the kernel has already cost this project once: substituting `""` for a missing value made null-vs-null score a perfect **1.0**, so two records whose only shared evidence was a shared *absence* merged at every threshold.

The header says so at length, because the next person to add a bitmap parameter will think they're being helpful.

## Measured cross-language parity

Rust-internal parity is a unit test. The claim that matters is **Python ↔ C across the FFI boundary**, probed over 4000 generated pairs × 99 thresholds:

| scorer | max\|C − Py\| | exact | threshold flips |
|---|---|---|---|
| `jaro_winkler` | 0.000e+00 | **4000/4000** | 0 |
| `levenshtein` | 0.000e+00 | **4000/4000** | 0 |
| `token_sort` | 1.110e-16 | 3530/4000 | 0 |

Two of three are **bit-exact**. Zero threshold flips across **396,000 comparisons** — which is the P3 test that matters: *a tolerance is fine for a score you report, not one you threshold.*

The `token_sort` delta is **not the kernel**. Python's `token_sort_ratio` returns 0–100 and callers divide by 100; that division is where the ULP enters. So it's **pre-existing** — the same delta `spark/scorers.py::_pure_scores` already carries — not something this crate introduces.

I left it alone deliberately. `score-core` exposes `token_sort_normalized_ratio`, and switching Python onto it would change shipped scores — that needs its own justification and gate run rather than riding along here.

## In the blocking lane

`goldengraph-cabi` is built in an informational workflow. This one goes in `ci-required`'s `rust` job, because it wraps the canonical scorer dispatcher and its whole claim is cross-binding identity — **an unbuilt binding cannot make that claim.**

The cdylib is built as well as tested: `cargo test` links the rlib, so a symbol that fails to export would pass the tests and ship a library no host can load. Verified locally — all three symbols present in a 406 KB dll.

The ABI also reports its version and scorer-id count, so a host fails loudly on a skew instead of silently scoring an unknown id as `0.0` — the "new kernel symbol, older caller" trap this repo paid for in #688.

## Scope

9 unit tests: parity with `score_one` across ids 0–8, multi-byte offsets (the classic Arrow bug — treating byte offsets as char indices truncates every non-ASCII value), and refusal of null pointers, decreasing offsets, out-of-range offsets, invalid UTF-8, and negative lengths.

Worth noting: `score_one` dispatches ids **0..=14**, so this ABI exposes all 15 scorers. The Spark tier's current limit of 4 is a restriction of its Python config surface, not of the kernel.

## Next, and the open question

The JVM side — a jar that loads the `.so` and exposes a Spark UDF. Before designing it, one unknown decides its shape: **the tier is Spark Connect**, which deliberately doesn't expose Catalyst, so a custom `Expression` can't be registered from a Connect client. `addArtifact` accepts jars and PySpark has `registerJavaFunction`; whether that combination works over Connect is a cheap P0-style probe, and it's the difference between a drop-in optimization and something that needs installing on the customer's cluster.


benzsevern:
## Probe result: **a Java UDF IS registrable over Spark Connect**

Run [31609185942](https://github.com/benseverndev-oss/goldenmatch/actions/runs/31609185942) — pyspark 4.2.0, Temurin 17, `remote=local[*]`:

| | path | detail |
|---|---|---|
| **PASS** | `addArtifact(jar)` | jar accepted by the session |
| **PASS** | `spark.udf.registerJavaFunction` | registered **and called** → `[1.0, 0.0]` |
| FAIL | `CREATE TEMPORARY FUNCTION … AS` | `AnalysisException [NO_HANDLER_FOR_UDAF]` |
| **PASS** | CONTROL: python udf | `[1.0, 0.0]` |

**So the JVM binding can be a drop-in.** Ship the jar per session via `addArtifact`, exactly as P1 ships the Python env — the customer's cluster needs **nothing installed**. That was the open question in this PR's description, and it's the one that decided the shape of the work rather than whether to do it.

The single failure isn't a Connect restriction: the SQL DDL path resolves the class as a *UDAF* and says so, pointing at the API that works. I kept it in the record rather than dropping it — a future reader would otherwise rediscover it, and if it ever starts working a scorer becomes callable from plain SQL, which is a separate and useful surface.

**The control passing is what makes that FAIL trustworthy.** Without it a broken session would have produced a confident "Connect can't do this" when the truth was "nothing ran at all" — the probe exits `INCONCLUSIVE` rather than `FAIL` in that case, deliberately.

The result is recorded in the committed harness (`scripts/probe_jvm_udf_connect.py`), not just in this thread, and the workflow takes a `pyspark_spec` input so the answer can be re-checked per Spark version.

### Two self-inflicted failures on the way, both worth knowing

- **A shell comment inside `run:` is still expression-interpolated.** My fix for a GitHub expression error contained a literal empty `${{ }}` to *explain* the error — and that was the next parse error.
- **`ls | head` under `set -o pipefail` fails the step.** `head` closes the pipe, `ls` takes SIGPIPE, pipefail propagates it. A diagnostic line that existed to print three filenames killed the build.

`check_workflow_yaml` caught neither, because it validates YAML parsing and duplicate keys — not GitHub expression syntax. That's a real gap in a gate this repo relies on, and it's the same family as the rest of this arc: a check that exists and doesn't fire. Not fixed here; worth its own issue.

### Next

The real JVM binding: a jar that loads the `score-cabi` `.so` via Panama/FFM and exposes it as a `UDF2<String,String,Double>`, plus the Python side calling `registerJavaFunction` after `addArtifact`. The probe already proves the registration and the argument shape; what's left is the FFM call and getting a stable address out of Spark's `UTF8String` without a copy.

benzsevern:
## Array-shaped probe: **works, at n=10,000 in one call**

Run [31611464914](https://github.com/benseverndev-oss/goldenmatch/actions/runs/31611464914):

| | path | detail |
|---|---|---|
| **PASS** | ARRAY arg type (observed) | `scala.collection.immutable.ArraySeq$ofRef` |
| **PASS** | ARRAY-shaped UDF (batched) | one call scored 3 pairs → `[1.0, 0.0, 1.0]` |
| **PASS** | ARRAY-shaped UDF at n=10,000 | **10,000 scores in one call**, all correct |

Plus the earlier rows unchanged (jar delivery, `registerJavaFunction`, control).

**This is the result that makes the native path worth building.** Connect only permits row-shaped UDFs, so without batching every pair costs an FFM downcall. With it, one downcall covers 10,000 pairs — the overhead stops mattering.

## The finding that earned the probe

Spark passes **`scala.collection.immutable.ArraySeq$ofRef`** — a *Scala* Seq, not a `java.util.List`.

Declaring `java.util.List` is the obvious guess, and it's what the Spark Java UDF examples suggest. It throws `ClassCastException` at runtime — which reads as *"the array shape doesn't work"* when the shape is fine and only the declaration was wrong. That's precisely the false negative I'd have shipped by reasoning instead of measuring, and it would have killed the batching idea on a wrong premise.

Useful consequence for the real binding: `ofRef` wraps an `Object[]`, so we can reach the backing array rather than iterating a Scala collection.

## What is still not proven

**Zero-copy.** Nothing here hands the UDF an Arrow buffer. Arrow Java exists and is on Spark's classpath — you were right to push on that — but the entry points that would expose it (`ColumnarBatch` / `ArrowColumnVector`) sit behind Catalyst, which Connect doesn't expose. So the array path amortises the **call**; it still copies the strings.

That's recorded in the harness rather than left implied, because "batching works" could easily be read as "zero-copy achieved", and they're different claims.

## Where that leaves the design

```
UDF2<ArraySeq, ArraySeq, List<Double>>   -- proven, batches to 10k+
    -> Object[] backing array            -- cheap, no Scala iteration
    -> pack into offsets + UTF-8 bytes   -- one copy, off-heap
    -> FFM downcall into score-cabi      -- one per batch
```

The one copy is the honest cost. Whether it matters is a measurement, not an argument — and it's the same measurement I flagged earlier as missing for the tier generally. I'd build the copying version, measure it against the current `pandas_udf` path, and only chase zero-copy if the copy shows up in the profile.

benzsevern:
## J0 landed on this branch

Ship a scorer jar to a Connect session and register it. **This PR is now score-cabi + the probe + J0.**

### A correction to the spec I wrote an hour ago

The spec said J0 ships *"a pure-Java implementation"*. That was wrong, and I'd have built the thing this whole arc argues against: a Java jaro-winkler is a **fourth implementation** of a kernel that already exists once in Rust and is reached from Python, WASM and DataFusion. "Temporary" implementations that produce plausible numbers don't stay temporary.

So J0 implements **`exact` and nothing else**. String equality isn't an algorithm — it's identical by inspection in any language, so there's nothing to diverge. It's a real score-core scorer (id 3), which makes this a genuine subset of the kernel rather than a stand-in. Every other scorer is refused loudly, with a message naming where the kernel arrives.

### Shape

| class | role |
|---|---|
| `GoldenScorer` | interface, **Spark-free** |
| `ExactScorer` | id 3 only; refuses the rest |
| `SeqCoercion` | **Spark-free**; the piece most likely to be wrong |
| `GoldenScoreUdf` | the **only** class touching Spark — coerce, delegate, box |

Keeping the decisions in Spark-free classes means they're unit-tested with no Spark on the classpath: **13 assertions, run locally** via a `main()` that exits non-zero. No JUnit — a test framework means a dependency resolver, and this package deliberately has no build system. Plain `javac` + `jar`, as the probe proved.

`--release 17` is load-bearing: Spark supports 17/21/25 and the runner's JDK is only one of them. Without it the jar runs in CI and dies on a 17 executor with `UnsupportedClassVersionError`.

`SeqCoercion` encodes the measured fact, not a guess — `ArraySeq$ofRef`, reaching its backing `Object[]` rather than iterating, with a reflective `size()`/`apply(int)` fallback so the jar isn't pinned to a Scala version.

### Python side

`goldenmatch/spark/jvm.py` ships and registers. Failures raise a typed `JvmScorerUnavailable` so callers can fall back to the `pandas_udf` path rather than catching bare `Exception` — a distributed run must not fail because one executor couldn't load a library.

Scorer ids are duplicated from score-core deliberately (this path must work with no compiled kernel) and a test pins them against the loader's map.

### CI

The jar is built and self-tested **inside** the `spark_connect` lane, which already has the JDK and Spark jars. A separate job would mean building it twice or passing an artifact — both add a way for the tested jar and the shipped jar to differ.

`packages/jvm/**` joins the `spark` path filter **and** `check_filter_coverage`'s REQUIRED table. Without the filter, a Java edit would never trigger the lane that builds it — the same staleness the P2 rename introduced, one language over. Verified RED against the un-patched filter first.

### Verification

13 Java self-tests (local, JDK 25, compiled `--release 17`) · 12 Python unit tests · 6 lane tests written, skipping locally without Spark · full gate chain green: workflow YAML, filter coverage (RED then GREEN), claude refs, docs consistency, version consistency, context budget, docs staleness, agent codemap, ruff.

One lane test is aimed squarely at J1's risk: a fixture where **every position has a distinct expected value**, so a rotation or reversal changes the result. Counting 1.0s would pass on a misalignment; position won't.

### Next

**J1** — the plan reshape: group pairs into arrays, score once, explode back, gated by pair-set parity against the row-shaped path. Still no native code, on purpose. That's where the alignment hazard actually lives.

PR #2508, opened 2026-08-12
fix(spark): the tier is arrow-native — remove pandas from every UDF and executor

benzsevern:
**pandas was a runtime requirement on every executor** of a package that doesn't declare it as a dependency, in a repo that evicted polars and made pyarrow the hard dep. The environment P1 ships had to `pip install pandas` explicitly, and the comment I wrote there documented the problem instead of fixing it:

> *"goldenmatch does NOT depend on pandas, but `pandas_udf` needs it in the WORKER"*

Spark has first-class Arrow UDFs, so there was never a need. All **8 UDFs** across `scorers` / `golden` / `identity` / `config_pipeline` move to `arrow_udf` over `pa.Array` — same shape, and pyarrow is already a hard dependency.

The pattern came from the June Sail work. **I added two more in `config_pipeline.py` during P4**, by matching the surrounding code — the wrong instinct when the surrounding code is the defect. I also swept the arrow seam at the start of this session and never looked at the tier's UDF layer.

## What falls out beyond dropping a dependency

**Nulls stop being NaN.** A pandas Series of an all-null string batch infers float64, so iterating yielded NaN — and NaN is *truthy*, so `x or ""` didn't rescue it and a float reached the scorer as `TypeError: 'float' object is not subscriptable`. That was a real crash I fixed defensively three commits ago. A `pa.Array` carries a validity bitmap and `to_pylist()` gives `None`, so the **bug class is gone rather than guarded**. `pd.isna` disappears from `identity.py` for the same reason.

**The batch is already the shape the kernel wants.** A `pa.Array` for a string column *is* the offsets+data layout `score-cabi` takes — so the Python path can hand buffers to Rust with no conversion. I said that was out of reach.

**Two variadic `*cols` + `pd.concat` UDFs became one struct argument.** `arrow_udf`'s variadic form isn't in the documented contract, and a struct is self-describing: `to_pylist()` yields a dict per row with field names attached.

**`_arrow.arrow_udf` raises** when pyspark can't provide it rather than falling back to `pandas_udf`. A fallback would be worse than the original problem — it would pass in test and quietly reinstate the dependency in production.

## The ratchet

- `pip install ... pandas` is gone from the shipped executor env, and the build now **fails if pandas is importable there**.
- `tests/test_spark_no_pandas.py`: no `pandas_udf(`, no `import pandas`, in any tier module; the arrow helper must not fall back; pandas must not appear in `pyproject.toml`.

It matches **usage** rather than the word, so the comments recording this history survive.

Without the ratchet the dependency returns the first time somebody reaches for `pandas_udf` out of habit, and nothing notices — the env simply grows.

## Note

This branch also carries the bounded-batch commit from #2507 (already merged); only the arrow-native change is new here.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3091,19 +3091,32 @@ jobs:
           set -euo pipefail
           python -m venv /tmp/gmenv
           /tmp/gmenv/bin/pip install --upgrade pip
-          # goldenmatch does NOT depend on pandas, but `pandas_udf` needs it in
-          # the WORKER -- that is why P0 logged `No module named 'pandas'` 13
-          # times alongside the goldenmatch ones. pyarrow is goldenmatch's own
-          # dep but is named here so the worker contract is explicit rather than
-          # transitive. NOT -q: a silent pip failure is how you ship an env that
-          # unpacks fine and imports nothing.
-          /tmp/gmenv/bin/pip install ./packages/python/goldenmatch pandas pyarrow
+          # NO PANDAS. The tier's UDFs are `arrow_udf` over `pa.Array`, so the
+          # executor needs pyarrow (goldenmatch's own hard dependency) and
+          # nothing else. Shipping pandas was an UNDECLARED runtime dependency
+          # of a package that does not depend on it -- in a repo that evicted
+          # polars and made pyarrow the hard dep. pyarrow is named explicitly so
+          # the worker contract is stated rather than transitive. NOT -q: a
+          # silent pip failure is how you ship an env that unpacks fine and
+          # imports nothing.
+          /tmp/gmenv/bin/pip install ./packages/python/goldenmatch pyarrow
           echo "--- shipped env contents ---"
           /tmp/gmenv/bin/pip list | grep -iE "goldenmatch|goldenphonetic|pandas|pyarrow"
           # cd out of the repo: its ROOT DIRECTORY is itself named `goldenmatch`,
           # so a check run from there can import the source tree instead of the
           # installed package and pass while the env is empty.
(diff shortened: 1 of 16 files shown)


benzsevern:
## CI green — and it settled the pysail question

`spark_connect` (blocking) passes: **255 tests**, and the executor probe now reports

```
goldenmatch: True   strsim: True   pyarrow: True   native_kernel: True   pandas: False
```

That's the change working end to end: the shipped executor environment no longer carries pandas, and everything else still imports.

One assertion had to be inverted — P1's `test_executor_can_import_goldenmatch` asserted `pandas is True`, which was correct when the tier used `pandas_udf`. It now asserts pandas is **absent**, with the note that `goldenmatch` / `strsim` / `pyarrow` remain asserted present, so an empty environment would fail those rather than pass this.

## The `sail` lane now fails, and that is a finding

pysail pins **pyspark 3.5.9**, which has no `arrow_udf`. So `_arrow.arrow_udf` raises — exactly as designed, rather than silently falling back to `pandas_udf` and reinstating the dependency.

**The legacy Sail backend can no longer run the tier.** That resolves the question I deferred in P2: I kept pysail as an advisory cross-check because it had caught a real regression during P2a. It can't exercise the UDF paths any more, so most of that value is gone.

The lane is `continue-on-error` and absent from `ci-required`, so this doesn't block — but a permanently-red advisory lane is noise, and noise trains people to ignore lanes. That's a call worth making deliberately rather than letting it rot:

- **Retire the `sail` lane** — the P2 plan's original position, now justified by more than tidiness.
- **Or** narrow it to the tests that don't touch `arrow_udf` (connectivity, WCC, feature gates), keeping a partial cross-check.

I'd take the first. The second keeps a backend that can't run the thing it's meant to cross-check.

Not doing either in this PR — it's a separate decision and this one is already green.

PR #2516, opened 2026-08-13
feat(spark): J2 -- run the Rust kernel inside the executor JVM (JNI)

benzsevern:
## What

J0's jar could score `exact` and nothing else. J2 removes that restriction by **calling** the Rust kernel from the executor JVM rather than reimplementing it:

```
NativeScorer (Java) -> JNI -> score-cabi -> score-core::score_one
```

The same dispatcher already behind pyo3 `native`, `datafusion-udf` and `score-wasm`. A fifth host, not a fifth implementation.

## What crosses the boundary

Not `String[]`. That signature costs a JNI transition and usually a copy **per string** -- 20,000 of them per 10,000-pair batch, to avoid one, which undoes the reason J1 reshaped the plan into batches at all.

So `Utf8Batch` flattens to Arrow's string layout (`int32` offsets + packed UTF-8) and the JNI layer pins five primitive arrays: a fixed cost per **batch**, independent of how many pairs it holds. That is the exact layout `score-cabi` was designed around, so nothing is converted at the seam.

Pinning is deliberately **non-critical**. `GetPrimitiveArrayCritical` saves one `memcpy` per array per batch and suspends GC to do it -- five memcpys against ten thousand string comparisons is not where the time goes.

Nulls stay a **host** decision, as score-cabi's header states. Null slots cross as zero-length slices and their scores are discarded via the presence mask. Wasted work, chosen over an index remapping that could silently attach a score to the wrong pair.

## One jar, nothing installed

The `.so` rides **inside the jar** and is extracted per JVM, so `addArtifact` still delivers everything: no `java.library.path`, no library placed on every executor. Replacing the packed-virtualenv apparatus with a different deployment apparatus would have been no progress.

CI builds `linux-x86-64` only (what a Spark executor runs). The loader already resolves the other os/arch names, so adding one is a build change, not a code change.

## The fallback announces itself

If the library will not load, the jar falls back to the `exact`-only scorer -- a distributed job must not die because one executor could not `dlopen` something.

But a fallback nobody can see is worse than none: the query still returns numbers, from a narrower path, and every version string still looks right. This repo has lost real time to exactly that shape. So:

- `golden_score_impl` reports what an **executor** resolved and why.
- `NativeSelfTest` and the lane parity test both **assert** it resolved `NativeScorer`.
- The probe takes a column it ignores. A zero-arg deterministic UDF is foldable (`children.forall(_.foldable)` is vacuously true), so Catalyst would evaluate it on the **driver** and answer the wrong question perfectly.

Load-time skew gate: ABI version and scorer-id count are checked before use, and an id outside the loaded kernel's range is refused rather than falling to `score_one`'s catch-all and returning a confident `0.0`.

## A real divergence, found and fixed

Building the parity gate turned one up. For a `token_sort` pair whose exact answer is 12/13:

| path | value |
|---|---|
| kernel (`score_one`) | `0.9230769230769231` |
| the tier's Python floor | `0.923076923076923` |

Self-inflicted: the floor called the 0-100 `token_sort_ratio` and divided by 100, and `(x * 100) / 100` is not `x`. One ULP.

One ULP is ignorable for a score you **report**. This tier **thresholds** these -- which is the argument `scorers.py`'s own docstring already makes about f32 narrowing. Fixed by splitting `strsim.token_sort_similarity` (`[0, 1]`) out of `token_sort_ratio` (0-100, byte-identical to before; rapidfuzz parity intact).

**Not bundled:** the same `/ 100.0` round trip is written at ~8 other call sites (`core.scorer`, `core.boost`, `core.explainer`, `backends.score_buckets`, `native`, `backends.datafusion_backend`) and carries the same deviation. Moving those changes thresholded scores repo-wide against pinned expectations in several suites. Flagged in `strsim`'s docstring for its own change and its own gate run.

## Verification

Run locally, end to end, before CI -- library built, extracted from the classpath, `dlopen`'d, JNI boundary crossed:

- `NativeSelfTest`: 21 assertions green, including 10,000 pairs in one call, batch alignment, multi-byte identity, and `null`-vs-`null` returning `null` rather than `1.0`.
- **All 36 (scorer x pair) values byte-identical** between the JVM and the tier's Python scorer, after the `token_sort` fix. Before it, 2 of 36 differed -- which is how the bug was found.
- `cargo test` / `clippy -D warnings` / `fmt` clean on `score-jni`; 138 Python tests green across the Spark tier and `strsim` suites.

## Gates added

- `rust` job: `cargo test` / `clippy` / `build --release` for `score-jni`. In the **blocking** job, not an informational lane -- a change to `score-core` or `score-cabi` that broke this binding would otherwise compile nowhere until someone touched a Spark file.
- `spark_connect`: build the library, assert the JNI symbols are **exported**, assert the `.so` is **in the jar**, run `NativeSelfTest`, then the Spark parity gate.
- `filters.yml`: the lane now watches `score-jni`, `score-cabi`, `score-core` and `strsim.py` -- the kernel and the oracle it is measured against.
- `javac -encoding UTF-8`: sources carry multi-byte parity fixtures and javac otherwise reads the platform default.

## Not in scope

No routing change. Nothing selects the JVM path by default -- this makes it *correct and gated*, not *on*. The `batched_jvm` OOM from the bench arc is still open and unexplained; the next experiment there is a bisect stage that calls the **shipped** `score_pairs_batched` rather than a lookalike.

### DIFF
--- a/.github/filters.yml
+++ b/.github/filters.yml
@@ -902,6 +902,17 @@ spark:
   # lane, so a Java edit that never triggered that lane would ship untested --
   # the same staleness the P2 rename introduced, one language over.
   - 'packages/jvm/**'
+  # J2: the jar carries the Rust kernel, reached through score-jni -> score-cabi
+  # -> score-core. The lane BUILDS all three, and the cross-language parity gate
+  # (test_spark_jvm_native_parity.py) is the only place a divergence between the
+  # JVM's score and Python's would surface. A kernel edit that skipped this lane
+  # would ship a jar nothing had compared against.
+  - 'packages/rust/extensions/score-jni/**'
+  - 'packages/rust/extensions/score-cabi/**'
+  - 'packages/rust/extensions/score-core/**'
+  # The tier's own Python scorer is the ORACLE that gate compares against, so a
+  # change to it moves the reference the JVM is measured by.
+  - 'packages/python/goldenmatch/goldenmatch/core/strsim.py'
 action:
   - 'packages/actions/**'
 dbt:
--- a/.github/workflows/bench-spark-scoring.yml
+++ b/.github/workflows/bench-spark-scoring.yml
@@ -35,6 +35,14 @@ on:
         description: "Runner label"
         required: false
         default: "large-new-64GB"
+      driver_memory:
+        description: >-
+          Heap for the local[*] JVM, which IS the executor. Set EQUALLY for both
+          arms. The batched path materialises each group as an array in JVM heap,
+          so an unset ceiling (Spark's 1g default) measures a configuration
+          rather than a design -- and that is what OOM'd the JVM arm.
+        required: false
+        default: "12g"
 
 permissions:
   contents: read
@@ -64,15 +72,33 @@ jobs:
       # Same build the spark_connect lane does, and for the same reason:
       # `--release 17` so the jar runs on Spark's oldest supported JDK rather
       # than only on whatever the runner happens to have.
+      # J2: the kernel's JNI door, built BEFORE the jar because it ships inside
+      # it. Without this step the jar loads, falls back to J0's `exact`-only
+      # scorer, and this bench silently measures the wrong thing against a
+      # native row_python arm. The arm asserts on it too -- belt and braces,
+      # because a silent fallback here produces a NUMBER, and a number is
+      # believed.
+      - name: Build the native scorer library (JNI)
+        run: |
+          set -euo pipefail
+          cargo build --release \
+            --manifest-path packages/rust/extensions/score-jni/Cargo.toml
+          ls -lh packages/rust/extensions/score-jni/target/release/libgoldenmatch_score_jni.so
+
       - name: Build the JVM scorer jar
         run: |
           set -euo pipefail
           SPARK_JARS="$(.venv/bin/python -c 'import pyspark, os; print(os.path.join(os.path.dirname(pyspark.__file__), "jars"))')"
           JVM=packages/jvm/goldenmatch-spark
-          mkdir -p "$JVM/build/classes"
-          javac --release 17 -cp "$SPARK_JARS/*" -d "$JVM/build/classes" \
+          mkdir -p "$JVM/build/classes/native/linux-x86-64"
+          javac --release 17 -encoding UTF-8 -cp "$SPARK_JARS/*" -d "$JVM/build/classes" \
             "$JVM"/src/dev/goldensuite/spark/*.java
+          cp packages/rust/extensions/score-jni/target/release/libgoldenmatch_score_jni.so \
+            "$JVM/build/classes/native/linux-x86-64/"
           jar --create --file "$JVM/build/goldenmatch-spark.jar" -C "$JVM/build/classes" .
+          jar --list --file "$JVM/build/goldenmatch-spark.jar" \
+            | grep -qx "native/linux-x86-64/libgoldenmatch_score_jni.so" \
+            || { echo "::error::the native library is not in the jar; the JVM arm would bench the J0 fallback"; exit 1; }
 
       # P1's executor environment. The row_python path forks a Python worker,
       # which has NO access to the client's site-packages -- the first run of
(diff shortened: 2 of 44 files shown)


benzsevern:
## Bench result: the JVM path loses on throughput, and the kernel is not why

Like-for-like at last (both arms `jaro_winkler`, same `score_one` underneath), 200k rows -> 1.9M candidate pairs, `local[*]`, 12GB heap, 16 cpus, executor confirmed on `NativeScorer`:

| arm | median | how it reaches Rust |
|---|---|---|
| `row_python` pure floor | 2.361s | not at all (Python floor) |
| `row_python` **native** | **1.463s** | pyo3 wheel, forked Python worker |
| `batched_jvm` | **3.496s** | JNI, in the executor JVM |

The Rust kernel is worth **1.61x** over the pure floor (same UDF, same plan, same worker, one flag). The JVM path is **2.4x slower** than the Python-worker path it was built to beat.

### Where the JVM arm's time actually goes

Plan bisect, cumulative stages, same run:

```
3  groupBy/collect (NO UDF)        2.3s
4b + UDF, scores CONSUMED          2.5s     <- scoring: +0.2s
5  + arrays_zip/explode            3.7s     <- un-batching: +1.2s
6  + dedup_max                     4.0s
7  SHIPPED score_pairs_batched     3.3s
```

Stage 4b exists because stage 4 could not be trusted: it ends in `.count()`, which does not need the `scores` column, and Catalyst is entitled to prune a deterministic side-effect-free UDF. Stage 4 came in *below* stage 3, which reads as "scoring is free" and is equally consistent with scoring never running. 4b consumes every score without exploding, so it cannot be pruned.

With that: **scoring 1.9M pairs through JNI into Rust costs ~0.2s.** Roughly 9.5M pairs/sec. Spec section 1's claim about the kernel is real.

It is also ~6% of the arm's wall. Making the kernel infinitely fast turns 3.5s into 3.3s. **The JVM arm cannot win by scoring faster.**

The cost is the **un-batching** -- `arrays_zip` + `explode` back to 1.9M rows, 6x the scoring cost. And the batching is not optional: Spark Connect only permits row-shaped UDFs, so reaching a native kernel from the JVM *requires* group -> array -> score -> zip -> explode. The row-shaped Python path never groups at all, so it has no un-batching to pay for. Its entire wall (1.463s) is less than the batched path's plan overhead before any scoring.

### What this does and does not settle

**Settled:** the throughput argument for the JVM path, negatively, on Spark Connect. I said before building J2 that I would stop claiming it was faster; this is the number behind that.

**Untouched:** the two reasons this work was actually asked for -- one kernel with many host bindings, and a jar instead of a Python environment on every executor. The `row_python` arm needs a packed virtualenv shipped to the cluster. The JVM arm needs nothing installed. That is a deployment difference these timings do not capture.

**Caveat, stated because it cuts against the JVM arm:** the bench runs `threshold=0.0`, so nothing is filtered before the explode -- the maximally unfavourable case for batching. A realistic threshold could be applied inside the batch, shrinking the explode by whatever the filter removes. That is a real optimisation and it is unmeasured; I have not claimed it.

Also fixed here: the comparison printed `batched_jvm (exact)` while running jaro_winkler (stale hardcoded label), and printed `native_resolved=False` beside the JVM arm -- that field is the *driver's* pyo3 state and reads as "the kernel did not run" when the same run had asserted it did.

Bench runs: [31656227516](https://github.com/benseverndev-oss/goldenmatch/actions/runs/31656227516) (bench), [31656840690](https://github.com/benseverndev-oss/goldenmatch/actions/runs/31656840690) (attribution).

benzsevern:
## Inventory: what actually runs with NO Python on the executors

Measured, not read off the source. An executor env is shipped that is deliberately **empty** (a bare virtualenv, asserted unable to import goldenmatch), so anything needing a Python worker fails exactly as it would on a cluster with no cluster-side install. Omitting the env would prove nothing -- under `local[*]` the worker forks from the driver's interpreter, which has goldenmatch, so everything would pass and report a capability the tier does not have.

**jar-only today: 3 of 7.**

| probe | result |
|---|---|
| pure Spark SQL (block join) | WORKS -- 6 candidate pairs |
| clustering (connected components) | WORKS -- 6 labelled nodes, 6 components |
| **JVM scoring (J2, via the jar)** | **WORKS -- 6 scored pairs, max=0.9429** |
| Python scoring (the shipped path) | NEEDS a worker |
| normalization (transform chain) | NEEDS a worker |
| survivorship (golden merge) | NEEDS a worker |
| identity (record fingerprints) | NEEDS a worker |

### What this settles

**J2's premise holds.** Scoring runs on an executor with zero Python present, through the jar, returning real scores. That is the deployment claim demonstrated rather than argued -- and it is the one the throughput bench could not speak to.

**`clustering.py` was already there.** It defines no UDFs; connected components is expressed in Spark SQL. It needed nothing from this arc.

**The gap is three subsystems**, not one: normalization (`config_pipeline._transformed`), survivorship (`golden.make_merge_udf`) and identity (five `arrow_udf`s in `identity.py`). Until those move, a cluster still needs a packed virtualenv, and the jar removes one dependency rather than the requirement.

### A correction

The first run of this inventory reported **6 of 7**, claiming normalization, survivorship and identity all worked without Python. That was wrong and impossible on its face -- all three define `arrow_udf`s.

Those probes ended in `.count()`. Counting rows does not need any particular column, so Catalyst pruned the UDFs, and "never ran" was reported as "works without Python". Every probe now collects its column and prints a sample value, so a pruned probe is self-evident: there is no value to print if nothing ran.

This is the same mistake as the batched-plan bisect's stage 4, found and written up a few hours earlier, then repeated in a new file. The control is what makes the harness trustworthy either way: Python scoring failed with a real `PythonException` from the worker in BOTH runs, so the empty env was always in effect -- only the probes were wrong.

Run: [31658365757](https://github.com/benseverndev-oss/goldenmatch/actions/runs/31658365757).

Issue #2518, opened 2026-08-13
FS threshold refit stays default-off: the cluster-shape accept criterion is a single-outlier statistic


benzsevern:
Split out of #2483. This is the blocker between the FS link cutoff being *reported* (PR #2517) and it being *chosen from the data*.

## Why this matters

`GOLDENMATCH_FS_REFIT_THRESHOLD` already does what #2483 asks for: pick the link cutoff from the actual scored-pair distribution instead of the fixed 0.50 default, wired uniformly into every FS route via `_maybe_refit_link_threshold`. It is default OFF, and while it is off the reported symptom stands — that reporter's corpus links 94.4% of records into multi-member clusters, fusing 2,554 distinct names into one entity.

Measured on that corpus, the fixed default is far from the useful range:

| `link_threshold` | match rate | entities | largest entity |
|---|---|---|---|
| unset (0.50 default) | 94.40% | 2,553 | 2,554 names |
| 0.90 | 62.08% | 11,205 | 573 names |
| 0.95 | 37.55% | 16,272 | 547 names |

And the percentile branch of `compute_thresholds` would have picked ~0.95 on a comparable shape — the value that reporter found by hand.

## The defect

`_fs_refit_threshold_enabled`'s docstring records why it was reverted in #2387: turning it on cost **−0.0966 F1** on `ncvr_synthetic` (0.9828 → 0.8881), confirmed by single-variable A/B on one sha.

The cause is the cluster-shape guard's accept criterion, not a baseline needing a re-bless. The guard accepts when `max_candidate < max_default`. On `ncvr_synthetic` it commits `0.5000 -> 0.7000 (max cluster 5 -> 2, over-merge reduced)` — and **every gold cluster in that dataset is exactly size 2** (2,500 disjoint pairs). So the guard's own objective is satisfied while F1 falls, because `max_candidate < max_default` is a single-outlier statistic: it fixes the one or two genuinely over-merged clusters and is structurally blind to the true pairs the raised cutoff drops across the other ~2,500 correct size-2 clusters.

It works on `household_hardneg` (0.947 → 1.000) only because there over-merge *is* the dominant error mode.

This is the same shape as several defects fixed this week: a quantity that decides something is computed over the wrong population. The guard measures the tail and rules on the whole distribution.

## What re-flipping needs

From the docstring, and I agree with it after reading the code:

[...]

PR #2522, opened 2026-08-13
feat(probabilistic): turn the FS link-threshold refit on by default (#2518)

benzsevern:
## What and why

Follow-up to #2519 (which fixed the guard defect). Turns `GOLDENMATCH_FS_REFIT_THRESHOLD` on by default.

The FS path is otherwise non-iterated: it commits a fixed 0.50 link cutoff, which over-merges on shapes where a shared attribute co-blocks non-matches (distinct households sharing a surname; co-tenants sharing an address). With the refit on, the cutoff comes from the actual scored-pair distribution - a bimodality-gated valley, accepted only when it **both** reduces the largest cluster and strands under 1% of matched records as singletons. On the lever panel that recovers `household_hardneg` 0.9687 -> 1.0000 and `cotenant_hardneg` 0.4193 -> 0.9963, and is a no-op everywhere else. `GOLDENMATCH_FS_REFIT_THRESHOLD=0` is the kill-switch, byte-identical to the fixed cutoff.

## The evidence bar, and why it is this one

This default was flipped on once before (#2377) and reverted after five red nights (#2387). The lesson that failure established is specific: **validate on the panel that GATES the change, not the panel that motivated it.** #2377 rested on `ab_lever` + QIS while the gating panel went unrun.

So this flip was measured with that gate itself - `scripts.suggest_quality`, single-variable `=0` vs `=1` on one sha, native kernel present (`suggest_config` symbol asserted):

| tier | result |
|---|---|
| fast `gate` (FULL_DIST=0), local `=0` vs `=1` | **byte-identical on every metric**, both arms PASS |
| `ncvr_synthetic convergence_final_f1` | **0.9847 in BOTH arms** |
| fast `gate` in CI on this branch | PASS (2m49s) |
| heavy `gym-gate` (FULL_DIST=1) in CI on this branch | **PASS (15m17s)** |

Both CI tiers ran as [31661415506](https://github.com/benseverndev-oss/goldenmatch/actions/runs/31661415506) on `large-new-64GB`, with the flip live on the branch, and `gym-gate` measured with `=1` against a `gym_scorecard.json` blessed at `=0`.

0.9847 is exactly `ncvr_synthetic`'s `=0` value in the #2387 A/B, where default-on scored **0.8881**. Mechanically the flip is a no-op on that dataset because the valley detector no longer finds any candidate above 0.50 there at all, so the accept criterion is never reached - the #2519 guard fix and a shift in the scorer set the controller picks both point the same way.

`refit-lever` is green here too, now running the widened 4-dataset PR tier.

## Closing the two gates that failed to fire

A check that exists and does not run is what turned a catchable defect into a five-night red, so both holes are closed here rather than left for later:

1. **`bench-suggest-quality`'s push filter** listed only `suggest/**` and `scripts/suggest_quality/**`. A `probabilistic.py` edit ran **no suggester gate at all** - the mechanical reason #2377 went green on push while only the nightly reddened. Now includes `probabilistic.py` and `pipeline.py` (where `_maybe_refit_link_threshold` lives). Both are hot files and this fires the fast tier (minutes) on their pushes; that cost is deliberate - an absent check is strictly worse than a chatty one.
2. **`fs-lever-gate`'s PR tier** ran `person,household_hardneg,cotenant_hardneg` only. It now includes `ncvr_synthetic` - the dataset that *caught* the regression belongs on the per-PR tier, not only the nightly. 7.5k rows, ~30s/arm, so the "fast tier" claim still holds.

Fix 1 demonstrated itself on this PR: pushing this branch fired [31661340760](https://github.com/benseverndev-oss/goldenmatch/actions/runs/31661340760) on event `push`, which would not have happened before this change.

## Changes

- `probabilistic.py`: `_fs_refit_threshold_enabled` default `"0"` -> `"1"`; docstring rewritten around the A/B evidence and the two gate widenings.
- `.github/workflows/bench-suggest-quality.yml`: push filter gains the FS refit paths.
- `.github/workflows/fs-lever-gate.yml`: PR tier gains `ncvr_synthetic`.
- `test_fs_refit_threshold.py`: `test_flag_off_is_noop_both_forms` **encoded the old default** (it asserted an UNSET env is a no-op) and is replaced by two tests - `test_kill_switch_is_noop_both_forms` (`=0` byte-identical on both pair representations) and `test_default_and_opt_in_both_fire_on_over_merge` (unset env asserts the refit FIRES, so a silent revert to default-off fails the suite).
- `CHANGELOG.md`, `docs/context/code-patterns.md`, and the design spec - whose status line still read `DEFAULT-OFF` and would have been wrong the moment this lands.

## Testing

- `pytest tests/test_fs_refit_threshold.py` - **24 passed**.
- Post-flip focused suite over the FS / probabilistic / cluster / splink / autoconfig files - **1633 passed, 13 skipped, 1 failed**. The failure is `test_cluster_edges_df.py::test_datafusion_is_importable`, which fails identically on clean `main` here (optional dep not installed). This was the run most likely to surface fallout, since the flip changes behaviour for every FS run rather than sitting behind a flag.
- `scripts.suggest_quality gate`, `=0` vs `=1`, diffed: identical on every metric, both PASS. Both CI tiers green (table above).
- `ruff check packages/python/goldenmatch` clean; `check_context_budget.py` OK; `check_workflow_yaml.py` **121 files, no parse errors or duplicate keys** (this PR edits two workflows, so that gate matters here).

One note on the local runs: the heavy `gym-gate` could **not** be validated on the dev container - it was SIGKILLed (exit 137, OOM; 15 GB available against the 64 GB runner the gate is specced for). That is an environment limit, not a signal about this change, and a killed run is not evidence either way. It was dispatched to CI instead, which is where the PASS above comes from.

## Unrelated observation, filed separately

Both gate arms report `dblp_acm convergence_final_f1 = 0.1220` with `steps=0`, on a best-effort RED config (`stop_reason=BUDGET_TIME`). Identical in both arms, so unaffected by this change, and the dataset is `(NEW)` in the scorecard so it gates nothing. Filed as #2526.

## Checklist

- [x] Package `CHANGELOG.md` updated if behavior changed
- [x] Heavy `gym-gate` tier - PASS in CI
- [ ] `pre-commit run --all-files` - not run in full; ruff + context-budget + workflow-YAML + codemap gates run individually and are clean


benzsevern:
Validation status. Staying a draft until the heavy tier reports.

**Post-flip focused pytest: 1633 passed, 1 failed.** The failure is `test_cluster_edges_df.py::test_datafusion_is_importable`, which fails identically on clean `main` in this environment (optional dep not installed). No fallout from the flip across the FS / probabilistic / cluster / splink / autoconfig surface - which is the run that mattered most, since the flip changes behaviour for every FS run rather than sitting behind a flag.

**Correction to the PR body: the heavy `gym-gate` was NOT validated locally.** I claimed it was in flight; it was SIGKILLed - exit 137, OOM. This container has 15 GB and the gate is specced for the 64 GB runner. That is an environment limit, not a signal about this change, and a killed run is not evidence in either direction.

So it has been moved to where it belongs: dispatched on this branch as run [31661415506](https://github.com/benseverndev-oss/goldenmatch/actions/runs/31661415506). A `workflow_dispatch` is not a `push`, so that run executes **both** tiers - the fast `gate` and the heavy `gym-gate` - on `large-new-64GB`, with the flip live. Native kernel built and the `suggest_config` symbol asserted present. That is a better test than my local reconstruction anyway, because it is the actual gating infrastructure.

**Unplanned evidence that one of the two gate fixes works.** Pushing this branch also fired run [31661340760](https://github.com/benseverndev-oss/goldenmatch/actions/runs/31661340760) on event `push`. That only happens because `probabilistic.py` is now in `bench-suggest-quality`'s push filter. Before this PR, pushing an FS refit change triggered no suggester gate at all - which is the mechanical reason #2377 went green on push while only the nightly reddened. The hole is closing on its own PR.

Recap of what is already green: fast `gate` A/B byte-identical on every metric with both arms PASS (`ncvr_synthetic convergence_final_f1` **0.9847 in both**, against **0.8881** for default-on in the #2387 A/B); `refit-lever` passed on #2519 with the criterion fix in place; 24/24 refit tests; ruff, context budget, and workflow YAML (121 files) clean.

If the dispatched `gym-gate` comes back red, this does not ship and I will say so - shipping a flip on an unrun gating panel is precisely the #2387 mistake.

---

PR #2525, opened 2026-08-13
feat(spark): publish the jar, and route the kernels through the entry points

benzsevern:
Two things, both following from the same finding: the jar was unreachable from
outside this repo, and its kernels were unreachable from outside the module.

## 1. The jar can now leave the repo

It was built inside `ci.yml` and thrown away with the run. Nobody without a
checkout and a JDK could obtain it, which made "one file, nothing installed on
the cluster" untestable by anyone it was aimed at.

`publish-goldenmatch-spark-jar.yml` fires on a `goldenmatch-spark-v*` tag:
builds both architectures, self-tests each on its own hardware (aarch64 on a
real ARM runner, not cross-compiled), runs the assembled jar, attaches it to a
GitHub Release. A Release rather than Maven Central deliberately -- the jar is
not a compile dependency anybody declares, it is a file handed to `addArtifact`
at runtime, and a URL is the whole interface.

Draft, then attach, then publish: this repo has immutable releases enabled,
which seals assets at publish time and permanently burns a tag consumed by one.
`dry_run` exercises the whole pipeline without spending a tag.

**On building the jar twice.** Publishing forces a second build, which is the
"tested jar vs shipped jar" gap the CI lane's single-job build exists to close.
So the build moved into `scripts/build_spark_jar.sh` and the Spark-free
self-test into `scripts/spark_jni_selftest.sh`, both called by both workflows:
separate builds that cannot be built *differently*. The release re-runs both
self-tests before publishing.

**A shipped jar can now name itself.** Every jar of every version is a file
called `goldenmatch-spark.jar`, and the failure this arc keeps paying for is the
silent one -- a jar whose library will not load falls back to the exact-only
scorer and keeps returning numbers. "Which jar?" is the first question of that
investigation. So the build stamps `Implementation-Version`, `JarVersion` reads
it from its own archive (not the first manifest on the classpath, which on an
executor is Spark's), and `java -jar goldenmatch-spark.jar` reports the version
plus whether the native kernel loads *here*, exiting non-zero on a fallback.
That makes it a preflight check for a deploy script, not just something a human
reads. CI gates the assembled jar, which is the only place `JarVersion` can find
a manifest and the only place the library comes from jar resources.

## 2. The kernels reach the ENTRY POINTS

The inventory measured internals -- `_transformed`, `merge_expr`,
`derive_record_ids` -- and reported 5 of 5 real capabilities routed. That
answers "is the kernel reachable", which is not the question a user has. A
kernel routed at an internal but not threaded through the entry point is, from
outside, **not routed at all**: there is no argument to pass. `transform_udf`
existed on `_transformed` with no entry point able to supply one, so anyone
running the documented pipeline got the Python path regardless.

`generate_candidates`, `build_golden_from_rules` and `run_config_pipeline` now
take `transform_udf` / `survivorship_udf`.

## The finding: there is deliberately no `scorer_udf`

Scoring in the config pipeline is per-field and row-shaped (one Python UDF per
field of each matchkey). The jar's scorer is a batched array-shaped UDF over a
single value column -- the shape Spark Connect forces. Different call
structures, not the same call with a different backend, so routing that stage is
a design change rather than a parameter.

**An end-to-end dedupe therefore still needs Python on the executors for scoring
alone**, however thoroughly every other stage is routed. Stated in the docstring
where someone would otherwise be misled, and pinned by a test: a `scorer_udf`
added later for consistency that quietly did nothing would be worse than its
absence.

Three new probes -- blocking and golden at their entry points, plus one
end-to-end dedupe. The dedupe is EXPECTED to fail. It is listed rather than
omitted because a missing probe reads as "not thought about" while a failing one
reads as "measured, and here is the cost".

`MUST_NOT_PASS` names the two probes that can never pass, so the score has a
stated ceiling (**8 of 10**) instead of a denominator that reads as 80% of the
way there. A typo in that set would silently raise the ceiling and reclassify a
real gap as a pass, so it is bound to the actual probe names at import.

## Verification

Local: 164 spark/sail tests pass, ruff clean, both workflows parse,
`check_workflow_yaml` green. **No JDK on this machine**, so `JarVersion`,
`Main`, and both new shell scripts are first exercised by CI -- that is what the
new `java -jar` gate in `spark_connect` is for.

Also widens the `spark` path filter: the lane no longer contains the jar build,
it calls scripts shared with the publish workflow, and that sharing only holds
while an edit to either still re-runs the lane that tests it.


benzsevern:
Follow-up commit closes the `scorer_udf` gap this PR originally documented as remaining. Passing all three UDFs now runs a whole dedupe with no Python worker on any executor; the inventory ceiling goes 8 -> 9 of 10.

**The gap description in the PR body above is wrong**, and the correction is the useful part. I had it as a shape mismatch (per-field row-shaped vs batched array-shaped). But `make_scorer_udf` is an ARROW udf: Spark hands it a whole batch and it returns a whole batch, so its real signature is already `(a[], b[]) -> double[]` -- exactly `golden_score_batch`. Same function, same granularity.

The only real difference is **who forms the batch**. Spark forms it implicitly for an arrow_udf; a Java UDF over Connect sees one row and must be handed its batch explicitly, which `collect_list` builds. That also explains why the naive port is a disaster rather than a small regression: per-row calls would be ~10,000 JNI crossings where there is one Python call today.

Three defects found while writing it, none of which fail loudly:

- **PySpark reads a higher-order lambda's meaning from its parameter count.** `lambda r, g=group: ...` is read as `(element, index)` and `g` receives the INDEX -- the captured value silently discarded. It misaligns rather than raising, which is the exact failure the one-collect_list/one-explode discipline exists to prevent. Both lambdas were written that way first; now closure factories, pinned by a test.
- **A union of one select per matchkey re-derives the batch per matchkey**, so N matchkeys meant N times the native work for an identical result. Replaced with `greatest`, which is the union's `max` and drops the final groupBy.
- **`exact` fields in a weighted matchkey still need a slot** -- after `collect_list` there is no raw column left to compare.

Gated by an exact-equality parity test over `score_candidates` parametrized on batch_size 1/3/10,000 (the reshape is trivially correct for a single batch, so one batch would prove nothing), plus an end-to-end `run_config_pipeline` comparison. Both need a real session and run in `spark_connect` -- no JDK on my machine.

PR #2555, opened 2026-08-13
feat(fs): Phase 0 -- the counted EM path in the Rust core, gated

benzsevern:
Spec: `docs/superpowers/specs/2026-08-13-fs-em-rust-single-source-design.md`.

**Stacked on #2552.** Its commits are the first four here and drop out on rebase.

## The finding

The Fellegi-Sunter EM loop exists **three times**:

| implementation | covers | callers |
|---|---|---|
| `probabilistic.py` (`train_em`, `_em_iterate`) | NE, TF, monotonicity, missing modes, counted path | everything |
| `probabilistic.ts` (`trainEM`) | NE; no counted path | the TS surface |
| **`score-core/src/em_core.rs`** | the discrete E/M loop only | **none** |

The copy designated as the source of truth is the least complete *and* has zero callers — `grep -rn "train_em_core\|EmOutput\|EmParams"` across `*.rs`/`*.py`/`*.ts` returns nothing outside the file itself. Meanwhile `bridge::train_em` **embeds CPython** and calls the Python trainer, so every non-Python surface reaches Python for FS training. It's a PR-C that built its numeric core and was never wired.

And #2550/#2552 made it worse: `pair_weights`, `train_em_from_counts`, `estimate_u_from_counts`, `_neutral_u_for`, `_combine_em_sessions` all went into Python only. `em_core.rs` already had `neutral_u` with byte-identical constants.

## What this PR does

Ports the counted shape into the Rust core — `train_em_from_counts`, `estimate_u_from_counts`, `combine_em_sessions`, `fixed_blocking_weights` — and puts a parity gate under it.

**It adds no callers.** On its own it is a *fourth* copy; it is only worth landing because Phase 1 (wiring it over C-ABI/wasm/JNI, so Python and TS stop carrying their own loops) follows immediately. That is stated in the spec's risks section too.

## The gate found two real bugs in #2552

Both in code queued to merge, both #1835-family, both invisible from either side alone.

**1. A conditioned field got `log2(m/u)` instead of the bounded ramp.** Its `m` is never updated — it keeps the exponential init — so that ratio is an arbitrary initialisation divided by a random-pair estimate. Measured on a 2-level blocking field:

| | disagree | agree |
|---|---|---|
| `train_em` (correct) | **-3.0** | **+3.0** |
| counted path | -1.54 | +4.47 |

The disagreement penalty that drives precision was more than halved; the agreement weight bounded to preserve recall ran 49% over. On a near-unique key this is the mechanism behind the measured F1 0.83 → 0.57.

**2. It reported the random-pair `u` it was handed, not the neutral prior.** Doesn't move the trained `m` (a conditioned field is skipped in the E-step), but `u_probs` is half the persisted model.

Both now route through single definitions — `_fixed_blocking_weights` and `_neutral_u_for` — which `train_em` and `estimate_m_from_labels` were also duplicating.

Bug 2 was found *by the Rust gate on its first run*, which is the entire argument for the spec.

## The gate

Anchors come from `scripts/gen_fs_em_parity_fixture.py`, which is **committed**. The existing `em_core.rs` anchors came from "the C1 commit message / scratch script" — uncommitted, so when one goes red there is no way to tell a Python behaviour change from a bad port. Here, re-running the emitter and reading `git diff` is the whole diagnosis.

Five cases: the base case, the **#1836 near-unique-blocking** one, a 3-level conditioned field (a hard-coded 2-element ramp would index wrong at scoring time), unobserved `-1` handling, and one pinning that weights are **counts** — the `1e-6` smoothing is additive, so rescaling them moves the model.

Tolerances travel *in* the fixture. Parity is decision-level, not bitwise: libm's `ln`/`log2`/`exp` differ from CPython's in the low mantissa bits. The 2026-07-18 plan's "byte-parity" wording was wrong, and `em_core.rs`'s own test helper already said so.

`serde_json` is a **dev**-dependency, so the wasm bundle budget and the crate's minimal-dep posture are untouched.

## Verification

- `cargo test` (score-core): 36 passed
- `cargo clippy --all-targets -- -D warnings`: clean; `cargo fmt --check`: clean
- Python FS/EM/probabilistic suites: 160 passed
- `regen_docs.py --check`: current

`cargo fmt` reformatted three files this PR does not touch (`examples/bench_*.rs`, `src/lib.rs`); those were reverted.

## Not in this PR

Phases 1-4 of the spec: wiring (the load-bearing one), TF adjustment on the distributed path, other backends, and the Splink head-to-head benchmark. **No comparative claim about Splink is supported until Phase 4 exists.**

### DIFF
--- a/docs/agent-codemap.json
+++ b/docs/agent-codemap.json
@@ -4155,6 +4155,7 @@
             "_field_score_matrix_dedup",
             "_field_values_for_block",
             "_field_values_from_list",
+            "_fixed_blocking_weights",
             "_fs_arrow_column",
             "_fs_batch_rows",
             "_fs_calibrate_threshold_enabled",
--- a/docs/superpowers/specs/2026-08-13-fs-em-rust-single-source-design.md
+++ b/docs/superpowers/specs/2026-08-13-fs-em-rust-single-source-design.md
@@ -1,0 +1,178 @@
+# Fellegi-Sunter training: one implementation, served everywhere
+
+**Date:** 2026-08-13
+**Status:** design; Phase 0 in progress
+**Supersedes nothing.** Resumes PR-C of
+`docs/superpowers/plans/2026-07-18-fs-rust-arrow-only.md`, which built its
+numeric core and was never wired.
+
+---
+
+## The finding
+
+The Fellegi-Sunter EM loop exists **three times** in this repo, and the copy
+designated as the source of truth is the only one nothing calls.
+
+| implementation | covers | callers |
+| --- | --- | --- |
+| `packages/python/goldenmatch/goldenmatch/core/probabilistic.py` (`train_em`, `_em_iterate`) | NE dims, TF tables, monotonicity guard, missing-value modes, weighted/counted rows | every Python surface, and `bridge::train_em` |
+| `packages/typescript/goldenmatch/src/core/probabilistic.ts` (`trainEM`) | NE dims; no counted path | the TS surface |
+| `packages/rust/extensions/score-core/src/em_core.rs` (`train_em_core`) | the discrete E/M loop only | **none** |
+
+Verified 2026-08-13: `grep -rn "train_em_core\|EmOutput\|EmParams"` across
+`*.rs`, `*.py`, `*.ts` returns nothing outside `em_core.rs` itself.
+`score-core/src/lib.rs` does `pub mod em_core;` and stops there.
+
+Meanwhile `packages/rust/extensions/bridge/src/api.rs::train_em` is a pyo3 shim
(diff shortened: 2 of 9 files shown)


benzsevern:
Correction to this PR's own claim: the body said `cargo clippy --all-targets -- -D warnings` was clean. It was not — the `cargo` proxy on my box exits 0 **without running clippy**, so that check was a no-op and I reported it as a pass.

CI caught the real lint (`needless_range_loop` in `combine_em_sessions`). Fixed and amended into the commit, and re-verified with the toolchain's direct `cargo-clippy.exe`, which does run: score-core clippy clean, 36 tests pass.

PR #2558, opened 2026-08-13
feat(pg): Phase 1c -- FS training in Postgres with no embedded CPython (0.19.0)

benzsevern:
Spec: `docs/superpowers/specs/2026-08-13-fs-em-rust-single-source-design.md`. **Stacked on #2556.**

## Scope finding first

The spec's Phase 1 item reads *"`bridge::train_em` stops embedding CPython"*. **That is not a slice.** The bridge *is* the Python embedding layer by design — its own docstring says "Python embedding layer for SQL extensions" — and `goldenmatch_train_em` cannot be de-bridged because it samples pairs and bins them into comparison vectors, and neither is ported. Porting them is PR-C's "the big one", whose hard part is reproducing a seeded sampler exactly.

So this takes the split the repo's own history already blesses. `goldenmatch_score` and the goldenflow transforms were **partially** de-bridged — some paths native, others still Python, no version bump. Here the split is by **input shape**:

| function | input | path |
|---|---|---|
| `goldenmatch_train_em` | rows | bridge → CPython (unchanged) |
| **`goldenmatch_train_em_from_counts`** | counted vectors | **`score-core::em_core` native-direct** |

That's the design, not a workaround. Counting agreement patterns is `GROUP BY` over the gamma columns — the one part a SQL engine is unambiguously right for — and the distinct-vector count is bounded by `prod(levels + 1)`, so the result is thousands of rows however many pairs were compared. **Postgres can now do both halves of FS training with no interpreter in the backend.**

## The version dance

New function → version bump, matching `hnsw_pairs` (0.10.0), `lsh_pairs` (0.11.0), goldencheck (0.13.0), key-integrity (0.17.0), docs (0.18.0). Control + Cargo to **0.19.0**, new base SQL, a `0.18.0→0.19.0` migration, both added to the CI install list.

## No `#[pg_test]`, deliberately

I wrote four, then found **`cargo pgrx test` cannot run for this crate** — CI's own comment says so, and it uses psql smoke instead. Those tests would have compiled and never executed: untested code wearing a test's clothes.

They're a `rust_pgrx` **psql smoke** now, running against a real `CREATE EXTENSION`, pinned to the *same anchors* as `score-core/tests/fixtures/em_counts_parity.json` so this surface cannot drift from the wheel's trainer. Covers the learnable case, the **#1836 conditioned-field case** (bounded `-3..+3` ramp + neutral `u` — the failure that took F1 0.83 → 0.57), and the fail-soft envelope.

## Input handling

Flat arrays per the `goldenmatch_hnsw_pairs(flat_vecs, dim)` idiom, since pgrx flattens multidim arrays. `u_probs` is **ragged** — field `j` contributes `n_levels[j]` entries — flattened in field order and split back by `n_levels`; a rectangular block is a length error rather than a silently misaligned model. Levels are range-checked, not cast. NaN counts rejected explicitly: NaN fails every comparison, so it slips past `<= 0.0`, poisons every weighted sum, and a NaN delta compares false against the convergence threshold — EM would run to the cap and return a model of NaNs.

## Verified / not verified

**Locally green:** `check_pgrx_sql_sync` (82 `#[pg_extern]` vs 82 `CREATE FUNCTION`), `check_workflow_yaml`, `regen_docs --check`, llms/README count gate, parity manifest updated.

**Not verified locally:** the crate does not compile on this box — `pgrx-pg-sys` needs PG headers there's no install for. So `fs_em.rs` compiling and the smoke passing rest on the `rust_pgrx` CI lane. Saying so explicitly because "it compiles" is not something I can claim here.

## Phase 1 status

- **1a** (#2556): Python trains through the kernel ✅
- **1c** (this): Postgres trains through the kernel, no CPython ✅
- **1b**: TypeScript still carries its own `trainEM` — not started
- Row-shaped `bridge::train_em` still embeds CPython, and will until the sampler and binning are ported.


benzsevern:
Fixed a real break this PR introduced. My new psql-smoke step was inserted **into the middle of an existing step's `run:` block** — on `main`, `psql smoke (scalar scorer, dedupe, autoconfig)` is one 501-line step, and my edit split it: 45 lines kept the original name and the remaining ~456 (including the `gm_resolve` smoke) ran under mine. That is why `rust_pgrx` failed on all three PG versions.

The YAML parse gate passed the whole time, because a step swallowed into another is still valid YAML. Rebuilt `ci.yml` from `main`'s copy, re-applied only the 0.19.0 `cp` lines, and appended my step **after** the job's last step. Verified structurally, not by parse: the job now has 12 steps with the 501-line one intact and mine at 55, and `jobs` is byte-identical to main outside `rust_pgrx`.

The extension crate itself compiled fine in CI, so `fs_em.rs` and `is_multiple_of` are good.

benzsevern:
Closing as **superseded, not abandoned** — every line of this PR is already on `main`.

#2559 (Phase 1b) was stacked on this branch, so its head carried this commit too. The merge queue squashed that whole branch, which landed the Postgres work under #2559's commit `02c0412b5`. Verified on `main`: `packages/rust/extensions/postgres/src/fs_em.rs` and `sql/goldenmatch_pg--0.19.0.sql` are both present, along with the control bump, the parity-manifest entry and the psql smoke.

**And the smoke ran green on all three PG versions** — `rust_pgrx (15|16|17): success`, with `NOTICE: FS EM from counts: OK (no CPython in this path)` in the log. That is the first execution of these assertions; the three earlier attempts died before reaching the SQL (step-boundary break, then `PGPORT` unbound). So `goldenmatch_train_em_from_counts` is now verified end to end against a real `CREATE EXTENSION`, native-direct over `score-core::em_core` with no embedded CPython.

Not merging this PR: its branch predates #2559, so its diff against `main` is now purely the **deletion** of the TypeScript work. Merging it would revert #2559.

One attribution wart worth recording: the Postgres 0.19.0 surface landed under a commit titled "Phase 1b -- TypeScript". The code is right; the commit title undersells what it carries.

PR #2559, opened 2026-08-13
feat(ts): Phase 1b -- TypeScript trains FS through the shared Rust kernel

benzsevern:
Spec: `docs/superpowers/specs/2026-08-13-fs-em-rust-single-source-design.md`. **Stacked on #2558.**

`fs-wasm` gains `train_em_from_counts` over `score-core::em_core`; `fsWasm.ts` exposes it as `trainFsFromCounts`. **TypeScript reaches the shared kernel for the first time.**

## The design decision: counted shape only, `trainEM` untouched

This does **not** reroute `trainEM`, and that's the call I made rather than an omission.

TS `trainEM` supports **negative-evidence dimensions**. `em_core` explicitly does not — its own header says *"NOT yet ported: negative-evidence dims…"*. Routing `trainEM` through the kernel today would silently drop NE and return a model the config never asked for, and nothing downstream could detect it. The counted path has no such gap: it refuses NE on every surface.

So this ships the kernel + parity as an **alternative surface first** — exactly how fs-wasm's own scoring kernel shipped before the TS scorer was rerouted onto it. **The row-shaped duplicate stays until `em_core` covers NE**, named rather than implied fixed.

## Shared anchors

`tests/parity/fs-em-counts.parity.test.ts` and the Rust-side tests both pin the numbers in `score-core/tests/fixtures/em_counts_parity.json`, emitted from the Python reference. Three surfaces asserting against three sets of hand-copied numbers would each agree with their own copy and nothing else.

Includes the `weights_are_counts_not_proportions` case: the same shape at 1/100th the counts must **not** give the same model, because EM's `1e-6` smoothing is additive. A surface that normalised counts would pass every other test and shift exactly the low-probability cells where FS weights are largest.

## Bundle discipline

`score-core` added with `default-features = false`, matching what `fs-core` already does — drops the `alias` feature and its `regex` engine, which must not reach an edge bundle.

**Measured:** embedded wasm went **506,300 → 488,540 bytes** of base64. *Smaller*, despite the added code — a newer toolchain. The size guard this package maintains is intact.

## Drive-by: the regen script never worked on Windows

`scripts/build_fs_wasm.mjs` could not run on a Windows box at all. `execFileSync` neither applies PATHEXT (bare `wasm-pack` → `ENOENT`) nor will spawn the `.cmd` shim since Node hardened against CVE-2024-27980 (→ `EINVAL`). Fixed, with args quoted explicitly because `shell: true` re-parses the line.

Clippy also caught `manual_is_multiple_of` in the new code — fixed here and in the Postgres sibling sharing the idiom.

## Verified / not verified

**Green locally:** `cargo test` (fs-wasm — 8 passed, 5 new), clippy `-D warnings`, `cargo fmt --check`, `regen_docs --check`.

**Not verified locally:** the TS tests, and the regenerated wasm blob's *scoring* parity. This package has no `node_modules` here and installing + running vitest OOMs this box. CI's `fs-wasm.parity.test.ts` gate — whose fixture is authored by the Python **native** kernel, an independent oracle — is what validates the rebuilt blob. Flagging it because the blob is a binary artifact I changed and could not exercise.

## Phase 1 status

| slice | state |
|---|---|
| 1a — Python | ✅ #2556 |
| 1c — Postgres | ✅ #2558 |
| **1b — TypeScript** | ✅ this (counted shape) |
| row-shaped `trainEM` / `bridge::train_em` | still duplicated / still embeds CPython |

Both remaining items need `em_core` to grow NE and the sampler+binning port respectively — separate arcs, not slices of this one.


benzsevern:
Fixed the `typecheck` failure. This package runs `noUncheckedIndexedAccess`, so `out.match_weights[0][0]` is `number | undefined` — 14 TS2532/TS2345 errors, all in my new test file. `src/core/fsWasm.ts` typechecked clean.

Rewrote the assertions through a `cell(table, j, k)` helper that throws when a cell is absent. That is better than a non-null assertion anyway: `expect(undefined).toBeCloseTo(x)` is exactly the shape that looks like coverage while testing nothing, and a model missing a whole field should name the field.

I could not have caught this locally — no `node_modules` here and installing plus running vitest OOMs this box, which is why the PR body listed the TS tests as unverified.

PR #2615, opened 2026-08-16
diag(zeroconfig): record the number the blocking RED rule should be using

benzsevern:
The blocking RED rule is:

```python
avg = n_rows / max(n_blocks, 1)
if block_sizes_p99 > 10 * avg:   # RED
```

That denominator is a **mean**, and it is pinned near 1 whenever blocking is fine-grained. Measured on person@100k:

| | |
|---|---:|
| n_blocks | 84,293 |
| n_rows | 100,000 |
| **avg** | **1.19** |
| bar (10 x avg) | 11.9 |
| p99 | 72 |
| reduction_ratio | **0.9757** |
| singleton_fraction | **0.0** |

Any block over a dozen rows trips it. That config has *excellent* reduction and *zero* singletons, and it grades RED — which is why zero-config refuses it.

## Why this PR does not fix the rule

Because the first version of this diagnostic did not capture the quantity a better rule would use. Skew is dangerous when a few blocks own most of the **work**, and work is quadratic in block size — so the honest measure is what fraction of total candidate pairs the largest block contributes, not a tail percentile over a mean.

Now recorded, all from fields `BlockingProfile` already carries:

- `total_comparisons` — the denominator that was missing
- `largest_block_pairs` — `max*(max-1)/2`
- `largest_block_pair_share` — the candidate replacement signal
- `p99_over_avg` — the current rule's ratio, made explicit
- `p99_over_p50` — the obvious alternative denominator

**Recorded, not acted on.** A RED rule decides whether zero-config refuses a user's run. Swapping one heuristic for another on reasoning rather than evidence is presumably how the current rule got here. Re-run the diagnostic, read the share, then change the rule.

### DIFF
--- a/scripts/bench_er_headtohead/diagnose_zeroconfig_refusal.py
+++ b/scripts/bench_er_headtohead/diagnose_zeroconfig_refusal.py
@@ -83,15 +83,38 @@ def _blocking_report(bp, n_rows: int) -> dict:
         fired.append(f"block_sizes_p99 {p99} > 10*avg {10 * avg:.1f}  [SKEW]")
     if rr < 0.5:
         fired.append(f"reduction_ratio {rr:.4f} < 0.5")
+    # Pair-share of the biggest block: the measure the RED rule arguably should
+    # be using. `p99 > 10 * avg` compares a tail percentile to a MEAN that is
+    # pinned near 1 whenever blocking is fine-grained (n_blocks -> n_rows), so
+    # it fires on configs with excellent reduction and no singletons. What
+    # actually makes skew dangerous is that a few blocks own most of the WORK,
+    # and work is quadratic in block size -- so the honest question is what
+    # fraction of total candidate pairs the largest block contributes.
+    #
+    # Recorded, not acted on: changing a RED rule needs this number first, and
+    # the first version of this diagnostic did not capture it.
+    total_pairs = getattr(bp, "total_comparisons", 0) or 0
+    biggest = getattr(bp, "block_sizes_max", 0) or 0
+    biggest_pairs = biggest * (biggest - 1) // 2
     return {
         "n_blocks": n_blocks, "avg_block_size": round(avg, 2),
         "block_sizes_p50": getattr(bp, "block_sizes_p50", 0),
         "block_sizes_p95": getattr(bp, "block_sizes_p95", 0),
         "block_sizes_p99": p99,
-        "block_sizes_max": getattr(bp, "block_sizes_max", 0),
+        "block_sizes_max": biggest,
         "reduction_ratio": round(rr, 6),
         "singleton_block_count": singles,
         "singleton_fraction": round(singles / max(n_blocks, 1), 4),
+        "total_comparisons": int(total_pairs),
+        "largest_block_pairs": int(biggest_pairs),
+        "largest_block_pair_share": (
+            round(biggest_pairs / total_pairs, 6) if total_pairs else None
+        ),
+        "p99_over_avg": round(p99 / avg, 2) if avg else None,
+        "p99_over_p50": (
+            round(p99 / getattr(bp, "block_sizes_p50", 0), 2)
+            if getattr(bp, "block_sizes_p50", 0) else None
+        ),
         "red_rules_fired": fired,
     }
 


benzsevern:
Re-ran the diagnostic at 100k on both shapes. Here is the number this PR stopped for.

## `largest_block_pair_share`

| | person@100k | biblio@100k |
|---|---:|---:|
| n_blocks | 84,350 | 22,151 |
| avg_block_size | 1.19 | 4.51 |
| p50 / p95 / p99 | 2 / 10 / 72 | 6 / 31 / 39 |
| block_sizes_max | 2,170 | 70 |
| reduction_ratio | 0.9757 | 0.9996 |
| singleton_fraction | 0.0 | 0.0 |
| total_comparisons | 121,391,850 | 1,770,258 |
| largest_block_pairs | 2,353,365 | 2,415 |
| **largest_block_pair_share** | **0.0194** | **0.0014** |
| p99_over_avg (current rule, bar 10) | **60.73** | 8.64 |
| p99_over_p50 | 36.0 | 6.5 |
| blocking verdict | **RED** (skew) | GREEN |

**The current rule is a false alarm on person, by a wide margin.** It grades the config RED at `p99/avg = 60.7` against a bar of 10, while the largest block owns **1.9% of total candidate pairs**. Reduction is 0.9757 and there are zero singletons. Nothing here is carrying a dangerous share of the work.

Note the two shapes differ 14x on the share (1.94% vs 0.14%) yet both are trivially small in absolute terms.

## Why this still does not license changing the rule

**Neither shape is a positive control.** Both are healthy, so this run establishes where the signal sits for *good* configs and says nothing about where it sits for a genuinely dangerous one. Picking a bar from two negatives would repeat the mistake that produced the current rule: choosing a threshold with no positive examples on the other side of it.

What is needed before the rule changes is a deliberately degenerate blocking config (block on a low-cardinality column so one block owns most of the pairs) measured through the same diagnostic. Then there are two clouds and the bar goes between them. I have not built that yet.

## A second finding, unrelated to the denominator

person@100k is RED on **two** sub-profiles, not one:

```
health: blocking=RED  scoring=RED  (data/domain/matchkey/cluster all GREEN)
scoring: n_pairs_scored=0  dip_statistic=0.0  mass_above_threshold=0.0
```

Scoring is RED because **nothing was scored at all**, not because the scores were bad. So fixing the blocking denominator alone would not clear this refusal unless the zero-pair scoring profile is itself downstream of the blocking verdict. That relationship needs establishing; right now "fix lever 1 and person@100k stops refusing" is not supported.

biblio@100k, by contrast, now comes back **rollup=GREEN on every sub-profile** (stop_reason `BUDGET_ITERATIONS`), so the biblio refusal recorded in the head-to-head is not reproducing on this path.

## Reproduction

```bash
uv sync --package goldenmatch --extra polars
python scripts/bench_er_headtohead/diagnose_zeroconfig_refusal.py \
  --shapes person biblio --rows 100000 --out zeroconfig-refusal-100k.json
```

The `--extra polars` matters: the diagnostic imports polars directly, but polars is an optional extra of `goldenmatch`, so a default install fails with `ModuleNotFoundError` before reaching the controller.

Timings: person 135.35s, biblio 356.24s.

---

benzsevern:
Followed up on the second finding above (person@100k RED on **both** blocking and scoring). The scoring RED resolves cleanly, and it changes what lever 1 is worth.

## person@100k scoring sub-profile

```json
{
  "candidates_compared": 0,
  "n_pairs_above_threshold": 0,
  "scoring_ran": false,
  "dip_statistic": 0.0,
  "mass_above_threshold": 0.0,
  "red_cause": "scoring-never-ran (candidates_compared == 0)"
}
```

**Scoring never ran.** The emitter never received a `ScoringProfile`, so `autoconfig_controller.py`'s `scoring = emitter.scoring or ScoringProfile()` substituted the all-zero default, and `ScoringProfile.health()` graded that default RED via its `candidates_compared == 0 and n_pairs_scored == 0` guard.

So the scoring RED is **not** an independent finding about thresholds or scorers. It is an artifact of the profile never being produced. That is the good case for lever 1: it is not blocked by a second, unrelated RED.

Two caveats before that is treated as "fix blocking and person passes":

1. **Why no candidates were compared is still unestablished.** The blocking profile for the same run reports `n_blocks=84,350` and `total_comparisons=121,391,850`, so blocking was not empty at the profile level. The zero has to come from the scoring path, not from an absence of blocks.
2. `first_red_subprofile` is `blocking` and the run stopped `POLICY_SATISFIED`, so the search converged and committed a config that grades RED on a sub-profile whose numbers were never measured.

## The naming trap, for the record

`n_pairs_scored` is **not** the number of pairs scored. `scorer.py:89` builds it from `find_fuzzy_matches` output, which is already filtered to pairs above the threshold, and `autoconfig_policy.py:298` renders the same field as `n_pairs_above_threshold`. The name reads like the denominator and is actually the numerator. Reading it by its name is what made the first report ambiguous; `candidates_compared` is the field that resolves it. Both names are now emitted by the diagnostic (#2631) rather than renaming a field other surfaces read.

## A defect this exposes, independent of the above

`scorer.py:2185`:

```python
if not blocks:
    return []          # returns WITHOUT emitting a ScoringProfile
```

Same shape at `scorer.py:2629`. When either fires, the controller falls back to the all-zero `ScoringProfile()`, which is **indistinguishable from a real scoring pass that compared candidates and found nothing** — and `health()` grades both RED. "Never ran" and "ran and found nothing" are different diagnoses that should not collapse into the same profile, especially when one of them is what makes zero-config refuse a user's run.

A sentinel already exists for exactly this problem on a neighbouring path: `_yellow_sentinel_profile` sets `candidates_compared=1` with the comment *"The actual scoring didn't run for these pathological-input short-circuit paths."* The same honesty is missing on the early returns above.

Suggested split: fixing the silent-default is worth its own change and is not gated on the RED-rule decision. I have not made it.

---

PR #2664, opened 2026-08-18
HELD: zero-config blocking rule fixes (regresses recall; needs a replacement plan)

benzsevern:
**HELD.** This is the behaviour half of the original PR. It **regresses zero-config end to end** and must not merge in this shape. The safe half was split out into #2667.

## The regression, measured

person@100K, same fixture, `gm_zeroconfig`:

| | before (main) | with this PR |
|---|---:|---:|
| pairwise R | 0.4684 | **0.383** |
| pairwise F1 | 0.6380 | **0.554** |
| scored pairs | 11,319 | **9,250** |
| block_count | 14,081 | **3,287** |

Fewer candidates, not more -- the opposite of the intent. For reference on the same run: `gm_probabilistic_shipped` F1 0.997, splink 0.995.

## Why

`rule_low_reduction_ratio` fires at iteration 0 on an **unmeasured** profile (`reduction_ratio` defaults to `0.0`, and `0.0 < 0.5`), then rebuilds `passes` from `keys` -- discarding six of eight passes. Both are genuine bugs and both are fixed here, with tests.

But the discarded-plan bug had an accidental *benefit*: the rule's own addition, `dob soundex`, was contributing candidates. Once the blocking profile is measured properly (`reduction_ratio 0.9757`, GREEN) the rule correctly stops firing -- and the soundex pass stops being added too. **The fixes remove a bad rule's side effect without replacing it.**

So the open question is not "are these bugs real" (they are, and they are fixed) but **what blocking plan zero-config should actually commit**. That needs the 100K ablation, which needs #2667 on main to dispatch.

## What is still in here

* both `rule_low_reduction_ratio` defects (unmeasured-profile firing; `keys` vs `passes`)
* `_diversify_probabilistic_blocking`: arrow-seam scale-guard repair (it built raw polars expressions that RAISE on a `pyarrow.Table` and landed in `except: return 0`, where 0 reads as "safe" -- so the #1857 guard has been silently disabled on the default arrow lane) plus `n_rows_full` projection for sample callers
* controller: `_should_measure_blocking` extended to `multi_pass`, and measure-before-refusing
* `test_planner_integration` updated -- those tests injected pair counts through `extrapolate_to`, a seam the controller stops using once it measures

## Corrections recorded in-branch

Two of my own claims here were wrong and are corrected in `c9b7dd6c` rather than quietly superseded: there is **no** pair-count mismatch (I read `scored_pairs`, the retained post-cut set, as the candidate count -- `measure_blocking_profile`'s 121,391,850 was right), and the controller was **not** hanging (121M comparisons in pure Python with `GOLDENMATCH_NATIVE=0`; CI does the same work in ~27 s).

Depends on #2667. Rebase onto main once that lands, so this carries only the behaviour delta.

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/bench-blocking-ablation.yml
+++ b/.github/workflows/bench-blocking-ablation.yml
@@ -81,7 +81,7 @@ jobs:
             --out "out/ablation_${{ inputs.rows }}.json"
 
       - name: Upload results
-        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02  # v4
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a  # v4
         with:
           name: blocking-ablation-${{ inputs.run_tag }}
           path: out/
--- a/.github/workflows/bench-blocking-ablation.yml
+++ b/.github/workflows/bench-blocking-ablation.yml
@@ -1,0 +1,88 @@
+# Which blocking passes earn their comparisons?
+#
+# auto-config's probabilistic plan for person@100K costs 121,391,850
+# comparisons, and four of its eight passes run 11-22x over the engine's own
+# pairs-per-row budget while carrying 99.7% of that work. Whether those passes
+# are worth it is a measurement, not a deduction -- dropping them is what
+# `rule_low_reduction_ratio` already did by accident on this shape, for pairwise
+# recall 0.4684.
+#
+# Runs WITH the native kernel on purpose: pure Python does not finish 121M
+# comparisons in 500 s, CI does the same work in ~27 s. A local arm would be
+# measuring the absence of the kernel.
+name: bench-blocking-ablation
+
+on:
+  workflow_dispatch:
+    inputs:
+      rows:
+        description: "Fixture rows"
+        default: "100000"
+      dupe_rate:
+        description: "Duplicate rate"
+        default: "0.20"
+      shape:
+        description: "Fixture shape"
+        default: "person"
+      runner:
+        description: "Runner label"
+        default: "large-new-64GB"
+      run_tag:
+        description: "Artifact suffix"
+        default: "ablation"
+
+permissions:
+  contents: read
+
+env:
+  ARROW_DEFAULT_MEMORY_POOL: system
+  GOLDENMATCH_AUTOCONFIG_MEMORY: "0"   # no cross-run config cache: every arm rebuilds
+
+jobs:
+  ablate:
+    runs-on: ${{ inputs.runner }}
+    timeout-minutes: 240
+    steps:
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10  # v6
+
+      - uses: dtolnay/rust-toolchain@29eef336d9b2848a0b548edc03f92a220660cdb8  # stable
+
+      - uses: Swatinem/rust-cache@e18b497796c12c097a38f9edb9d0641fb99eee32  # v2
+        with:
+          workspaces: packages/rust
+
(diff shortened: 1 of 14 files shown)


benzsevern:
Closing. The behaviour half is superseded by #2671, which reaches a better result than this PR was aiming for, and the safe half already landed.

## The open question this PR was held on is answered

It ended with: *"the open question is not 'are these bugs real' (they are, and they are fixed) but **what blocking plan zero-config should actually commit**."*

The answer turned out not to be a blocking-plan choice at all. Both defects named here are real and are fixed in #2671, but they were two of **four** chained defects, and this PR had only the first two -- which is exactly why it regressed:

1. `_should_measure_blocking` gated full-frame measurement on `strategy == "static"`; zero-config commits `multi_pass`, so it always reasoned over an all-zero sample-extrapolated profile.
2. `rule_low_reduction_ratio` fired on that unmeasured `0.0` and rebuilt `passes` from `keys`, dropping 6 of 8 passes. (This PR's fix, correct.)
3. Fixing (1) makes the planner see a real ~121M candidate pairs, so `rule_chunked` sets `backend='chunked'` -- which `_fs_use_bucket_route` excluded, silently dropping FS to the legacy batched scorer (9,250 pairs retained vs bucket's 120,269 on identical candidates).
4. `_stamp_resolved_link_thresholds` recorded a FALLBACK cutoff, which counts as `configured` and skips the full-data threshold refit.

Result on the same fixture and lane, person@100K `gm_zeroconfig`:

| | main (then) | this PR | #2671 |
|---|---:|---:|---:|
| pairwise R | 0.4684 | 0.383 | **0.9949** |
| pairwise F1 | 0.6380 | 0.554 | **0.9970** |
| scored pairs | 11,319 | 9,250 | **120,269** |

#2671 matches `gm_probabilistic_shipped` exactly. So the `_diversify_probabilistic_blocking` call this branch adds to the v0 path is not needed -- the recall was never really about which anchors v0 picks, it was about a measurement that never ran and a routing gate that silently swapped scorers.

Full account, including the four instruments that lied and the eight hypotheses tested and discarded: `docs/superpowers/notes/2026-08-18-zeroconfig-recall-incident.md`.

## What already landed from here

`_projected_max_block` arrow repair, multi_pass vectorized measurement, the pair budget, the bench telemetry fix, the ablation harness and the shard-crash note all shipped via #2667 / #2669 / #2670 and are on `main`.

## Two things stranded here that should NOT be lost

**1. `.github/workflows/bench-blocking-ablation.yml`** -- the dispatch lane for the ablation harness. `scripts/bench_er_headtohead/ablate_blocking_passes.py` is on `main`; the workflow that runs it at 100K is only here, so the harness currently has no way to be dispatched at the scale its own docstring says decides the question.

**2. The `ci.yml` `suggest_full_dist` OOM fix -- documented as done on `main`, but not actually there.** `docs/superpowers/notes/2026-08-18-suggest-full-dist-shard-crash.md` is on `main` and opens with *"Status: root-caused and fixed. The single crashing test is `--deselect`ed from the parallel `python_goldenmatch` shards and re-run SERIALLY"*. `main`'s `ci.yml` contains **zero** occurrences of `deselect` or `suggest_full_dist`, and `tests/test_suggest_full_dist.py` still exists. Only the writeup landed; the change is stranded on this branch.

That is a latent landmine rather than a live failure -- the note explains the crash is triggered by pytest-split group reshuffling, so it bites whenever a PR adds test files in the wrong combination. Worth re-landing on its own, or the note should be corrected to say the fix is unshipped. A doc asserting a fix that is not in the tree is the same failure mode as a gate nobody has watched fail.

Filing both separately rather than reopening this branch, whose behaviour half should stay closed.

Issue #2663, opened 2026-08-18
Zero-config silently returns "no duplicates" when the committed threshold sits above the whole score distribution


benzsevern:
Found by the `orgs_hard` corpus (#2662) within minutes of it existing. Reproduces in ~2 seconds, deterministically, no network.

## The symptom

True zero-config, no escape hatch, on a corpus with **1,055 true duplicate pairs**:

```python
res = goldenmatch.dedupe_df(df)      # 845 rows
# clusters = 845   (all singletons)
# scored_pairs = 0
# no exception
```

A user runs zero-config on data that is 30% duplicates and is told it is clean. That is the worst shape of wrong answer: not an error, not low precision, but a confident empty result.

A working config on the same data gets **F1 0.7732** (P 0.8115 / R 0.7384).

## Root cause

Auto-config commits `threshold=0.80` on the weighted matchkey. The score distribution on this shape never reaches it, so nothing clears. Dropping to 0.60 gives **678 pairs / F1 0.6325** on the otherwise-identical config, so the config is one number away from working.

`rule_no_matches` exists for exactly this (`mass_above_threshold == 0` → lower the threshold). It never runs. The rule order is:

```
#3   rule_blocking_key_swap            structural: mass_above==0 with prior decision
#11  rule_cross_blocking_disagreement
#12  rule_low_transitivity
#13  rule_no_matches                   tuning: nothing matches
```

and the observed trace is:

| iter | rule fired | threshold | candidates | mass_above |
|---|---|---|---|---|
| 0 | `cross_blocking_disagreement` (#11) | 0.80 | 1327 | **0.0** |
| 1 | `blocking_key_swap` (#3) | 0.80 | 1327 | **0.0** |
| 2 | `low_transitivity` (#12) | 0.80 | 9019 | 1.0 |
| 3 | `low_transitivity` (#12) | 0.75 | 9019 | 1.0 |
| 4 | — | 0.80 | 1327 | **0.0** |

committed: **v0**, threshold 0.80, `stop_reason=BUDGET_ITERATIONS`, `failing_subprofile=scoring`

**The dedicated rule sits at #13, behind ten structural rules, and the 4-iteration budget is exhausted before it is ever tried** — even though the total-failure condition was present from iter 0 and persisted.

Two further things the trace shows:

- iters 2-3 found configs with `mass_above == 1.0` (everything matches). So the search saw both failure directions and committed the one that matches **nothing**.
[...]

Issue #2526, opened 2026-08-13
Zero-config scores F1 0.1220 on DBLP-ACM and the suggester recovers none of it — and the scorecard does not gate it


benzsevern:
Found while A/B-ing the FS refit default flip (#2518 follow-up, #2522). Not caused by that change - both arms of the A/B report it identically, so this is pre-existing on `main`.

## What

`scripts.suggest_quality report --datasets dblp_acm`, native kernel present, on `02c77628ba76`:

```
  dataset      kind   rows   gt_pairs   base_f1   n_sugg   rank_corr   sugg_prec   conv_f1   steps
  dblp_acm     real   4910   2224       0.1220    2        n/a         1.00        0.1220    0
```

Two things in one row:

1. **Zero-config auto-config scores 0.1220 F1** on DBLP-ACM. For reference, the package README puts goldenmatch at **0.918** on DBLP-ACM, and the FS probabilistic path at **0.968 F1** (97.8% P / 95.8% R). Those are explicit/tuned numbers, so 0.12 vs 0.92 is not an apples-to-apples gap - but a ~0.8 F1 spread between zero-config and configured on a standard benchmark is worth a look on its own terms, given zero-config is the product.
2. **The suggester recovers none of it**: 2 suggestions emitted, `suggester_prec` 1.00, but `steps=0` and `conv_f1 == base_f1`. Nothing it proposes has positive measured lift, so convergence exits immediately.

The controller's own verdict says it knows the config is bad:

```
All 1 exact-eligible columns were excluded by auto-config guards
  (id (full-frame cardinality_ratio=1.0000 >= 1.0 -- perfectly-unique surrogate key,
   no shared identity to match)).
  Falling back to fuzzy-only matchkeys
auto-config committed best-effort RED config
  (iter=v0, stop_reason=BUDGET_TIME, failing_subprofile=scoring);
  downstream pipeline will run but output may be low-precision
```

The `id` exclusion looks correct (it is a surrogate key). The interesting part is `stop_reason=BUDGET_TIME` with `failing_subprofile=scoring` - the controller ran out of **time**, not iterations, and shipped a RED config. That also makes the result machine-speed-dependent: a slower or loaded runner would get a different number of iterations, so this may not be deterministic across environments.

## Why nothing is red

[...]

PR #2687, opened 2026-08-19
fix(autoconfig): a constant column halved the bar on the ones that matter (#2668)

### LINKED ISSUE #2668: Zero-config over-merges since #2648: bucket ScoringProfile changed what the controller converges on
`goldengraph-pipeline` has been red on `main` since #2648 landed. Bisected to a single commit; the numbers below are measured, the mechanism at the end is a hypothesis and labelled as such.

## First bad commit

**`6599780c` — "fix(backends): bucket / duckdb / ray now emit a ScoringProfile (#2648)"**

`git bisect` over 72 revisions, `8145b498` (good) → `origin/main` (bad), 6 steps, each running the single decisive test. Clean transition, no skips:

```
a6b207d8  passed        ec92ffd0  passed  <- parent
b4fa5bc2  passed        6599780c  FAILED  <- first bad
17e08adf  passed        847f5a17  FAILED
bac7cd44  FAILED
```

The `goldengraph-native` wheel was installed non-editable and held fixed across every step, so this isolates the goldenmatch Python change rather than the Rust crate.

## What actually changed: it over-merges

Same 10-surface / 4-entity Wikidata TINY fixture, `resolve_mode="real"`, directly across the boundary:

| | distinct keys | merged groups | largest group |
|---|---|---|---|
| `ec92ffd0` (parent) | **8** | 2 | 2 |
| `6599780c` (first bad) | **3** | 4 | 3 |

Ten surfaces collapse into three keys, including merges across distinct entities. The failing assertion is the *conservative* one, not the "did anything merge" one:

```python
assert any(len(ks) > 1 for ks in keys_by_entity.values())   # False
#      "at least one entity kept >1 cluster key (conservative real ER)"
```

Worth stating plainly because it inverts the obvious reading: the resolver is not failing to cluster, it is clustering too much. The other three failures follow from it — set recovery drops to 0.8 against a 0.99 floor, and co-occurrence count accuracy goes to 0.0.

Runtime is a corroborating signal: the parent settles in ~0.9s, the bad commit burns the full 4-iteration budget at ~3.8s, so the controller is searching harder and landing somewhere looser.

## Hypothesis for the mechanism (NOT verified)

#2648 is a good fix for a real problem — the bucket path emitted an all-zero `ScoringProfile()`, which read as "nothing happened" and made the controller REFUSE at `n_rows >= REFUSE_AT_N`. Nothing here argues for reverting it; that would reinstate a refusal at 100k.

The suspect is the new `profile_threshold()` fallback. For probabilistic matchkeys, where `mk.fuzzy_threshold` raises, it returns the lowest returned score. Since `find_fuzzy_matches` only returns pairs at or above the cut, `mass_above_threshold` is then **1.0 by construction**. That is tautologically true and carries no information, but the controller consumes it as a scoring signal — so a run that admitted almost nothing and a run that admitted everything both report perfect mass. On a 10-row frame that plausibly removes the pressure that previously kept the config tight.

Someone should confirm that against the actual profiles before fixing; I have not.

## Why nothing caught it

The four tests are gated on the `goldengraph-native` wheel, so they skip on any machine without it and no PR lane builds it. `goldengraph-pipeline` is scheduled and not in `ci-required`, so #2648 went green on its PR and the lane reddened that night. Same shape as the gaps in #2666 and #2635: the check exists and does not fire where it would matter.

## Reproduce

```
maturin build --release -m packages/rust/extensions/goldengraph-native/Cargo.toml --out dist
uv pip install dist/*.whl
cd packages/python/goldenmatch/benchmarks/er-kg-bench
POLARS_SKIP_CPU_CHECK=1 pytest tests/test_realworld_aggregation.py -q
# 4 failed, 8 passed
```

Related: the Abt-Buy RED-config problem in #2457 is a different dataset but the same family — the controller committing a config it graded RED and the output being untrustworthy in a way no gate caught.

benzsevern:
`goldengraph-pipeline` has been red on `main` since #2648, including run #593 this morning — **after** #2672 closed #2668 last night. Those PRs fixed the related zero-config *empty-result* arc (#2663/#2673); they did not fix this. #2668 is reopened.

## What it actually is

Bisected to `6599780c` (#2648) over 72 revisions, holding the `goldengraph-native` wheel fixed so the result isolates the goldenmatch Python change.

**#2648 did not cause it, it exposed it.** The committed configs either side of that commit are identical except blocking passes (2 → 5), and the fixture's `type` column is `'concept'` on all ten rows. Scored at equal weight beside `name`:

```
score = (name + 1.0) / 2     threshold 0.8  ->  the real bar on `name` is 0.6
ensemble("Beta Corp", "Delta LLC") = 0.7037  ->  0.8519    <- false merge
```

Looser blocking made Beta × Delta a **candidate**; the constant column let it **pass**. Two different companies merged into one cluster. Worth stating plainly because it inverts the obvious reading: the resolver was not failing to cluster, it was clustering too much.

### Two hypotheses measured and discarded

Recorded so nobody re-runs them:

- **An ambient `OPENAI_API_KEY`** — refuted: keyed Abt-Buy scores 0.1838, not the 0.5037 baseline.
- **`profile_threshold` making `mass_above_threshold` 1.0 and tripping the precision-collapse guard** — refuted by instrumenting `pick_committed`: `candidates_counted=True`, `admitted_fraction` 0.39–0.42 against a 0.9 floor, so **the guard never fires**.

## The fix

A sibling of #2526's rule, for the other way a scored field can carry no information. There the field varies but blocking guarantees agreement *within a block*; here it does not vary at all.

**It deliberately does not rescale the threshold.** Rescaling preserves the same absolute bar — precisely the bar that produced the merge — so it would be a no-op on every decision the rule exists to change. #2526 rescales because a blocking-agreeing field is informative elsewhere; a data-constant column is informative nowhere, so the honest config is the one auto-config builds when the useless column isn't in the frame: 0.8 on `name` alone. `test_threshold_is_NOT_rescaled` pins it.

**Sample proposes, full frame decides** (review catch, thank you). `n_distinct` is a 1,000-row sample statistic, so a column that is 99.99% one value reads 1 and its field would be dropped — a verdict flipping with *scale*, which the `#876` note 35 lines above forbids, and failing in the same direction as the bug being fixed, since a field that disagrees only on rare pairs carries its whole signal in exactly those pairs. `full_n_distinct` now carries the verdict, computed beside `full_cardinality_ratio` and gated on `n_distinct <= 1`. Demonstrated on one `'us'` row in 200,000, which the sample did miss:

```
tenant   sample n_distinct=1     full_n_distinct=2
fields kept: ['name', 'tenant']
```

## Two gates that could not fire

Both are the shape this PR keeps running into.

- **`goldengraph-pipeline`'s push filter** listed only goldengraph's own paths, so a goldenmatch change never fired it — which is how #2648 went green on its PR and reddened `main` that night. Now includes `packages/python/goldenmatch/goldenmatch/**`: the whole package source rather than an enumerated list, because an enumerated list is the same bug one level up, and the job is ~70 seconds.
- **`bench-suggest-quality` still does not run on this PR.** Its filter covers `core/suggest/**`, `probabilistic.py` and `pipeline.py`, but not `core/autoconfig.py` — arguably the most ER-semantics-heavy file in the package. I ran it by hand (result below); **flagging it as a follow-up rather than widening a second filter in this PR.**

## Fixture re-base

The fix resolves the TINY fixture completely in CI, so the two tests measuring the oracle-vs-real ER gap had no gap left (`assert real_f1 < oracle_f1` → `1.0 < 1.0`). Both assert real ER is *imperfect*. Rather than relax them — weakening ER-quality assertions to accommodate an ER change — they now run on `wikidata_companies_HARD`.

TINY's gap was never robust: its hardest alias is `"Acme"` vs `"Acme Holdings"`, ensemble **0.8615** against a 0.8 cut, so whether it merges turns on whether blocking made the pair a candidate. HARD gives every entity one alias string similarity **cannot** close:

| | score | clear of the cut |
|---|---|---|
| `Helios Industrial Group` / `HIG` — must not merge | 0.5845 | 0.2155 |
| `Meridian Foodworks` / `MFW` — must not merge | 0.5926 | 0.2074 |
| `Helios…` / `Helios… Inc.` — must merge | 0.9643 | 0.1643 |
| worst cross-entity pair | 0.5926 | — |

**Verified the re-base does not weaken the gate:** with the autoconfig fix reverted, the HARD-based suite fails the *same four tests* the nightly failed (4 failed, 8 passed); with it, 12 passed.

## Testing

| check | result |
|---|---|
| CI on `2815442c` | **all six workflows green**, `goldengraph-pipeline` included |
| goldengraph gate | **4 failed → 12 passed**, in a CI-shaped py3.12 polars-free env and a py3.11 polars one |
| goldenmatch suite | green in CI on this commit (the `ci` job runs it sharded) |
| `suggest_quality gate` (run by hand) | **PASS** — 9 ok, **0 fail**, 4 new, 0 missing |
| `anchor_person_match` | 1.0000, **+0.0000** — the dataset #2526 recorded collapsing to 0.7303 on a similar change |
| new tests | 11, incl. `test_threshold_is_NOT_rescaled`, `test_near_constant_column_is_NOT_dropped` |
| `ruff` / `regen_docs --check` / `check_workflow_yaml` | clean |

Quality-gate deltas are against the *committed* scorecard baseline rather than current main, so the `ncvr_synthetic` +0.0125 and `historical_50k` +0.0233 improvements are not necessarily attributable to this change. The load-bearing number is **0 fail**.

## Not explained

CI resolves the old TINY fixture more completely than any local environment I could build — py3.11+polars, py3.12 polars-free, native on and off, eight hash seeds. **The re-base removes the sensitivity, not the cause.**

## Checklist

- [ ] `pre-commit run --all-files` — not run, `pre-commit` is not installed here; `ruff check` run directly and clean
- [x] `regen_docs.py --check` — clean (the codemap regen is commit 2/4)
- [x] Package `CHANGELOG.md` — flagging for review: this changes zero-config ER behaviour and may warrant an entry
- [x] Docs / `CLAUDE.md` updated — rationale lives on the rule, the guard and the filter, next to the code each governs

### DIFF
--- a/.github/workflows/goldengraph-pipeline.yml
+++ b/.github/workflows/goldengraph-pipeline.yml
@@ -12,12 +12,9 @@ name: goldengraph-pipeline
 
 on:
   workflow_dispatch:
-  # Nightly backstop. This lane installs goldenmatch from SOURCE but is
-  # path-filtered on its OWN paths, so a goldenmatch change never triggers it --
-  # a break can sit latent until something unrelated happens to fire the lane
-  # (that gap ran 15 days once). The schedule bounds detection to a day, and a
-  # scheduled run is on `main`, so `main-health` sees the red and files the
-  # tracking issue. Symbol-level breaks are caught earlier, on the PR, by
+  # Nightly backstop. The schedule bounds detection to a day, and a scheduled run
+  # is on `main`, so `main-health` sees the red and files the tracking issue.
+  # Symbol-level breaks are caught earlier, on the PR, by
   # scripts/check_downstream_symbols.py.
   schedule:
     - cron: "13 4 * * *"
@@ -29,6 +26,23 @@ on:
       - "packages/python/goldenmatch/benchmarks/er-kg-bench/erkgbench/qa_e2e/**"
       - "packages/python/goldenmatch/benchmarks/er-kg-bench/tests/test_qa_goldengraph_smoke.py"
       - ".github/workflows/goldengraph-pipeline.yml"
+      # goldenmatch's ER core. This lane installs goldenmatch from SOURCE and its
+      # Phase 1.5 tests assert on what zero-config actually CLUSTERS, so a
+      # goldenmatch change can break it -- and until #2668 the filter listed only
+      # this lane's own paths, so such a change never fired it. That is exactly how
+      # #2648 landed: it edited core/scorer.py, went green on its PR because this
+      # lane did not run, and reddened `main` that night with an over-merge nobody
+      # had a PR signal for. The nightly detected it; detection is not prevention,
+      # and the break still sat in `main` until someone acted on the red.
+      #
+      # Deliberately the WHOLE package source, not an enumerated list of "the ER
+      # files". An enumerated list is the same bug one level up: it goes stale the
+      # moment the blast radius moves, and the thing being guarded here is a
+      # cross-package behavioural coupling nobody can enumerate in advance. The
+      # cost argument does not survive contact either -- the job is ~70 seconds
+      # (run #593: 04:24:09 -> 04:25:18), so this is a cheap, chatty check, and an
+      # absent check is strictly worse than a chatty one.
+      - "packages/python/goldenmatch/goldenmatch/**"
 
 permissions:
   contents: read
--- a/docs/agent-codemap.json
+++ b/docs/agent-codemap.json
@@ -1889,6 +1889,7 @@
             "_diversify_pair_budget_enabled",
             "_diversify_probabilistic_blocking",
             "_diversify_unused_orthogonal_blocking",
+            "_drop_constant_scored_fields",
             "_drop_uninformative_blocking_fields",
             "_embedder_available",
             "_emit_data_profile",
--- a/packages/python/goldenmatch/benchmarks/er-kg-bench/erkgbench/qa_e2e/fixtures/wikidata_companies_HARD.json
+++ b/packages/python/goldenmatch/benchmarks/er-kg-bench/erkgbench/qa_e2e/fixtures/wikidata_companies_HARD.json
@@ -1,0 +1,18 @@
+{
+  "meta": {
+    "source": "wikidata",
+    "pulled": "test",
+    "sparql_sha": "test",
+    "domain": "companies",
+    "why": "TINY's hardest alias is 'Acme' vs 'Acme Holdings' -- a NEAR MISS that lands within a few hundredths of the 0.8 cut, so whether it merges flips on small environment differences (#2668: it fragments locally and resolves in CI, from the same commit). A fixture whose ER gap sits on a threshold boundary measures the environment, not the resolver. Here every entity carries one alias that string similarity CANNOT close -- an acronym or a rename -- so the oracle-vs-real gap is structural and stays stable as string matching improves. Each entity also carries a suffix variant that SHOULD resolve, so 'some variants merged' and 'some entity stayed fragmented' are both true by construction rather than by luck. Canonicals are lexically far apart so no cross-entity merge is plausible."
+  },
+  "entities": [
+    {"qid": "Q1", "canonical": "Helios Industrial Group", "aliases": ["Helios Industrial Group Inc.", "HIG"]},
+    {"qid": "Q2", "canonical": "Meridian Foodworks", "aliases": ["Meridian Foodworks Limited", "MFW"]},
+    {"qid": "Q3", "canonical": "Norvex Systems", "aliases": ["Norvex Systems Ltd"]},
+    {"qid": "Q4", "canonical": "Ostara Media", "aliases": ["Ostara Media LLC"]}
+  ],
+  "facts": [
+    {"anchor_qid": "Q1", "relation": "has_subsidiary", "member_qids": ["Q2", "Q3", "Q4"]}
+  ]
+}
--- a/packages/python/goldenmatch/benchmarks/er-kg-bench/tests/test_realworld_aggregation.py
+++ b/packages/python/goldenmatch/benchmarks/er-kg-bench/tests/test_realworld_aggregation.py
@@ -103,13 +103,25 @@ def test_real_resolution_clusters_variants_and_is_not_oracle():
     the surface universe -- it merges >=1 entity's distinct surfaces into one shared key
     (variants resolved), it is NOT the oracle km (a conservative real resolver keeps more
     keys than the entity count), and on this clean fixture it makes no wrong cross-entity
-    merge."""
+    merge.
+#
+# FIXTURE (#2668): HARD, not TINY. The two assertions below need real ER to leave
+# something unresolved, and on TINY the only thing it could ever leave unresolved
+# was "Acme" vs "Acme Holdings" -- ensemble 0.8615 against a 0.8 cut, so it merges
+# or not depending on whether blocking happened to make the pair a candidate. That
+# flipped between environments on one commit (fragmented locally, resolved in CI),
+# which made these tests measure the environment rather than the resolver. HARD
+# gives every entity one alias string similarity CANNOT close (an acronym: 0.58
+# against the same cut, 0.21 clear) plus a suffix variant that plainly should
+# resolve (0.94-0.96, 0.14+ clear). Every decision now sits far from the boundary,
+# so the oracle-vs-real gap is structural and does not erode as matching improves.
+    """
     _wheel_or_skip()
     from collections import defaultdict
 
     from erkgbench.qa_e2e.realworld import _realworld_entity_surfaces, _resolution_km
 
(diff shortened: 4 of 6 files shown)


benzsevern:
Root cause looks right, and the refutation of the `mass_above_threshold` theory is the useful part. I chased the same hypothesis off the issue text before finding this PR. `admitted_fraction` 0.39-0.42 against a 0.9 floor settles it.

One problem with the fix as written.

**`n_distinct` is a 1,000-row sample statistic, and the drop treats it as exact.** `profile_columns(df, sample_size: int = 1000)`, and all four call sites take the default. So a column that's 99.99% one value reads `n_distinct == 1` and its field gets dropped from the matchkey.

Miss probability for a value at frequency `p` over 1,000 sampled rows:

| p | rows carrying it at 5M | P(sample misses it) |
|---|---|---|
| 1e-3 | 5,000 | 37% |
| 1e-4 | 500 | 90% |
| 1e-5 | 50 | 99% |

The verdict then flips with scale rather than with the data: a 2,000-row frame samples that column fully and keeps the field, a 5M-row frame reads it constant and drops it. `autoconfig.py` already has that exact bug written up 40 lines above your hunk, in the "#876 surrogate guard" note where `cardinality_ratio` read 0.72 / 0.94 / 0.97 / 0.98 / 1.00 at 100K through 5M on a column whose true ratio was flat 0.28. That comment calls a scale-dependent verdict forbidden, and the remedy it landed on is sitting right there: an EXACT full-frame pass for the columns the sample flags.

Direction of the damage is the same one this PR exists to fix. A field that agrees on nearly every pair and disagrees on the rare ones is carrying its whole signal in those rare pairs. Dropping it removes the penalty exactly where it discriminates, so the rare cross-entity pair merges.

Suggest gating the drop behind a confirm: when the sample says `n_distinct <= 1`, run `df[col].n_unique()` on the full frame before dropping. Only fires for the handful of apparently-constant columns, so it's one cheap pass per candidate rather than a full re-profile. Sample says "candidate", full frame decides.

Two things I checked that are NOT problems, recorded so nobody re-runs them:

- The native and pure-Python branches compute `n_distinct` differently on the surface (`len({v for v in _vals if v.strip()})` vs `len(set(values))`), but `values` is already blank-filtered at line 536 and `names_to_sample_values` maps to that same list, so they agree. No parity gap.
- `_drop_constant_scored_fields` guards the empty-matchkey case and the `None` case correctly.

Not blocking, it's still a draft. The constant-column mechanism holds either way.

benzsevern:
You're right, and the precedent you point at is 35 lines above my hunk — I read that `#876` note while working in this function and still wrote the same bug underneath it. Fixed as you suggested: sample proposes, full frame decides.

**What changed**

`n_distinct` is now documented as a *candidate signal only*, and a new `full_n_distinct` carries the verdict — the exact mirror of `full_cardinality_ratio`, computed in the same pass, gated on `n_distinct <= 1` so only apparently-constant columns pay for a scan. `_drop_constant_scored_fields` reads `full_n_distinct` exclusively; absent means "cannot answer" and nothing is dropped.

Restricting the exact count to sample-flagged columns is sound rather than a cost heuristic, same argument as yours: a column with >1 distinct value in the sample has >1 in the full frame, so it cannot be constant and is skipped.

**Demonstrated on the shape you described** — one `'us'` row in 200,000, which a 1,000-row sample misses ~99.5% of the time:

```
  name     sample n_distinct=1000  full_n_distinct=None
  tenant   sample n_distinct=1     full_n_distinct=2
  fields kept: ['name', 'tenant']
  VERDICT: tenant KEPT (guard worked)
```

The sample did miss it. Before this change that field was dropped; now it survives. Two tests pin it: `test_near_constant_column_is_NOT_dropped` (sample says 1, full frame says 2, field kept) and `test_sample_flag_alone_never_decides` (sampled 1 with no full count is "cannot answer", not "constant").

Your point about the direction of the damage is the part I'd have missed on my own, so it's in the code comment rather than just the commit: a field that agrees on nearly every pair and disagrees only on the rare ones is carrying its whole signal in exactly those rare pairs, so dropping it removes the penalty precisely where it discriminates — the same failure this rule exists to prevent, one level down.

Thanks also for recording the two non-problems. I'd have re-run the native/pure-Python `n_distinct` comparison otherwise; agreed that `values` is already blank-filtered at line 536 and `names_to_sample_values` maps to that same list.

**Re-validated after the change:** goldengraph gate 12 passed in a CI-shaped py3.12 polars-free env, `ruff` clean, no docs drift (the codemap tracks symbols, and this adds only a field). Full-suite before/after and the `suggest_quality` gate are re-running now — I'll post if either moves.

One thing still open and not caused by this: CI resolves the old TINY fixture more completely than any local environment I could build (py3.11+polars, py3.12 polars-free, native on and off, eight hash seeds). That's why the two ER-gap tests moved to a HARD fixture whose margins are ≥0.13 from the cut instead of TINY's 0.06. The re-base removes the sensitivity, not the cause.

---

PR #2711, opened 2026-08-21
fix(ci): the two bespoke publish workflows had the same metadata break

benzsevern:
`_publish-pypi.yml` was fixed in #2710. The six packages that delegate to it are covered. `golden-suite` and `goldensuite-mcp` do **not** delegate -- each carries its own tag-vs-pyproject version check -- so both kept the stale-validator failure.

## Verified, not assumed

Building `golden-suite` 0.5.1 locally:

```
Successfully built golden_suite-0.5.1.tar.gz and golden_suite-0.5.1-py3-none-any.whl
Checking dist/golden_suite-0.5.1-py3-none-any.whl: PASSED
Checking dist/golden_suite-0.5.1.tar.gz: PASSED
Metadata-Version: 2.5
```

That is the exact pair -- a 2.5 wheel and the action's bundled validator -- that failed `goldenanalysis-v0.5.0` (run 32507401780).

## Why now

`golden-suite` 0.5.1 is genuinely unpublished (PyPI has 0.5.0). It is the one package in the tag back-fill that will really upload, and it uploads through this workflow.

## One detail worth reading

The `twine check` path is `dist/*`, not the full path: both jobs set `defaults.run.working-directory` to the package dir. The `packages-dir` on the publish step stays fully qualified, because `uses:` steps do not honour working-directory. Getting this backwards is what the #2710 dry run caught.

### DIFF
--- a/.github/workflows/publish-golden-suite.yml
+++ b/.github/workflows/publish-golden-suite.yml
@@ -48,10 +48,25 @@ jobs:
           fi
           echo "Version OK: $PKG_VERSION"
 
-      - run: pip install build
+      - run: pip install build twine
       - run: python -m build
+      # Validate with a CURRENT twine, not the one bundled inside the publish
+      # action. Modern build backends emit `Metadata-Version: 2.5`; the
+      # action's bundled twine carries a `packaging` old enough to reject it
+      # outright --
+      #   InvalidDistribution: '2.5' is not a valid metadata version
+      # -- which is what failed goldenanalysis-v0.5.0 (run 32507401780). PyPI
+      # itself accepts 2.5, and current twine passes the wheel, so the wheel is
+      # fine and the stale validator is not. Same fix as _publish-pypi.yml;
+      # this workflow is not a caller of that template (it carries its own
+      # tag-vs-pyproject version check), so it needed the fix separately.
+      - name: Validate distribution metadata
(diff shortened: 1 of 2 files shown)


benzsevern:
Correction to this PR's own reasoning, recorded after using the repaired path.

`golden-suite-v0.5.1` published successfully (run 32511971803) and the wheel PyPI now serves carries `Metadata-Version: 2.5`. That run went through the action with `verify-metadata: true` and the same pinned digest that rejected `goldenanalysis-v0.5.0` with `InvalidDistribution: '2.5' is not a valid metadata version`.

So the two outcomes diverge on the same action pin and the same metadata version, and I cannot account for why from the logs. What this PR claimed -- that both bespoke workflows carried a break certain to fire -- was stated with more confidence than the evidence supports.

What still holds: validating with a current twine and not letting the action re-check with its bundled copy makes the outcome deterministic instead of dependent on whatever that image resolves at run time. The change is right; the certainty in the description was not.

Also worth recording, because it caught me out here: a `release: published` event runs the workflow file **from the tag's commit**, not from main. `golden-suite-v0.5.1` points at c6247768b, which predates this PR, so the run above executed the OLD workflow -- the fix in this PR was not what let it through. Any back-filled tag cut at an older commit will likewise run that commit's workflow.

Issue #2717, opened 2026-08-21
Product-text matching collapses on Abt-Buy and Amazon-Google (F1 0.17 / 0.10, both commit a RED config)


benzsevern:
Split out of #2457, which was a bot-maintained lane tracker carrying a real quality bug it could never close. #2470 tracked this and is **closed**, so as of now nothing open describes it.

## Measured, nightly, reproducible

From run [32457009104](https://github.com/benseverndev-oss/goldenmatch/actions/runs/32457009104) (2026-08-21), and every scheduled run since 2026-08-18:

| dataset | F1 | precision | recall | health | stop_reason |
|---|---|---|---|---|---|
| Abt-Buy | 0.1723 | 0.1068 | 0.4463 | **RED** | budget_iterations |
| Amazon-Google | 0.1014 | 0.2084 | 0.0670 | **RED** | budget_time |

For contrast, the same run: DBLP-ACM 0.9640, Febrl3 0.9912, NCVR-synthetic 0.9925. **The failure is specific to messy product text**, not to the lane.

Both commit a RED config, so the F1 numbers are not even trustworthy -- auto-config never converged on a usable config, it exhausted its budget and committed a best-effort guess.

## Two distinct problems, and the second may be the cause

1. **Quality.** F1 0.10-0.17 on product matching. Amazon-Google recall is **6.7%**.
2. **The controller exhausts its budget** on both -- `budget_time` on one, `budget_iterations` on the other. Whether the quality is bad *because* the controller never converges, or the controller thrashes *because* the data defeats it, is not established. That is the first thing to measure, because it decides which problem to work on.

## Already ruled out -- do not re-derive

From the `Abt-Buy` note in `_F1_FLOORS` (scripts/run_benchmarks.py), all on the same 1118 ground-truth pairs:

```
committed baseline (2026-08-11, native=1)  0.5037  P 0.8219   494 pairs
CI nightly, native=0, no key               0.1723  P 0.1068
local HEAD, native=0, no key               0.1723  P 0.1068  4673 pairs
local HEAD, native=1, no key               0.1723  P 0.1068
local 8145b498 (the baseline's OWN commit) 0.1723  P 0.1068
local HEAD, native=0, WITH an LLM key      0.1838  P 0.1132
```

[...]

PR #2728, opened 2026-08-22
fix(blocking): make the matchkey recall denominator reachable, and stop trading recall away silently (#2717)

benzsevern:
Two fixes that are only correct **together**. Either alone regresses DBLP-ACM — measured against ground truth, not argued.

## 1. The good denominator was unreachable

`_target_pairs_from_matchkey` built row ids with `sample_frame.native.with_columns(...)` — a **polars** call. On the arrow-native lane `native` is a `pa.Table`, so it raised, `_build_recall_target` caught it, and **every recall estimate silently came from the character-similarity proxy** instead.

That proxy's own docstring calls it weak: on Amazon-Google it selects 2,355 sample pairs of which 35 are true (**1.5% precision**), *"so a candidate is judged largely on how many NON-matches it co-blocks"*. Nothing failed loudly — the warning names a matchkey, not a lane.

Routed through the seam's `ensure_row_ids`, which both lanes implement.

### What that changes, on real DBLP-ACM vs `DBLP-ACM_perfectMapping.csv`

| denominator | candidate | estimate | **TRUE recall** |
|---|---|---|---|
| weak proxy | `venue[:3]` | 0.2032 | 0.0594 |
| weak proxy | `title[:5]` | 0.0407 | **0.9820** |
| matchkey | `venue[:3]` | 0.3913 | 0.0594 |
| matchkey | `title[:5]` | **0.9565** | **0.9820** |

The proxy is not merely noisy, it is **anti-correlated** — it rates the key retaining 5.9% of true pairs 5× above the one retaining 98.2%. The matchkey denominator is **63.8% true pairs** against the proxy's 1.5%, and its estimates are stable across sample sizes (0.9565 at n=1000, 0.9535 at n=4910).

## 2. Ranking traded that recall away

`score * estimated_recall` lets a cheap plan outrank a much better one. On DBLP-ACM `title[:5]+authors[:5]` (1,781 comparisons) outranked `title[:5]` (39,648) while retaining **0.4164** of true pairs against **0.9820**.

`_apply_recall_tradeoff_gate` refuses a plan retaining under `_RECALL_TRADEOFF_RATIO` of the best available. The constant and its justification already existed — it only ever produced a warning.

## Why the coupling matters, and how I know

**I implemented the gate first, over the broken denominator, and measured it.** It promotes `venue[:3]` and takes DBLP-ACM from **98.2% recall to 5.9% at 61× the comparisons**. It was reverted, and `_apply_recall_tradeoff_gate`'s docstring records that so nobody enables it against the proxy.

**Result:** DBLP-ACM chooses `title[:5]` — est 0.9275, **TRUE recall 0.9820**, 39,648 comparisons.

## Conservatism, and the tests that pin it

Ordering **within** the qualifying set is untouched, so cost still decides among plans that clear the bar. Gated candidates are demoted, not dropped. If nothing clears the bar — or recall could not be estimated at all — the list is returned unchanged. Each of those has a test; without them the gate could silently starve the caller of candidates.

6 new tests; `test_config_critique` + the numeric-component suite: 37 passed.

## Reviewer note

This changes candidate ranking for every dataset. I validated on DBLP-ACM with real ground truth; **Amazon-Google and Abt-Buy are quarantined (#2718) so a regression there will report rather than fail.** Worth a scheduled `benchmarks` run before treating it as settled.

### DIFF
--- a/docs/agent-codemap.json
+++ b/docs/agent-codemap.json
@@ -2400,6 +2400,7 @@
           "defines": [
             "BlockingSuggestion",
             "_apply_candidate_transforms",
+            "_apply_recall_tradeoff_gate",
             "_build_recall_target",
             "_mean_token_count",
             "_retention",
--- a/packages/python/goldenmatch/goldenmatch/core/block_analyzer.py
+++ b/packages/python/goldenmatch/goldenmatch/core/block_analyzer.py
@@ -520,26 +520,46 @@ def _target_pairs_from_matchkey(sample_frame: Any, matchkey: Any) -> set[tuple[i
     per-row lists directly.
     """
     from goldenmatch.core.frame import to_frame
-    from goldenmatch.core.scorer import find_fuzzy_matches
+    from goldenmatch.core.scorer import score_pair
 
-    # Row ids through the seam, not `native.with_columns` (#2717). That is a
-    # POLARS call, and on the arrow-native lane `native` is a `pa.Table`, which
-    # raises `AttributeError: 'pyarrow.lib.Table' object has no attribute
-    # 'with_columns'`. `_build_recall_target` catches that and silently drops to
+    # POLARS-FREE, deliberately. The previous implementation built row ids with
+    # `sample_frame.native.with_columns(pl.Series(...))` and handed the frame to
+    # `find_fuzzy_matches`, which is polars-typed (`block_df: pl.DataFrame`).
+    # On the arrow-native lane `native` is a `pa.Table`, so that raised, and
+    # `_build_recall_target` silently fell back to
     # `_target_pairs_from_similarity` -- the proxy its own docstring calls WEAK
-    # (on Amazon-Google: 2,355 sample pairs, 35 true, 1.5% precision, so a
-    # candidate is judged mostly on how many NON-matches it co-blocks).
+    # (Amazon-Google: 2,355 sample pairs, 35 true, 1.5% precision). Every recall
+    # estimate on the default lane came from it, and the estimates it produces
+    # are ANTI-correlated with true retention: on DBLP-ACM it rates `venue[:3]`
+    # (5.9% of true pairs) above `title[:5]` (98.2%).
     #
-    # So the good denominator was unreachable on the default lane, and every
-    # recall estimate there came from the weak one. Nothing failed loudly: the
-    # warning names a matchkey, not a lane, and the analyzer carried on.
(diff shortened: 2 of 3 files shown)


benzsevern:
Closing: measured, and it does not earn its cost. The diagnosis is on #2717; the branch stays as the record.

## What the change did

Two coupled fixes -- make the matchkey recall denominator reachable on the arrow lane (it crashed and silently fell back to a proxy), and stop ranking from trading recall away.

## Why it should not land

Measured on real DBLP-ACM against `DBLP-ACM_perfectMapping.csv`:

| version | chosen | TRUE recall | wall |
|---|---|---|---|
| `origin/main` | `title[:5]` | **0.9820** | 21.1s |
| this branch, n=1000 | `title[:5]` | **0.9820** | 38.6s |
| this branch, n=400 cap | `title[:5]+authors[:5]` | **0.4164** | 4.8s |

**`main` already picks the right key**, despite an estimate of 0.0407 from the broken proxy -- the comparison-count term dominates and rescues it. So the fix buys no quality on this dataset and costs 1.8x.

And it cannot simply be made cheaper. `score_pair` is a Python call per pair, so the polars-free target build is O(n^2) in the interpreter where the polars-typed `find_fuzzy_matches` it replaces is vectorised (rapidfuzz cdist): 44.1s at n=1000, 14.2s at n=600, 6.1s at n=400. 44s crashed four xdist workers on the `allow_red_config` at-scale tests, which already sat at 107-110s against a 120s timeout. Capping to 400 makes it affordable and leaves only **17 target pairs** -- too thin to discriminate, so the ratio gate stops firing and the pick regresses.

## The blocker, named

`find_fuzzy_matches(block_df: pl.DataFrame, ...)` is vectorised **and** polars-typed. Staying polars-free forces the slow path; going fast forces polars back in, which is not on the table. **Porting that function off `pl.DataFrame` is the prerequisite**, and with it this change becomes both correct and affordable.

## What was worth finding

- The matchkey denominator has been unreachable on the arrow lane -- `sample_frame.native.with_columns(pl.Series(...))` on a `pa.Table` raises, and `_build_recall_target` swallows it.
- The proxy it falls back to is **anti-correlated** with true retention, not merely noisy.
- On Amazon-Google `main` commits `title[:5] + manufacturer[:5]`, estimate 0.3921, **true blocking recall 0.0408**. That is #2717's F1 0.10 explained: no scorer recovers pairs blocking never emitted.

Issue #844, opened 2026-06-10
Phase 5 distributed: __cluster_id__ assigned per row, not per cluster (all output singletons at 346M)


benzsevern-mjh:
## Summary

Running `run_dedupe_pipeline_distributed` (Phase 5, `GOLDENMATCH_DISTRIBUTED_PIPELINE=2`) on a 346M-row input emits output where every record has a unique `__cluster_id__`. `_join_assignments_distributed` appears to be setting `__cluster_id__` equal to each row's `__row_id__` instead of the post-union-find cluster id. Reproduced at v1.30.0 and v1.25.0 (the v1.25 run had a column-name bug in my analysis script that hid this until v1.30).

## Environment

- `goldenmatch` 1.30.0 (commit `2b59828f6bc12c0c863faecbd15a037b63c878ad`)
- `goldenmatch-native` 0.1.5 (PyPI wheel)
- `ray[data]` 2.49.0 in `rayproject/ray:2.49.0-py312` docker image
- 1 head + 5 workers (`e2-standard-16`), 96 vCPU / ~280 GB RAM / ~120 GB plasma total
- Env: `GOLDENMATCH_DISTRIBUTED_PIPELINE=2`, `GOLDENMATCH_FS_NATIVE=1`, `GOLDENMATCH_FS_AUTOCONFIG_V2=1`, `GOLDENMATCH_NATIVE=1`, `GOLDENMATCH_ENABLE_DISTRIBUTED_RAY=1`
- `DataContext.shuffle_strategy = "hash_shuffle"` (required for `Dataset.join`)
- `DataContext.target_max_block_size = 32 MiB`
- Input: 346M rows, hash-shuffled (`__row_id__ % 17`) into 17 source-balanced parquet buckets on GCS, all rows carry a global `__row_id__`

## Repro shape

```python
from goldenmatch.distributed import read_parquet_partitioned
from goldenmatch.distributed.pipeline import run_dedupe_pipeline_distributed

ds = read_parquet_partitioned(paths, n_partitions=256)  # also tested 1024
result = run_dedupe_pipeline_distributed(
    ds,
    config=build_raw_c_config(),     # explicit, honored per #757
    confidence_required=False,
    allow_red_config=True,
    output_path="gs://.../golden/",
)
```

[...]

PR #2730, opened 2026-08-22
fix(#2717): stop re-cutting on a truncated distribution, apply exclusions to both match frames, report both benchmark lanes

benzsevern:
**The port I called a prerequisite was not one.** `find_fuzzy_matches` already accepts a `pa.Table` — its body reads through the `_to_frame_d5` seam and branches on `to_dicts`/`to_pylist`. **Only the type annotation said polars**, and that alone was enough to make this caller convert and to make me conclude a large port was needed.

Verified byte-identical on the real DBLP-ACM matchkey including the composite `ensemble` scorer:

```
polars:  111 pairs in 0.60s
arrow :  111 pairs in 0.37s
identical pair sets: True     max score delta: 0.00e+00
```

Arrow is *faster*. Annotation widened to `Any`, with that recorded next to it.

## The bug

`_target_pairs_from_matchkey` built row ids with `sample_frame.native.with_columns(pl.Series(...))` — polars, on a `pa.Table`. It raised, `_build_recall_target` caught it, and **every recall estimate on the default lane silently came from the character-similarity proxy**. Nothing failed loudly: the warning names a matchkey, not a lane.

That proxy is **anti-correlated** with truth, not merely noisy:

| denominator | candidate | estimate | **TRUE recall** |
|---|---|---|---|
| weak proxy | `venue[:3]` | 0.2032 | 0.0594 |
| weak proxy | `title[:5]` | 0.0407 | **0.9820** |
| matchkey | `venue[:3]` | 0.3913 | 0.0594 |
| matchkey | `title[:5]` | **0.9565** | **0.9820** |

`block_analyzer.py` now makes **zero** polars calls (it made one before). It drops any existing `__row_id__` first, because `ensure_row_ids` *reuses* one (#844) and this function's contract is that ids are positional into `sample_frame` — a test pins that.

**Rejected on measurement:** a `score_pair` loop, the obvious polars-free route, costs **44.1s** at n=1000 against 0.37s here, and crashed four xdist workers on the `allow_red_config` at-scale tests.

## The ranking gate, now that the estimate is trustworthy

`score * estimated_recall` let a cheap plan outrank a better one. The gate refuses a plan under `_RECALL_TRADEOFF_RATIO` of the best available — a constant that already existed and only ever warned.

**It is only sound with the denominator fix.** Applied over the weak proxy it promotes `venue[:3]` and takes DBLP-ACM from 98.2% to 5.9% recall at 61× the comparisons. That was built, measured against ground truth, reverted, and recorded in the gate's own docstring.

## Measured outcome

| dataset | version | chosen | **TRUE recall** | wall |
|---|---|---|---|---|
| DBLP-ACM | main | `title[:5]` | 0.9820 | 21.1s |
| DBLP-ACM | **this** | `title[:5]` | **0.9820** | **~1s** |
| Amazon-Google | main | `title[:5]+manufacturer[:5]` | 0.0408 | 18.6s |
| Amazon-Google | **this** | `manufacturer[:5]` | **0.0715** | **0.4s** |

No regression, and 18.6s → 0.4s.

## What this does NOT fix

Amazon-Google reports `est=1.0000` against a **true 0.0715**. Its matchkey target is **375 pairs of which one is true** (0.27% precision); DBLP-ACM's is 69 pairs / 44 true (63.8%). The recall estimate can only be as good as the matchkey it is measured against, and on product text the **scorer** is the problem.

**#2717 is not closed by this.** Blocking selection cannot be fixed independently there.

---

# Update: the rest of #2717, and it closes it

Three more defects, all in the path between "the config auto-config committed" and "the pairs the run emits". Each was measured with a committed harness, `scripts/diagnose_product_lanes.py`, which attributes a lane's F1 across blocking / scoring / emission in one run.

## 1. Postflight re-cuts on a distribution truncated at the cut

The weighted block scorer applies `mk.threshold` itself and returns only survivors, so the pair list `postflight` receives **starts at the cut**. Every valley findable in it lies above the cut, so the auto-adjustment can only ever ratchet the threshold **up** — never down, never back — and it never appears in the committed config.

Amazon-Google, linkage lane. The committed 0.65 was **already the F1 optimum** of the candidate set:

| cut | precision | recall | F1 |
|---|---|---|---|
| 0.60 | 0.3348 | 0.5769 | 0.4237 |
| **0.65** | 0.3834 | 0.4854 | **0.4284** ← committed |
| 0.70 | 0.4274 | 0.3938 | 0.4099 |
| 0.95 | 0.4626 | 0.0523 | 0.0940 |

Postflight read the clipped tail as bimodal, found a "valley" at 0.965, and re-filtered there. **1646 qualifying pairs became 120.**

So the threshold half of this issue was never a threshold-derivation problem. The controller picks the right number; a later stage overrides it on invalid evidence. Postflight now refuses when the lowest scored pair is at or above the current threshold, and says so in an advisory. The signal stays live wherever pairs genuinely span the cut — a probabilistic matchkey scores down to the review threshold.

Same family as measuring a posterior split on `scored_pairs`, which is also truncated at the cut.

## 2. Auto-config exclusions were applied to the target frame only

In match mode the GoldenCheck detectors ran on `df`, dropped the excluded columns from it, and left `reference` untouched. The two frames then reached `run_match_df` at different widths, and its polars lane concatenates with a strict `pl.concat`:

```
ShapeError: unable to append to a DataFrame of width 5 with a DataFrame of width 6
```

**Every** controller iteration raised it, so the run fell back to v0 + a RED sentinel. The arrow lane hid it — `pa.concat_tables(promote_options="permissive")` backfills the missing column with nulls, so this never surfaced as a test failure. The detectors flagged `description` on both frames independently; only the drop was one-sided.

## 3. The benchmarks measured dedupe and compared it to published linkage figures

67.5% of the dedupe lane's candidate pairs on Amazon-Google are same-source, and against a cross-source mapping those cannot be true matches — they land entirely in the false-positive column. The engine already prints a warning naming this and recommending `match_df(left, right)`.

Each product dataset now emits **both** rows. The dedupe row keeps its exact name, so `_F1_FLOORS`, `_QUARANTINE` and the committed series all still apply to the same measurement; the `(linkage)` row is new and deliberately unfloored until CI publishes a baseline.

## Measured outcome

Zero-config `match_df`, end to end:

| dataset | before | after | precision |
|---|---|---|---|
| Amazon-Google | 0.0761 | **0.4636** | 0.4500 → 0.5961 |
| Abt-Buy | 0.1723 | **0.5658** | 0.1068 → 0.9163 |

## A standing mystery, resolved

`_F1_FLOORS` records Abt-Buy's 0.45 floor as **DISPUTED**, derived from a 0.5037 / P 0.8219 / **494 emitted pairs** baseline that five separate reproductions could not reach — every one landing at 0.1723 / P 0.1068 / ~4673 pairs.

The linkage lane reproduces it: **0.5658 / P 0.9163 / 490 pairs.**

The floor was derived from a linkage run and then enforced against a dedupe run. Not a regression, not unreproducible — the two numbers were never comparable. Recorded in the table.

## Scope note

The dedupe lane improves too (0.0888 → 0.1097, recall 0.0592 → 0.5551) but stays quarantined: at 68.5% same-source candidates its ceiling is structural, and changing what that benchmark *is* is a separate decision.


benzsevern:
Second commit pushed. This PR now carries the blocking half of #2717 end to end.

## Headline

**Amazon-Google blocking recall 0.0408 -> 0.9531**, measured against `Amzon_GoogleProducts_perfectMapping.csv` through the real `build_blocks`, on the **default path with no env vars**. DBLP-ACM is untouched (`static` on `__title_key__`, its 0.9820-recall key).

## Token blocking was never broken -- it was unreachable

`block_analyzer.py:140` said *"`build_token_blocks` never logs, and the committed RED config yields no candidate pairs at all. That integration gap is not diagnosed yet"* -- and turned the strategy off on that basis.

The observation was right, the inference wasn't. Nothing ever set `strategy="token"`:

1. `auto_suggest` defaults to `False`, so `_run_auto_suggest` returns on its first line;
2. its token branch sits behind `if not config.blocking.keys`, which auto-config has already populated.

Absence of logging meant *never ran*, not *failed*. Invoked directly the blocker produces 2,718 blocks / 94,938 pairs at df<=50.

## The change

`defer_free_text_blocking_to_analyzer` clears both locks for the one case the codebase already calls near-useless -- *"a prefix key on a product title is a near-useless block"*. When a key field is free text by the **measured** token-count bar, it returns `keys=[], auto_suggest=True` and lets the analyzer (which the first commit made trustworthy) decide.

Narrow by construction: names/addresses sit at 2-4 tokens, below the 5.0 bar; titles at 8-30. ANY field rather than EVERY, because compounding a free-text prefix with a short one doesn't rescue it. Deliberate strategies (canopy/ann/lsh/learned) are never overridden. Fails open.

`_token_candidates_enabled` now defaults **ON**, since the measurement its OFF default rested on is refuted.

## What cost the time, recorded so the next person skips it

I first wired only `_legacy_auto_configure_v0` and saw no effect. **Zero-config routes through `auto_configure_probabilistic_df`** -- that's the `build_blocking` call that runs. Two of my intermediate "it doesn't fire" results were worthless: an instrumenting script printed success without verifying its replacement matched, and the autoconfig memory cache was returning a config without running v0 at all.

## Scope

**This does not close #2717.** The 0.80 matchkey threshold is a second, independent cap (scorer recall 0.214 where 0.65 gives 0.510) -- quantified on the issue, not addressed here.

8 new tests, 54 passing across the touched suites. One of them re-validates the deferred config, because `static`/`multi_pass` normally *require* keys and this empties them -- legal only via the validator's `auto_suggest` exemption, which the test pins.

PR #2757, opened 2026-08-25
fix(layers): close the measured error in identity-layer detection (#2574 Part A)

### LINKED ISSUE #2574: Automatic identity-layers detector: detect the entity roles present in a dataset (bank / lender / payor / debtor / machine / person / company)
## Idea

Given a dataset, automatically detect **what kinds of entities (roles) it contains** — e.g. a transactions table that references banks, lenders, payors, debtors, machines, people, and companies simultaneously. Today nothing produces this: the closest capability (`infermap.detect_domain`) picks **one industry vertical for the whole dataset**, winner-take-all, from **column names only**.

This is the detector that makes heterogeneous-data handling (incl. #2575 per-block config) actionable — you can't sensibly vary matching config per segment until you can detect the segments.

## Grounded state of the code (assessed 2026-08-14)

**Exists / reusable:**
- `packages/python/infermap/infermap/detect.py` — `detect_domain(df)` scores 16 verticals by name-hint coverage fraction, returns a single winner, refuses on ties. **Name-only** — values never inspected during detection.
- Domain packs (`packages/python/goldencheck-types/goldencheck_types/_domains/*.yaml`) already carry per-type `name_hints` **and** `value_signals` — e.g. `finance.yaml` `merchant: [merchant, vendor, payee, counterparty]`. The raw signal exists; it's just consumed by the *mapping* scorers, not by *detection*.
- The M×N scorer engine (`infermap/engine.py`) can already label each column with a canonical type within a chosen domain.
- `infermap/identity.py` already turns ID-columns into per-record identity aliases keyed by "kind" (`customer_id`, `ssn`, `ein`, `vin`, `npi`…).

**Gaps (the actual net-new work):**
1. **Roles are not first-class.** Packs model *field types* (`account_number`, `merchant`), not *entity roles/parties* (bank vs debtor vs payor). "merchant/payee/counterparty" is one flat field type. No taxonomy of entity kinds crossing verticals (person / company / machine).
2. **One-domain-per-dataset, winner-take-all** (`detect.py:127-128`). No mechanism to report "contains multiple entity types simultaneously." A **multi-label** detector is net-new.
3. **1:1 assignment constraint** (`engine.py:373`, deliberate non-goal in `docs/superpowers/specs/2026-03-29-infermap-design.md`). Two columns both referencing "bank," or one table hosting several roles, collides with the 1:1 model.
4. **Detection is name-only** — role detection ("9-digit numeric column → routing number → *bank*") needs value signals wired into a detector.
5. No spec, API, or tests anywhere in the monorepo. Name-collision caveats: goldengraph's `entity-type-canon` is KG/LLM extraction (different subsystem); the six-vertical roadmap "Identity Layer" (M1) means persistent agent state, unrelated.

## North Star fit

Strong. Serves **zero-config** (heterogeneous frames are exactly where single global config underperforms) and **approach-the-expert**, and must stay **never-black-box** (per-column role + evidence, mirroring the existing `DetectionResult.reason`).

## Architectural placement

Belongs in **infermap** (schema/type inference is its authoritative surface per decision 0047 — one authoritative semantic owner per capability), emitting a **multi-label role annotation** that goldenmatch / goldenpipe consume. Not a new subsystem.

## Anchor files
`infermap/detect.py`, `infermap/domain_pack.py`, `infermap/engine.py`, `infermap/identity.py`, `goldencheck-types/goldencheck_types/_domains/*.yaml`, `goldencheck-types/goldencheck_types/types.py`, consumer `goldenpipe/goldenpipe/stages/infer_schema.py`.

## Next gate

Design spec via the brainstorming gate (`docs/superpowers/specs/`): multi-label role taxonomy + value-aware detection, per-column role output with evidence. Lower-risk / higher-fit of the two ideas — recommended to spec first.

benzsevern:
Part A of the plan to close #2574 (`docs/superpowers/plans/2026-08-24-2574-identity-layers-close.md`, included here and annotated with what actually happened).

## Result

| corpus | before | after |
|---|---|---|
| FK ground truth — detectable conventions, exact partition | 92% | **100%** |
| FK ground truth — `initial` convention, exact partition | **0%** | **62%** |
| FK ground truth — `initial` convention, pairwise F1 | 0.08 | **0.88** |
| FK ground truth — `initial` mean party-count error | 2.62 | **0.38** |
| precision corpus — specificity | 100% | 100% |
| precision corpus — sensitivity | 100% | 100% |

## Why this is not the work the issue asks for

The issue's status comment says gap 4 — "detection is name-only" — is "the load-bearing gap", and that pack `value_signals` already carry the raw signal. I re-verified both halves against `main` before planning, and both are wrong.

**The value vocabulary does not exist.** Pack `value_signals` are profiling *shape* hints (`min_unique_pct`, `short_strings: true`, `mixed_case: true`). Nothing there tells a routing number from any other short string. The only value-pattern vocabulary in the repo is `pattern_type.py`'s eight `SEMANTIC_TYPES`, none of which identifies a party. It would have to be built, not wired.

**The measured residual was name-side, and values fix neither failure.** Running the harness on `main` showed exactly two failure modes, both root-caused here, and neither reachable from values: TPC-H's `c_name`/`s_name` are customer and supplier holding structurally identical value populations.

Gaps 1, 2, 3 and 5 are also already closed on `main` — gap 3 *dissolved* by design rather than deferred, since layer detection is a labelling pass that never reaches `engine.py`'s 1:1 assignment. Full ledger in the plan doc.

## Three commits

**1. `claim` is no longer stop-listed, and a lone key no longer invents a party** (aba82c3)

The one per-case miss was `claims_ledger`: all three *dimension* parties detected perfectly at 0.90–0.93, while the fact's own `claim_claimnumber` / `claim_lossdate` / `claim_paidamount` were dropped. Cause: `claim` tokenizes out of healthcare's `claim_status` hint, so the kernel stop-lists it, and `stop -= role_tokens` is the only escape. `abbrev4` passed only because `clai` is not a type hint. Same class fixed for nine other packs in 0.7.0; recorded there as a known miss.

**The YAML fix alone was not sufficient**, which the plan did not anticipate. It regressed precision-corpus sensitivity 100% → 86%, because a lone `claim_id` then opened a one-column party. That is not a labelling slip: all 6 positives in that corpus leave exactly one column uncovered, always the fact's key. The rule satisfying both corpora is that a single-column group is viable only when the token IS the whole column (`bank`, `lender`) — a kernel change across four surfaces. Cost, stated rather than absorbed: a lone `plant_code` no longer reports a `place` party; the kind is asserted on a two-column plant instead, so that test still proves what it is named for.

**2. Audit of what else the stop-list swallows** (33f91dd)

The obvious follow-up: was `claim` one of many? **206 distinct tokens across all 16 packs.** Deliberately not bulk-declared. Most should stay suppressed (`rate`, `price`, `mileage`, `sqft`, `gpa`), and with both corpora at 100% nothing in the set is costing measurable accuracy — so no instrument can say which of the plausible ones (`order`, `shipment`, `purchase`, `enrollment`) deserve a declaration. Declaring on a hunch trades a stop-list that suppresses a real party for a role vocabulary that invents one, and `MIN_SPECIFICITY = 1.0` decides that the same way every time. Harness and output committed so the list is not re-derived.

**3. Single-char qualifiers, when they cleanly partition a frame** (d5849d4)

The largest measured error: a quarter of the corpus at pairwise F1 0.08. 0.7.0 pinned it as a known blind spot with a note to invert the test if ever fixed; this inverts it.

**The evidence is the whole frame's shape, not the token.** A short qualifier is admitted only when every column carries one at the same end, from ≥2 distinct tokens each naming ≥2 columns. Lowering `MIN_QUALIFIER_LEN` instead would admit `f_`/`x_` everywhere and take specificity with it — so a stray `x_flag` and a table-wide `t_` prefix both stay unsplit, each pinned on all three surfaces.

**The residual 3 of 8 is irreducible from column names**, and the test derives that rather than pinning a list: exactly the schemas whose entity initials collide (`provider`/`patient` → `p`). Pairwise recall stays at 1.0 on every miss, so the failure mode is two parties fused, never one wrongly split. `initial` deliberately does not join the headline floors — folding a permanent ceiling into one number would report it as a score.

## Ratchets

`MIN_EXACT_PARTITION` 0.91 → 1.0 and `MIN_PAIRWISE_F1` 0.98 → 1.0. That corpus has now **saturated on detectable conventions**, which is a reason to grow it rather than to celebrate; the discriminating signal moves to `initial` and the blind-labelling corpus.

## Verification

- infermap 418 passed / 122 skipped; goldencheck-types 16; consumers `test_segments` 15 and `test_infer_schema_stage` 13
- `infermap-core` 30 tests, `cargo clippy -D warnings` clean, `rustfmt` applied
- Both harnesses re-run; codemap regenerated; `regen_docs.py --check` and `check_docs_staleness.py` both pass
- **Not run locally: the TypeScript parity suite** — `node_modules` is absent in this worktree. The TS mirror is a line-for-line port of the Rust kernel and reads the same re-pinned oracle (3 new cases); CI runs it.

## What remains on #2574

Part B (does the name-blind shape occur often enough to justify value-aware detection?) and Part C (blocked on B's gate and on the spec the issue asks for). A3's residual sharpens B considerably: the initial-collision schemas are a mechanically-labelled instance of the name-blind shape, and a better motivation for value signals than the issue's own `routing number → bank` example, because there the columns exist and only the qualifier is ambiguous.


benzsevern:
## Correction: the fused-layer claim in this PR's description is wrong

The description says A3's residual is

> a mechanically-labelled instance of the name-blind shape, and a better motivation for value signals than the issue's own `routing number -> bank` example, because there the columns exist and only the qualifier is ambiguous.

Part B's audit (`scripts/layers_nameblind_audit.py`) measured it. **All three fused layers have DISJOINT remainder vocabularies:**

```
tpch_orders_parties  initial  `s`  shipment+supplier   name-separable
sakila_payment       initial  `s`  staff+store         name-separable
claims_ledger        initial  `p`  patient+provider    name-separable
```

`p_npi` / `p_specialty` and `p_firstname` / `p_lastname` / `p_birthdate` share no remainder token. A name-side rule that sub-clusters a group by remainder vocabulary could split them. They are **not** evidence for value-aware detection.

Nothing in the code or the measurements changes -- the residual is still there, still irreducible *by the current rule*, and the test that derives it from the schema fixture is still correct. What was wrong was my attribution of the cause: I called it a value-signal problem when it is a name-side one I had not looked hard enough at.

The plan doc committed in this PR carries the same wrong claim in its Status block; Part B's branch corrects it there.

PR #2762, opened 2026-08-25
feat(autoconfig): measure the budget's cost, and record why the fix is not shippable (#2756)

### LINKED ISSUE #2756: ControllerBudget.for_dataset predicts iteration cost from row count and is wrong by 16x, so wide-text lanes commit v0 by construction
Split out of #2748, where it was measured but is off-topic for that issue.

## The defect

`ControllerBudget.for_dataset` keys the time budget on row count alone.
Measured on two lanes that land in the SAME `< 5_000` row tier:

| lane | rows | text width | iteration cost | budget | iterations run |
|---|---|---|---|---|---|
| Abt-Buy (dedupe) | 2,173 | 249 ch/row | 2.2s | 15s | 4 |
| Amazon-Google (dedupe) | 4,589 | 677 ch/row | **36.3s** | **15s** | **1** |

**16x cost error on a 2.1x row difference.** The `< 5_000` tier grants 15s, the
tightest in the table, on the assumption that the smallest datasets are the
cheapest -- which holds for row count and not for work.

The cut is `over_budget(...) and iteration > 0`, so iteration 0 always runs and
iteration 1 is cut immediately. Amazon-Google therefore **commits v0 by
construction**, however well-aimed its rules are. The direction is perverse:
expensive data is usually complex data, which is where adaptation earns most.

## What does NOT fix it

A minimum-iterations floor (`iteration >= 2`), on the reasoning that one
iteration MEASURES but cannot ACT. Mechanically it worked -- `iter 1 start`
appeared for the first time -- and bought nothing:

| | before | after |
|---|---|---|
| F1 | 0.1097 | 0.1097 |
| wall | 74.25s | **104.73s (+41%)** |

Reverted, not shipped. Abt-Buy was correctly unaffected (iteration-limited, not
time-limited), which confirms the two lanes fail for different reasons.

Why iteration 1's candidate still loses was not established -- the diagnostic
died with `numpy._core._exceptions._ArrayMemoryError` on the local box. Two
untested hypotheses: the proposal is genuinely not better, or it is better and
`pick_committed` discards it (that lane's `cluster_size_max` is 59 against a cap
of 100, ratio 0.59, barely over the 0.5 clip bar added in #2749 -- so the
saturation tiebreak may be too weak there).

## What is actually needed

A budget keyed on something cost-correlated -- text width, or measuring
iteration 0 and adapting from its observed cost. Predicting from row count is
wrong on its face at 16x error.

Interacts with the determinism caveat already recorded at that cut: a
wall-clock stop makes the committed config a function of host speed, which is
why `GOLDENMATCH_AUTOCONFIG_DETERMINISTIC=1` exists. A cost-adaptive budget has
the same property and needs the same escape hatch.

Caveat: measured on the Amazon-Google dedupe row, retired in #2753. The budget
defect is a property of `for_dataset`, not of that benchmark, but the specific
numbers above are no longer gated.

Context: #2748, #2749, #2753.

benzsevern:
Closes #2756. **No behaviour change** — the fix was built, measured, and abandoned. What ships is the harness and the reason.

## Both halves of #2756, measured

The issue says `ControllerBudget.for_dataset` predicts iteration cost from row count and is wrong by 16x, so wide-text lanes commit v0 by construction.

The proxy really is poor — ~4.5x today, and the 16x collapsed once #2747 capped an unbounded bridge measurement. But **it is not the mechanism**. Iteration 0 is the mandatory *measurement* pass; charging it against the budget that caps *optional* adaptation is what starves the search, and no proxy can fix that, because the budget was being spent on work the proxy was never asked to predict.

Restarting the clock after iteration 0 is worth **F1 0.1097 → 0.1490** on Amazon-Google dedupe. That number is real and reproducible with the committed harness.

## Why it cannot ship

The gain and the defect are the same fact. On a host slow enough for the budget to bind, that cut is precisely what makes the controller **cheap** — one adaptive iteration and stop. Restarting it makes the machine least able to afford the work do the most of it.

Measured on the arrow-parity `shared-email-switchboard` shape, only the budget varied:

| | main | clock restart | + 2× ceiling |
|---|---|---|---|
| **budget binds** | 2 calls, 6.1s | **4 calls, 15.8s** | 2 calls, 6.5s |
| budget idle | 8 calls, 18.9s | 8 calls, 22.4s | 8 calls, 20.8s |

Under `-n auto` that killed a CI worker on **five consecutive runs**, and never once on an idle box — at ~7s per call the 15s budget doesn't bind there. The 2× ceiling didn't save it either: it bounds each controller run, and that test makes two.

Frugality-under-load and adaptation-under-load are the same knob pointed in opposite directions.

## Four wrong explanations, then an experiment

Each fit the evidence I had bothered to collect at the time:

1. **pytest-split shard shift** — my new test isn't in `.test_durations`, and this repo has that exact recorded failure mode. Refuted: the crashing test is in shard 1 on a *passing* PR too.
2. **Extra iterations slowing the shard** — refuted by the summaries: the failing shard was **faster**, 640s vs 722s, same 2084 tests.
3. **An infrastructure flake** — refuted 5–0 against main.
4. **The mandatory pass being multiplied** — measured at 2.6×, ceiling built, **still crashed**.

The fifth was settled by an isolation commit rather than a fifth theory: strip my own tests, keep the controller change, and the crash reproduced exactly. That commit is not in this PR.

Worth naming the pattern: my own local measurement said the change was free (39.53s vs 39.52s on that file). It *is* free — precisely when the budget doesn't bind, which is never true on a contended runner and always true on my desk. I measured the right thing under the wrong condition and trusted it over three red CI runs.

## What ships

- `scripts/controller_budget_cost.py` — reproduces both sides, the gain and the cost.
- A comment at the budget cut saying **do not restart this clock**, with the numbers and the reason, so the next person to have this idea starts where this one stopped.
- A note at `for_dataset` recording that the proxy is still poor and deliberately untouched.

Zero production behaviour change. 72 tests across the controller and determinism suites pass unchanged, because nothing they cover moved.


benzsevern:
## Root-caused, and it was mine. Three explanations refuted first.

The CI crash (`worker gw0 crashed` on `test_arrow_native_cluster_equivalence[shared-email-switchboard]`, three runs running) is a real regression from this PR. Fixed by a second guard, pushed.

### What was wrong

Restarting the clock says the mandatory measurement pass must not **eat** the adaptive budget. It must not **multiply** it either — and without a ceiling it did. A host slow enough to blow the budget on iteration 0 then spends a full adaptive window as well, so the machine least able to afford the work does the most of it.

Measured on the shape that crashed, only the budget varied:

| | main | clock restart | + ceiling |
|---|---|---|---|
| **budget binds** | 2 calls, 6.1s | **4 calls, 15.8s** | 2 calls, 6.5s |
| budget idle | 8 calls, 18.9s | 8 calls, 22.4s | 8 calls, 20.8s |

2.6x the work on the most expensive test in the shard, times every xdist worker.

### Why my first measurement said it was free

That test takes ~7s per `dedupe_df` call on an idle box, so a 15s budget **never binds** there and both versions run identically. I measured the file at 39.53s vs 39.52s and concluded the change cost nothing. It costs nothing exactly when the budget does not bind — which is never true on a contended runner and always true on my desk.

### Three refuted explanations, each of which looked right

1. **pytest-split shard shift.** My new test is not in `.test_durations`, and this repo has a recorded shard-shift failure mode. Refuted: the crashing test runs in shard 1 on a *passing* PR too.
2. **Extra iterations making the shard slower.** Refuted by the shard summaries: the failing run was **faster** — 640s vs 722s, same 2084 tests.
3. **Infrastructure flake.** Refuted by three consecutive failures against zero on main in six runs.

### The #2756 gain survives

Amazon-Google dedupe still commits **0.1490** against the 0.1097 floor: its real iteration 0 costs 4.6s and its last adaptive iteration starts 14.0s into a 15s budget, comfortably inside a 30s ceiling. `2x` is the smallest multiple that fits it, so it is chosen rather than picked.

Pinned by a test that fails if either guard stops holding.

PR #2790, opened 2026-08-27
perf(ci): refresh .test_durations (first scheduled harvest)

benzsevern:
First output of `harvest-test-durations.yml` ([run 33099601407](https://github.com/benseverndev-oss/goldenmatch/actions/runs/33099601407)). All three harvest shards succeeded; the workflow's own `gh pr create` was refused by the "GitHub Actions is not permitted to create or approve pull requests" repo setting, so this PR is opened by hand — which is better anyway, since an Actions-opened PR would not have triggered CI.

## What changed

- committed: **7,529** tests, 1431s total
- harvested: **10,584** tests, 663s total
- **added 3,118, dropped 63**

3,118 added tests is the staleness, stated plainly: that is six weeks and 315 commits' worth of tests that pytest-split has been placing by its count-based fallback rather than by measured duration.

## The regression hypothesis is not supported

I flagged earlier that some tests read 11-24s in the committed file but measured 64-86s in CI, and said I would not guess whether that was real product slowdown or self-inflicted co-scheduling. The harvest answers it: **the largest increase anywhere is +2.1s**, and the fresh file totals *less* than the old one across 40% more tests.

So it was contention, not a product regression. The stale split concentrated the memory-heavy tests into shard 1, and their concentration is what made them slow — which is exactly the self-reinforcing loop the workflow's header comment describes.

## Do not merge on the diff alone

Refreshing durations re-shuffles pytest-split group membership, and this suite has tests sensitive to which shard they land in:

- `test_suggest_full_dist` OOM-kills its xdist worker when co-scheduled with other memory-heavy tests (deselected, run serially).
- `tests/test_memory_pipeline.py` / `test_memory_postflight.py` carry deselects for a registration clobber that only appears in certain groupings.

Both have been re-exposed by reshuffles before (#1632, #2664). **This PR's own CI run is the check.** Read the shard walls and the pass counts before merging, not just the green tick.


benzsevern:
Closing: this durations file was produced by the harvest job running pytest **without** `--cov`, while every `python_goldenmatch` shard in ci.yml runs **with** `--cov=goldenmatch --cov-report=`. Coverage is a ~3x term on this suite's slow tests (26.18s vs 78.53s on one, measured locally), so these numbers are systematically low and cannot balance an instrumented run.

That was on top of the original reason to hold: the refresh delivered no rebalance at all -- shard 1's membership came back byte-identical (2085 passed / 132 skipped, same slowest list, same order).

#2793 fixes the harvest job. The next scheduled run produces a file worth committing.

PR #2924, opened 2026-09-10
fix(fs): numpy route banded similarity in float32; pairs on a level cut dropped a level

### LINKED ISSUE #2922: FS numpy path bands similarity in float32, native in f64: pairs exactly on a level cut drop a level
## Symptom

The native FS kernel and the numpy scoring path produce different partitions for TF adjustment (`GOLDENMATCH_FS_TF_ADJUSTMENT=1`) on the same commit, clamp and data. Only the scoring route differs.

| dataset | native (run 34476231805) | numpy (run 34477566639) |
|---|---|---|
| historical_50k | F1 0.8195, **303,998** predicted pairs | F1 0.8204, **303,314** predicted pairs |
| cotenant_hardneg | F1 0.9750, **1,287** pairs | F1 0.9781, **1,279** pairs |
| person / ncvr_synthetic / household_hardneg | identical | identical |

The baseline (TF off) arm of both runs is identical, so this is specific to TF.

The numpy run used `GOLDENMATCH_FS_TF_CLAMP=10`, which forces TF fields off the kernel at the kernel's own clamp value, as a same-clamp control. That knob lands with the TF clamp PR.

## Why CI did not catch it

`test_fs_tf_route_matrix_1805.py` has native-vs-vectorized TF parity tests (`test_native_matches_vectorized_on_tf`, `test_bucket_native_matches_vectorized_on_tf`), but the only `ci.yml` lane that runs the file invokes it with `-k "not native and not Native"`, which deselects both. They also fail locally on any box whose installed `goldenmatch-native` predates the `tf_freqs` kwarg.

## What is ruled out

The per-pair logic looks like a faithful mirror. `fs-core` applies `tf.adjustment(a)` when `level == top && a == b`, which matches numpy's `equal & (lvl == top)`. `TfTable::adjustment` uses the same `clamp(log2(collision / freq), ±10)` with 0 for out-of-table values.

## Candidates, unverified

- **Score clamping at the ends of the scale under linear min-max.** A large rare-value bump can push a pair's normalized score past 1.0 (or a common-value penalty below 0). If one route clamps and the other doesn't, threshold decisions are unaffected but the cluster auto-split, which orders MST edges by score, could split differently.
- **Rounding at the threshold boundary.** Both routes are documented as rounding to 4dp. A float32/float64 difference at `>=` would flip only borderline pairs, which fits a 0.2% difference.
- **Empty string vs null.** Numpy assigns a code to `""` (only `None` is skipped); the kernel may treat empty as missing.

## Next step

Re-enable the two native parity tests in a lane that builds the kernel, then bisect: the unit fixture first, and if it passes, the historical_50k pairs that differ between the two runs.

benzsevern:
Closes #2922.

## The bug

The numpy FS route and the native kernel band similarity with identical rules (`sim >= 0.95`, `sim >= partial_threshold`) but at different precision:

| (`anderson`, `andersen`), Jaro-Winkler | similarity | `>= 0.95` | level | score |
|---|---|---|---|---|
| true value | 0.95 | | | |
| native kernel (f64) | 0.95 | yes | 2 | 1.0 |
| numpy route, before | 0.949999988079071 | **no** | 1 | 0.5773 |
| numpy route, after | 0.95 | yes | 2 | 1.0 |

`scorer._fuzzy_score_matrix` builds similarity in float32 (the weighted matcher's memory contract, from the native `score_field_matrix` kernel and `pure_field_matrix` alike), then casts to float64 before returning. The cast can't restore the lost digits, and it hides the truncation from any downstream dtype check.

**It was filed as a TF bug. It isn't one.** TF happened to be on when I first saw it. With TF off the panel still split: historical_50k 241,825 numpy vs 241,767 native predicted pairs.

## What it affected on the numpy route

- **Level banding** in `score_probabilistic_vectorized` and `score_probabilistic_vectorized_batch`.
- **Negative evidence**, which fires when similarity is strictly below its threshold. Float32 slack fired it where the kernel doesn't.
- **Train/score consistency.** EM builds levels with scalar `score_field` in f64, so the numpy route scored with different levels from the ones it trained on.

## The fix

Recompute only the cells within 1e-6 of a cut, using `score_field`: the scalar EM trains with, asserted byte-identical to the kernel's `score_one`. Float32 error on a [0, 1] similarity stays under ~6e-8, so every other cell bands the same at either precision and keeps its fast value. `_similarity_cuts` lives beside `_levels_from_similarity`, so a banding cut can't be added without a matching repair.

**Rejected alternatives:**
- **An epsilon on `>=`.** A true 0.9499995 would land in the upper level where the kernel keeps it in the lower one, trading one divergence for another.
- **A float64 field matrix everywhere.** That would replace the Rust field-matrix kernel with pure Python for every numpy-route user (live wherever a matchkey has an `ensemble` field), or else need a new f64 kernel symbol and a wheel republish.

My first attempt gated the repair on `raw.dtype == np.float32` and never ran, because the float32 is already cast to float64 by then. Instrumentation showed that. The gate is now on the scorer, excluding only model-backed scorers (float64 cosine, with no scalar equivalent).

## Evidence

- **Local probe, fresh kernel:** native-vs-numpy differing pairs, `jaro_winkler` **112 → 0** across partial threshold 0.80/0.75 and TF off/on. `levenshtein` and `token_sort` stay at 0.
- **Worst-case cost:** 3,000 distinct prefix-heavy names (9M cells). 11,022 cells (0.12%) sat on a cut, all previously truncated. Repair took 0.20 s against a 0.33 s matrix build.
- **Tests:** the exact measured pair and value, with known-positives that the raw matrix really is truncated and the pair really sits on the cut. They also cover all three banding modes, negative evidence, and both numpy routes against the scalar route. 236 existing FS tests pass, including the native parity suites CI filters out.
- **Panel confirmation** (native vs numpy, TF off, should now give identical partitions): a comment will follow before this is armed.


benzsevern:
## Panel confirmation: cotenant now identical, historical_50k still differs. Not arming yet.

Native vs numpy, TF off, run [34481167593](https://github.com/benseverndev-oss/goldenmatch/actions/runs/34481167593) on this PR's head `d7dbf4afa`:

| dataset | before fix (numpy / native pairs) | after fix |
|---|---|---|
| cotenant_hardneg | 1,230 / 1,234 **differs** | 1,234 / 1,234 **same** |
| historical_50k | 241,825 / 241,767 | 241,856 / 241,767 **still differs** |
| person / ncvr_synthetic / household_hardneg | same | same |

The description promised identical partitions before arming. historical_50k isn't identical, so this stays unarmed until the residual is explained.

### What I've found

- **historical_50k scores `first_name` with `given_name_aliased_jw`**, confirmed from a local auto-config run. That scorer has a separate native/numpy divergence: `score_matrix` treats two names as aliases when their *member* sets intersect, while `score_pair` and the kernel require their *canonical* sets to intersect. Filed as #2925.
- **This PR doesn't cause that divergence or change it.** On a fixed-seed fixture with a fresh kernel, native-vs-numpy differing pairs before and after this fix:

  | scorer | before | after |
  |---|---|---|
  | jaro_winkler | 72 | 0 |
  | name_freq_weighted_jw | 72 | 0 |
  | ensemble | 72 | 0 |
  | given_name_aliased_jw | 18 | 18 (identical pairs) |

  Same under both missing-value modes.
- The pair-count gap growing from 58 to 89 doesn't by itself mean this fix added wrong edges. A partition's pair count isn't monotone in edge disagreement: correcting some edges can merge clusters that contain alias-divergent members.

### Next

Fix the alias rule on a branch stacked on this PR, then rerun native vs numpy. If historical_50k comes back identical, both fixes are confirmed together, and this PR gets armed.

benzsevern:
## Confirmed: with #2925's fix stacked, native and numpy give identical partitions on every panel dataset

Native vs numpy, TF off, run [34482807520](https://github.com/benseverndev-oss/goldenmatch/actions/runs/34482807520) on `fix/given-name-alias-rule` @ `8968098a4`, which is this PR plus the #2925 alias-rule fix:

| dataset | numpy / native predicted pairs | partition |
|---|---|---|
| person | 379 / 379 | same |
| ncvr_synthetic | 2,508 / 2,508 | same |
| historical_50k | **241,767 / 241,767** | **same** |
| household_hardneg | 1,468 / 1,468 | same |
| cotenant_hardneg | 1,234 / 1,234 | same |

The historical_50k residual left after this PR's panel run was entirely #2925: `score_matrix`'s looser alias rule on the `first_name` field. Together, the two fixes remove the whole native/numpy divergence on the panel.

This PR stands on its own: it fixes the float32 banding for jaro_winkler, name_freq_weighted_jw, levenshtein, token_sort and ensemble, and cotenant was already identical on its own run. Arming now. The #2925 fix follows as a separate PR against `main` once this lands.

